PR #4, opened 2019-08-19
Custom airgap

areed:
All supported versions of Kubernetes, Docker and addons are bundled into
individual tar archives in CI and uploaded to S3. When a new installer
is created, the server will create a Job to generate an airgap bundle.
The server passes the packages required for the installer to the Job
along with a signed URL for S3 where it can upload the bundle. The Job
is a simple bash script that downloads all the packages and reassembles
into a single archive.

The online install script uses the same packages used in the airgap
bundle but downloads them at runtime. The online server redirects all
dist/ (pacakge) and bundle/ requests to S3.

Future work:
- For new named installers the bundle may exist with a different ID. It
would be more efficient to re-use bundles when available.
- After POSTing a new installer it takes a minute for the airgap bundle
to be ready.
- There's no reconciler that detects failed airgap bundles and retries
or escalates.
- No garbage collection for ConfigMaps and Jobs.


bin/create-bundle-alpine.sh
```diff
@@ -0,0 +1,24 @@
+#!/bin/sh
```
areed: This is the script that creates custom airgap bundles at runtime. The server adds it to a ConfigMap that the Job mounts. The server then passes the signed PUT url as the first arg and a list of URLs to individual packages to include.

bin/docker-save.sh
```diff
@@ -0,0 +1,15 @@
+#!/bin/bash
```
areed: This runs in CI to save docker images as tar archives so they can be included in individual packages.

bin/upload-dist.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/bash
```
areed: This uploads all the individual packages to S3 during CI. Online installs download at runtime and airgap bundles download these when the Job runs.

bin/upload-latest-bundle.sh
```diff
@@ -0,0 +1,39 @@
+#!/bin/bash
```
areed: This runs in CI to create the special airgap bundle named latest.

kustomize/base/namespace.yaml
```diff
@@ -0,0 +1,4 @@
+apiVersion: v1
```
areed: All the airgap Jobs run in this namespace so it's easier to keep track of what's been created.

packages/kubernetes/1.15.0/Manifest
```diff
@@ -0,0 +1,4 @@
+kubernetes gcr.io/google-containers/hyperkube:v1.15.0
```
areed: List of images to include in the package. This is parsed in CI.

scripts/Manifest
```diff
@@ -0,0 +1,8 @@
+KUBERNETES_VERSION=1.15.2
```
areed: Latest manifest. This is turned into a template for the server to serve other install scripts.

web/src/controllers/Scripts.ts
```diff
@@ -1,125 +1,86 @@
...
+
+      let installer = await this.installerStore.getInstaller(installerID);
+      if (!installer) {
         response.status(404);
         return notFoundResponse;
       }
-      installer = i;
-    } catch (error) {
-      return { error };
+      installer = installer.resolve();
```
areed: replace "latest" with actual versions

web/src/kubernetes/index.ts
```diff
@@ -0,0 +1,108 @@
...
+    private readonly templates: Templates,
+  ) {
+    this.distOrigin = `https://${process.env["KURL_BUCKET"]}.s3.amazonaws.com`;
+    this.ns = process.env["KURL_NAMESPACE"] || "kurl";
+    this.disabled = !!process.env["AIRGAP_BUNDLING_DISABLED"];
+    this.s3Signer = new S3Signer();
+    this.bucket = process.env["KURL_BUCKET"] || "kurl-dev";
+  }
+
+  public async runCreateAirgapBundleJob(i: Installer) {
```
areed: First create a ConfigMap with three scripts:
1. the bash script to rebundle all the individual packages
2. the rendered install.sh script
3. the rendered join.sh script

web/src/client/index_test.ts
```diff
@@ -268,6 +321,29 @@ spec:
...
 `);
+  });
+
+  describe("/installer/latest?resolve=true", () => {
+    before(async () => {
+      await client.postInstaller(latest);
+    });
+
+    it("returns yaml with version", async () => {
+      const yaml = await client.getInstallerYAML("latest", true);
```
divolgin: doesn't look like that second parameter is used in `getInstallerYAML`

web/src/kubernetes/index.ts
```diff
@@ -0,0 +1,108 @@
...
+      },
+    };
+    const { body: cm } = await coreV1Client.createNamespacedConfigMap(this.ns, configMap);
+    const name = cm.metadata ? cm.metadata.name : genName;
+    const packages = i.packages().map((pkg) => `${this.distOrigin}/dist/${pkg}.tar.gz`);
+
+
+    const { signedUrl } = await this.s3Signer.signPutRequest({
+      Bucket: this.bucket,
+      Key: `bundle/${i.id}.tar.gz`,
```
divolgin: is it worth checking here if the file already exists in s3 and then just not running the job if it does?

web/src/util/persistence/s3.ts
```diff
@@ -0,0 +1,88 @@
...
+  }
+
+  public baseURL(params: GetParams): string {
+    return `https://${params.Bucket}.s3.amazonaws.com/${params.Key}`;
+  }
+
+  public parse(url: string): GetParams {
+    const parsed = new URL(url);
+    const Bucket = parsed.host.split(".")[0];
+    const Key = parsed.pathname.replace(/\//, "");
```
divolgin: not sure if this function is used, and what the purpose of this is, but slashes are an important part of the key in S3.

web/src/kubernetes/index.ts
areed: For named installers belonging to teams the bundle will need to be replaced.

web/src/util/persistence/s3.ts
areed: not used, I'll delete

web/src/client/index_test.ts
areed: I forgot to implement that in the client. Added now.

PR #28, opened 2019-09-12
Add support for nodeless workers

ldx:
This adds support for creating nodeless clusters via Elotl Milpa: https://www.elotl.co/kiyotdocs

Currently only AWS is supported, and the IAM role pointed out in the README should be attached to the EC2 instance where the install script is run.

I tested a full replicated setup too, and currently containerd fails to pull `registry.replicated.com/library/premkit:1.3.1`. I can change the CRI setup to use dockershim instead of containerd for now to work around what looks to be a registry issue here.

### DIFF
--- a/README.md
+++ b/README.md
@@ -68,6 +68,7 @@ The install scripts are idempotent. Re-run the scripts with different flags to c
 | ip-alloc-range                   | Customize the range of IPs assigned to pods                                                        |
 | load-balancer-address            | IP:port of a load balancer for the K8s API servers in ha mode                                      |
 | service-cidr                     | Customize the range of virtual IPs assigned to services                                            |
+| nodeless                         | Create a nodeless setup (see [section below](#Nodeless))                                                       |
 | no-docker                        | Skip docker installation                                                                           |
 | no-proxy                         | If present, do not use a proxy                                                                     |
 | public-address                   | The public IP address                                                                              |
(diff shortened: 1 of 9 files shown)


areed:
Would it work to package this as three addons:
1. kubenet
2. aws cloud controller with ebs provisioner
3. nodeless

I think we'd just need to add a lifecycle hook for addons that can add patches for the kubeadm configs prior to running kubeadm init. Let's says we define `kustomize/kubeadm` as a well-known directory where addons can render their patches. Then the main install script renders the default kubeadm configs into that same directory as resources and runs `kubectl kustomize .` on it just before `kubeadm init`.

areed:
The current default kurl config is:
```
apiVersion: kurl.sh/v1beta1
kind: Installer
metadata:
  name: ""
spec:
  kubernetes:
    version: "1.15.3"
  weave:
    version: "2.5.2"
  rook:
    version: "1.0.4"
  contour:
    version: "0.14.0"
```

A nodeless install would look like:
```
apiVersion: kurl.sh/v1beta1
kind: Installer
metadata:
  name: ""
spec:
  kubernetes:
    version: "1.15.3"
  kubenet:
    version: "X"
  aws:
    enabled: true
  nodeless:
    version: "0.15.0"
```
That would render the install script without Weave or Rook so there would be no need to add the branches when nodeless is enabled. We plan to add validations so incompatible addons are rejected.

areed:
It would be great to follow the addon pattern of putting yaml to be applied in the `kustomize/addon` directory so end-users can inspect what was applied and potentially add their own patches for any tweaks they need to make.

ldx:
> Would it work to package this as three addons:
> 
>     1. kubenet
> 
>     2. aws cloud controller with ebs provisioner
> 
>     3. nodeless

There's also a step before kubeadm init where we replace the CRI (installing and configuring criproxy). Any way to do it before kubeadm init is performed?

ldx:
> I think we'd just need to add a lifecycle hook for addons that can add patches for the kubeadm configs prior to running kubeadm init. Let's says we define `kustomize/kubeadm` as a well-known directory where addons can render their patches. Then the main install script renders the default kubeadm configs into that same directory as resources and runs `kubectl kustomize .` on it just before `kubeadm init`.

That sounds good, any plans to work on this?

areed:
> That sounds good, any plans to work on this?

Not immediately but I'm sure once we get around to adding more CRIs we'll need it. If you want to take a stab at it, I think you'd just need to add `addons_pre_init` [around here](https://github.com/replicatedhq/kurl/blob/master/scripts/install.sh#L172) and then do something like [this](https://github.com/replicatedhq/kurl/blob/master/scripts/common/addon.sh#L31) to see if the addon defines `<addon>_pre_init`. And then run `kubectl kustomize` on the kubeadm config in the init step.

PR #40, opened 2019-10-08
serve airgap bundles from Go container

areed:
The Go container handles /bundles and proxies everything else to the
typescript server. The typescript server /bundles handler now returns a
json manifest with a list of packages to extract and include plus a list
of files with their contents, allowing the Go server to be a dumb
component.

### DIFF
--- a/cmd/server/main.go
+++ b/cmd/server/main.go
@@ -1,0 +1,148 @@
+package main
+
+import (
+	"archive/tar"
+	"compress/gzip"
+	"encoding/json"
+	"fmt"
+	"io"
+	"io/ioutil"
+	"log"
+	"net/http"
+	"net/http/httputil"
+	"net/url"
+	"os"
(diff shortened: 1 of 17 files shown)


cmd/server/main.go
```diff
@@ -0,0 +1,148 @@
...
+		http.Error(w, "Internal Server Error", http.StatusInternalServerError)
+		return
+	}
+
+	w.WriteHeader(http.StatusOK)
+	w.Header().Set("Content-Type", "binary/octet-stream")
+	w.Header().Set("Content-Encoding", "gzip")
+
+	wz := gzip.NewWriter(w)
+	archive := tar.NewWriter(wz)
```
divolgin: defer wz.Close() and defer archive.Close() instead of closing them at the end.  I'm not sure if these leak anything on the returns below, but they could

PR #54, opened 2019-10-24
Past and next kotsadm

marccampbell:


web/src/installers/index.ts
```diff
@@ -379,7 +379,7 @@ export class Installer {
       "0.33.0",
     ],
     kotsadm: [
-      "0.9.9",
+      "0.9.11",
```
areed: Why only 0.9.11 if we have 0.9.10 and 0.9.9?

PR #57, opened 2019-10-25
TLS ingress for kotsadm

areed:


addons/kotsadm/0.9.10/install.sh
```diff
@@ -136,3 +135,66 @@ function kotsadm_secret_session() {
...
+DNS.3 = kotsadm-web.default.svc
+DNS.4 = kotsadm-web.default.svc.cluster
+DNS.5 = kotsadm-web.default.svc.cluster.local
+IP.1 = $PRIVATE_ADDRESS
+EOF
+    if [ -n "$PUBLIC_ADDRESS" ]; then
+        echo "IP.2 = $PUBLIC_ADDRESS" >> kotsadm.cnf
+    fi
+
+    openssl req -newkey rsa:2048 -nodes -keyout kotsadm.key -config kotsadm.cnf -x509 -days 365 -out kotsadm.crt
```
emosbaugh: whats the plan for rotation?

addons/kotsadm/0.9.10/install.sh
```diff
@@ -136,3 +135,66 @@ function kotsadm_secret_session() {
...
+    kubectl apply -k "$dst/"
+}
+
+function kotsadm_tls_secret() {
+    if kubernetes_resource_exists default secret kotsadm-tls; then
+        return 0
+    fi
+
+    cat > kotsadm.cnf <<EOF
+[ req ]
```
emosbaugh: are these all the same? wanna make it a function to get this config?

emosbaugh: nm

addons/kotsadm/0.9.10/install.sh
areed: No plan. The registry cert has the same period. I think we need to make a task script that can be used to rotate the certs in the cluster.

PR #58, opened 2019-10-25
new ui to TLS flow

GraysonNull:


cmd/proxy/assets/insecure.html
```diff
@@ -36,7 +36,56 @@
...
+                <p class="text-muted">
+                  <script>document.write(
+                    'or visit <a class="replicated-link" href="' + httpsLink + '">' + httpsLink + '</a> to proceed to application'
+                  )</script>
+                </p>
+              </div>
+            </div>
+            <div class="flex-column flex1 right-block alignItems--center justifyContent--center">
+              <script>document.write(
+                '<img class="insecure-image" src="/assets/images/' + browserPlainText.toLowerCase().replace(" ", "") + '-insecure.png" />'
```
salahalsaleh1993: .replace() will not replace all occurrences? is that intentional? if not, you can do something like this `.replace(/ /g, '');

GraysonNull: I think it's ok for this yes

salahalsaleh1993: cool

PR #67, opened 2019-10-28
/tls/skip handler

areed:


cmd/proxy/kustomize/deployment.yaml
```diff
@@ -18,7 +18,7 @@ spec:
         image: kurl/proxy
         env:
         - name: NODE_PORT
-          value: "8800"
+          value: "30880"
```
emosbaugh: is this intentional?

areed: yes, to be compatible with microk8s dev env

PR #73, opened 2019-10-30
don't run 'upload_x_packages' jobs 8 times

laverya:


laverya:
not actually redundant.

PR #90, opened 2019-11-07
crt keyUsage incorrect

emosbaugh:


addons/registry/2.7.1/install.sh
```diff
@@ -103,7 +103,7 @@ IP.1 = $DOCKER_REGISTRY_IP
 [ v3_ext ]
 authorityKeyIdentifier=keyid,issuer:always
 basicConstraints=CA:FALSE
-keyUsage=keyEncipherment,dataEncipherment
+keyUsage=nonRepudiation,digitalSignature,keyEncipherment
```
emosbaugh: i know this already works but hopefully i didnt break anything

PR #112, opened 2019-11-23
kotsadm operator cluster admin

areed:
remove applicationNamespaces from kotsadm config.


addons/kotsadm/1.2.0/operator.yaml
```diff
@@ -11,16 +11,17 @@ rules:
...
 apiVersion: rbac.authorization.k8s.io/v1
-kind: RoleBinding
+kind: ClusterRoleBinding
 metadata:
-  name: kotsadm-operator-rolebinding
+  name: kotsadm-operator-clusterrolebinding
 subjects:
 - kind: ServiceAccount
   name: kotsadm-operator
+  namespace: default
```
emosbaugh: is this correct? i just dont know so im flagging it

areed: I believe so. The kotsadm add-on has a lot of places where the default namespace is hardcoded.

areed: We'll have to convert a lot of yaml to templates when we make that customizable

PR #134, opened 2019-12-14
Velero

areed:


addons/velero/1.2.0/Manifest
```diff
@@ -0,0 +1,3 @@
+velero velero/velero:v1.2.0
+velero-aws velero/velero-plugin-for-aws:v1.0.0
```
emosbaugh: is this missing 

```
      - image: velero/velero-plugin-for-gcp:v1.0.0
      - image: velero/velero-plugin-for-microsoft-azure:v1.0.0
```

areed: yes, I'll add

addons/velero/1.2.0/install.sh
```diff
@@ -0,0 +1,91 @@
...
+
+    cp -r "$src/crd" "$dst/"
+    kubectl apply -k "$dst/crd"
+
+    cp "$src/deployment.yaml" \
+        "$src/rbac.yaml" \
+        "$src/service.yaml" \
+        "$dst/"
+
+    if [ "$VELERO_USE_RESTIC" != "0" ]; then
```
emosbaugh: im not sure this is correct

addons/velero/1.2.0/install.sh
```diff
@@ -0,0 +1,91 @@
...
+
+    velero_local_rgw_store
+
+    kubectl apply -k "$dst"
+
+    velero_binary
+}
+
+function velero_binary() {
+    if [ "$VELERO_INSTALL_CLI" = "0" ]; then
```
emosbaugh: this too

addons/velero/1.2.0/install.sh
emosbaugh: i see the docs now

```
| velero-install-cli | Disable installing the velero CLI with `velero-install-cli=0`. |
| velero-use-restic | Disable the Restic integration with `velero-use-restic=0`. Pod volume data will not be included in backups if Restic is disabled. |
```

typically dont we do something like `disable-...`. its fine either way

areed: I'm pretty sure it's correct according to the usage of passing the flag `velero-use-restic=0` but I know that's not an intuitive usage. It's tricky because the default is true and the advanced option in the yaml is `useRestic` so I was trying to avoid changing the flag name.

PR #164, opened 2020-01-13
kots rest API returns 404

areed:


cmd/proxy/main.go
```diff
@@ -274,6 +274,10 @@ func getHttpsServer(upstream *url.URL, tlsSecretName string, secrets corev1.Secr
 	mux.Handle("/tls", r)
 	mux.Handle("/tls/", r)
 
+	mux.Handle("/api/v1/kots/", http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		http.Error(w, "Not found", http.StatusNotFound)
```
laverya: Should this be a forbidden error instead of not found? It feels weird to deceive like that here

PR #167, opened 2020-01-15
Labeled nodes as provisioned by kurl installation

moonlight16:


moonlight16:
The latest commits add the node labels by way of yaml files (kubeadm init).  Unfortunately I found these only add the node labels during fresh installs and node joins.  So I had to add the extra bit in the shell script to manually add the labels if they don't already exist.

PR #170, opened 2020-01-17
Change kotsadm-tls type from Opaque to kubernetes.io/tls

### LINKED ISSUE #132: Kotsadm addon: standard type for TLS secret
Right now with the kotsadm addon, kurl.sh will create a secret called `kotsadm-tls` of type `Opaque`. It would be great if this could be of type `kubernetes.io/tls` so it can be re-used in Ingress specs and other K8s tools which process Secrets of this type.

(Per @areed this was done so we can store additional TLS-related fields in the secret)

I'd like to be able to re-use these certs in any applications being managed by `kotsadm`, specifcally to integrate with the `countor` addon in this case.

There's an open question around whether kotsadm should further manage this by propagating secrets to application namespaces, but I think I'd consider that out of scope for a v1 of this.

moonlight16:
Background:
https://app.clubhouse.io/replicated/story/22517/standard-type-for-tls-secret

Fixes #132

### DIFF
--- a/addons/kotsadm/1.10.2/install.sh
+++ b/addons/kotsadm/1.10.2/install.sh
@@ -230,8 +230,7 @@ EOF
 
     openssl req -newkey rsa:2048 -nodes -keyout kotsadm.key -config kotsadm.cnf -x509 -days 365 -out kotsadm.crt -extensions v3_ext
 
-    kubectl -n default create secret tls kotsadm-tls --key=kotsadm.key --cert=kotsadm.crt
-    kubectl -n default annotate secret kotsadm-tls acceptAnonymousUploads=1
+    kubectl -n default create secret generic kotsadm-tls --from-file=tls.key=kotsadm.key --from-file=tls.crt=kotsadm.crt --from-literal=acceptAnonymousUploads=1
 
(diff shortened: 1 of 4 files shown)


addons/kotsadm/1.10.2/install.sh
```diff
@@ -230,7 +230,8 @@ EOF
 
     openssl req -newkey rsa:2048 -nodes -keyout kotsadm.key -config kotsadm.cnf -x509 -days 365 -out kotsadm.crt -extensions v3_ext
 
-    kubectl -n default create secret generic kotsadm-tls --from-file=tls.key=kotsadm.key --from-file=tls.crt=kotsadm.crt --from-literal=acceptAnonymousUploads=1
+    kubectl -n default create secret tls kotsadm-tls --key=kotsadm.key --cert=kotsadm.crt
+    kubectl -n default annotate secret kotsadm-tls acceptAnonymousUploads=1
```
laverya: I like the change, but I believe it should go as part of kotsadm 1.11.0, not 1.10.x
Also the latest in 1.10.x is 1.10.3, so this wouldn't be the patch to change in either case

PR #173, opened 2020-01-18
rbac for velero

areed:


addons/kotsadm/1.11.0/operator.yaml
```diff
@@ -27,11 +31,15 @@ apiVersion: v1
...
   name: kotsadm-operator
+  labels:
+    app.kubernetes.io/name: kotsadm
 ---
 apiVersion: apps/v1
 kind: Deployment
 metadata:
   name: kotsadm-operator
+  labels:
+    app.kubernetes.io/name: kotsadm
```
marccampbell: Same question in kots repo -- I think we need a more kots-specific label here

PR #175, opened 2020-01-20
include git sha in kurl-util docker image

laverya:


deploy/kurl-util/Dockerfile
```diff
@@ -19,3 +19,6 @@ COPY cmd/proxy/assets /assets
 COPY --from=0 /go/src/github.com/replicatedhq/kurl/yamlutil /usr/local/bin
 
 COPY --from=0 /go/src/github.com/replicatedhq/kurl/cmd/join/join /usr/local/bin
+
+ARG version=unknown
+ENV VERSION=$version
```
fshot: Are "commit" and "version" meant to match?  Not clear how the commit is actually going to get passed through to the docker build with the changes above.

laverya: whoops
fixing

PR #193, opened 2020-01-27
ADD support for centos/rhel 7.7

mzaneri:
remove strict yaml marshalling warning by filter kustomize added metadata field from config file

### DIFF
--- a/scripts/common/preflights.sh
+++ b/scripts/common/preflights.sh
@@ -30,7 +30,7 @@ function require64Bit() {
 
 function bailIfUnsupportedOS() {
     case "$LSB_DIST$DIST_VERSION" in
-        ubuntu16.04|ubuntu18.04|rhel7.4|rhel7.5|rhel7.6|centos7.4|centos7.5|centos7.6)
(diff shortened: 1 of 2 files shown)


scripts/common/preflights.sh
```diff
@@ -30,7 +30,7 @@ function require64Bit() {
 
 function bailIfUnsupportedOS() {
     case "$LSB_DIST$DIST_VERSION" in
-        ubuntu16.04|ubuntu18.04|rhel7.4|rhel7.5|rhel7.6|centos7.4|centos7.5|centos7.6)
+        ubuntu16.04|ubuntu18.04|rhel7.4|rhel7.5|rhel7.6|rhel7.7|centos7.4|centos7.5|centos7.6|centos7.7)
```
areed: The images we use to download and bundle the k8s and docker packages are `centos:7`, which has been `7.7` for the past 3 months. The only reason 7.7 install hasn't been working is because of this check. What about changing this to check for 7.4+ so that when 7.8 is released we support that automatically? Probably a product question.

PR #195, opened 2020-01-28
Rook object store backed registry

emosbaugh:


addons/registry/2.7.1/tmpl-deployment-objectstore.yaml
```diff
@@ -0,0 +1,92 @@
...
+    log:
+      fields:
+        service: registry
+    storage:
+      s3:
+        region: "us-east-1"
+        regionendpoint: http://$OBJECT_STORE_CLUSTER_IP
+        bucket: docker-registry
+        accesskey: $OBJECT_STORE_ACCESS_KEY
+        secretkey: $OBJECT_STORE_SECRET_KEY
```
emosbaugh: should this be a secret? this is how we do it in aka

areed: Probably. You should be able to reuse the `kotsadm-s3` secret for that 1 env var.

PR #203, opened 2020-01-30
Alpha kotsadm release

emosbaugh:


addons/kotsadm/alpha/Manifest
```diff
@@ -0,0 +1,10 @@
+image kotsadm-api docker.io/kotsadm/kotsadm-api:alpha
+image kotsadm-migrations docker.io/kotsadm/kotsadm-migrations:alpha
+image kotsadm-operator docker.io/kotsadm/kotsadm-operator:alpha
+image kotsadm docker.io/kotsadm/kotsadm:alpha
+image kurl-proxy docker.io/kotsadm/kurl-proxy:alpha
+image postgres postgres:10.7
+image bcrypt docker.io/epicsoft/bcrypt:latest
+image kurl-util docker.io/replicated/kurl-util:v2020.01.29-0
+
+asset kots.tar.gz https://github.com/replicatedhq/kots/releases/download/v1.11.4/kots_linux_amd64.tar.gz
```
emosbaugh: could we start releasing an alpha tag from master?

web/src/installers/index.ts
```diff
@@ -456,6 +456,7 @@ export class Installer {
       "0.9.11",
       "0.9.10",
       "0.9.9",
+      "alpha",
```
emosbaugh: @areed is this ok? how will this affect things?

areed: That's fine. Users will be able to specify alpha in their yaml spec and alpha will show as the last option on kurl.sh.

addons/kotsadm/alpha/Manifest
laverya: good idea, but even for now this could at least be `v1.12.0-beta`

```suggestion
asset kots.tar.gz https://github.com/replicatedhq/kots/releases/download/v1.12.0-beta/kots_linux_amd64.tar.gz
```

PR #209, opened 2020-02-06
kotsadm-postgres statefulset

areed:


addons/kotsadm/alpha/install.sh
```diff
@@ -81,17 +82,24 @@ function kotsadm_outro() {
     printf "\n"
 }
 
-function kotsadm_secret_auto_create_cluster_token() {
-    local AUTO_CREATE_CLUSTER_TOKEN=$(kubernetes_secret_value default kotsadm-auto-create-cluster-token token)
+function kotsadm_secret_cluster_token() {
+    local CLUSTER_TOKEN=$(kubernetes_secret_value default kotsadm-cluster-token token)
```
laverya: the field that it's stored under is `kotsadm-cluster-token` too

addons/kotsadm/alpha/tmpl-secret-cluster-token.yaml
```diff
@@ -1,8 +1,8 @@
 apiVersion: v1
 kind: Secret
 metadata:
-  name: kotsadm-auto-create-cluster-token
+  name: kotsadm-cluster-token
   labels:
     kots.io/kotsadm: 'true'
 stringData:
-  token: ${AUTO_CREATE_CLUSTER_TOKEN}
+  token: ${CLUSTER_TOKEN}
```
laverya: the key and the secret name are both `kotsadm-cluster-token` in kots
```suggestion
  kotsadm-cluster-token: ${CLUSTER_TOKEN}
```

scripts/Manifest
```diff
@@ -5,7 +5,7 @@ ROOK_VERSION=1.0.4
 CONTOUR_VERSION=0.14.0
 REGISTRY_VERSION=2.7.1
 PROMETHEUS_VERSION=0.33.0
-KOTSADM_VERSION=1.12.1
+KOTSADM_VERSION=alpha
```
laverya: was this supposed to change? What does it affect?

I just want to make sure we don't make alpha the default somehow

areed: This is used in dev to select addons without having to run the kurl web server.

addons/kotsadm/alpha/install.sh
areed: fixed

PR #214, opened 2020-02-07
minio addon

areed:


addons/minio/2020-01-25T02-50-51Z/deployment.yaml
```diff
@@ -0,0 +1,70 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  # This name uniquely identifies the Deployment
```
areed: All these yaml comments were copied directly from the source provided by minio

addons/prometheus/0.33.0/install.sh
```diff
@@ -66,3 +66,44 @@ function prometheus_crd_ready() {
     fi
     return 0
 }
+
+function prometheus_rook_ceph() {
```
areed: fix prometheus so it installs without the rook-ceph addon

web/src/client/index_test.ts
```diff
@@ -133,12 +132,21 @@ spec:
...
+  minio:
+    version: latest
+    namespace: minio
+`;
+
 describe("POST /installer", () => {
   describe("latestV1Beta1", () => {
     it(`should return 201 "https://kurl.sh/latest"`, async () => {
-      const url = await client.postInstaller(latestV1Beta1);
+      const uri = await client.postInstaller(latestV1Beta1);
```
areed: lots of tslint changes in this file

web/src/installers/index.ts
```diff
@@ -216,6 +233,161 @@ export interface InstallerObject {
 }
 
 export class Installer {
+  // first version of each is "latest"
```
areed: Lots of tslint changes in this file too - static methods have to come first

PR #216, opened 2020-02-07
openebs addon

areed:


scripts/Manifest
```diff
@@ -1,7 +1,8 @@
 KUBERNETES_VERSION=1.16.4
 DOCKER_VERSION=18.09.8
 WEAVE_VERSION=2.5.2
-ROOK_VERSION=1.0.4
+ROOK_VERSION=
```
mzaneri: are we changing default from rook to openebs?

areed: No, rook will still be the default if you request `latest`. The versions in this manifest are only for dev.

PR #217, opened 2020-02-08
Kotsadm 1.12.2 compatibility with minio

areed:
use generic object store function in kotsadm 1.12.2 for compatibility
with minio.


web/src/installers/index.ts
```diff
@@ -301,6 +301,7 @@ export class Installer {
       "1.7.4",
     ],
     kotsadm: [
+      "1.12.2",
```
areed: This was already released but lost in a bad rebase

PR #218, opened 2020-02-08
ADD injects both an Installer CRD and an Installer object into the in…

mzaneri:
…stall script, and applies them via kubectl


scripts/install.sh
```diff
@@ -229,6 +229,7 @@ function main() {
     tasks
     preflights
     prompts
+    setup_installer_crd
```
mzaneri: From talking to marc this looks like it may be a good place to add this, but want to hear feedback

web/src/util/services/templates.ts
```diff
@@ -61,6 +61,16 @@ interface Manifest {
   REPLICATED_APP_URL: string;
   VELERO_VERSION: string;
   FLAGS: string;
+  INSTALLER_YAML: string;
+  INSTALLER_CRD: string;
+}
+
+function readCRDFromYAML(): string {
```
mzaneri: This is the only additional functionality added to the TS api, don't know if there's some testing you think I should add around it.

scripts/common/yaml.sh
```diff
@@ -41,3 +41,25 @@ function setup_kubeadm_kustomize() {
...
+
+    echo $INSTALLER_CRD | base64 --decode > $CREATE_INSTALLER_CRD_YAML
+
+    kubectl apply -f $CREATE_INSTALLER_CRD_YAML
+
+    cat > $INSTALLER_BASE_YAML_FILE << EOF
+$INSTALLER_YAML
+EOF
+
+    sed -i 's/kurl.sh/cluster.kurl.sh/g' $INSTALLER_BASE_YAML_FILE
```
areed: What's this for?

scripts/install.sh
areed: kubectl doesn't exist at this point on new installs

scripts/common/yaml.sh
areed: I see, it's the API group. Can we update the Installer to have this group?

mzaneri: For now the controller-gen code requires a group while the existing kurl infra does not have one. So the CRD is cluster.kurl.sh and the yaml from the api is kurl.sh. This fixes that pending further discussion on if we want to address this on the controller-gen side or the kurl api side.

scripts/install.sh
mzaneri: edited

web/src/util/services/templates.ts
```diff
@@ -61,6 +61,16 @@ interface Manifest {
...
+    const CRDPath = path.join("/src/crds", "cluster_v1beta1_installer.yaml");
```
areed: I understand the instaler yaml spec has to be injected but the CRD seems like a better fit for the common package. https://github.com/replicatedhq/kURL/blob/master/Makefile#L25. For other yaml files such as kubeadm config yaml we put them in that package rather than injecting them into this script.

mzaneri: changed

PR #230, opened 2020-02-13
don't prompt for private IP if K8s already installed

areed:


scripts/common/discover.sh
```diff
@@ -240,3 +241,10 @@ discoverPublicIp() {
         return
     fi
 }
+
+function discover_private_ip() {
+    if [ -n "$PRIVATE_ADDRESS" ]; then
+        return 0
+    fi
+    PRIVATE_ADDRESS=$(cat /etc/kubernetes/manifests/kube-apiserver.yaml | grep advertise-address | awk -F'=' '{ print $2 }')
```
emosbaugh: this will show the error

```
cat: /etc/kubernetes/manifests/kube-apiserver.yaml: No such file or directory
```

PR #235, opened 2020-02-14
Fix upgrading k8s minor version error messages

### LINKED ISSUE #182: Error message when upgrading to lower k8s version needs line breaks
When attempting to run a kurl installer on a node that already has a later k8s version installed, the displayed error message is missing separators between sentences.

actual:
```The currently installed kubernetes version is 1.16.4The requested version to upgrade to is 1.15.3Since the currently installed version is newer than the requested version, no action will be taken```

expected: 
```The currently installed kubernetes version is 1.16.4. The requested version to upgrade to is 1.15.3. Since the currently installed version is newer than the requested version, no action will be taken.```

DevAndrewGeorge:
Fixes #182. Cheers @fshot!


fshot:
Thank you, @DevAndrewGeorge !

PR #236, opened 2020-02-14
Mount the ConfigMap with app metadata in kurl-proxy

moonlight16:
This patch also includes improved handling if the app slug was provided or not.
If not, then we'll create an empty ConfigMap (meaning no branding was provided).
If slug provided, then we'll query replicated.app for metadata.  Again, if no
metadata provided, then we'll leave the ConfigMap empty.  If provided, then
we'll populate the ConfigMap with the app's metadata (including branding).

This patch fixes branding on the TLS certs pages.

### DIFF
--- a/addons/kotsadm/alpha/install.sh
+++ b/addons/kotsadm/alpha/install.sh
@@ -26,11 +26,7 @@ function kotsadm() {
     kotsadm_etcd_client_secret
     kotsadm_kubelet_client_secret
 
-    if [ "$AIRGAP" != "1" ]; then
-        curl $REPLICATED_APP_URL/metadata/$KOTSADM_APPLICATION_SLUG > "$src/application.yaml"
-    fi
-    cp "$src/application.yaml" "$dst/"
-    kubectl create configmap kotsadm-application-metadata --from-file="$dst/application.yaml" --dry-run -oyaml > "$dst/kotsadm-application-metadata.yaml"
+    kotsadm_metadata_configmap $src $dst
 
     if [ -z "$KOTSADM_HOSTNAME" ]; then
         KOTSADM_HOSTNAME="$PUBLIC_ADDRESS"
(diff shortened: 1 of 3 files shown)


addons/kotsadm/alpha/install.sh
```diff
@@ -186,6 +182,25 @@ function kotsadm_api_patch_prometheus() {
...
+    # In that case we won't create the ConfigMap for the metadata.
+    touch $dst/kotsadm-application-metadata.yaml
+
+    if [ "$AIRGAP" != "1" ] && [ -n "$KOTSADM_APPLICATION_SLUG" ]; then
+        # If slug exists, but there's no branding, then replicated.app will return nothing.
+        # (application.yaml will remain empty)
+        echo "Retrieving app metadata: url=$REPLICATED_APP_URL, slug=$KOTSADM_APPLICATION_SLUG"
+        curl $REPLICATED_APP_URL/metadata/$KOTSADM_APPLICATION_SLUG > "$src/application.yaml"
+        cp "$src/application.yaml" "$dst/"
+        kubectl create configmap kotsadm-application-metadata --from-file="$dst/application.yaml" --dry-run -oyaml > "$dst/kotsadm-application-metadata.yaml"
```
areed: On airgap installs this file will be included already at "$src/application.yaml" so 199 and 200 should be in another block guarded by a check for the file existing.

addons/kotsadm/alpha/install.sh
```diff
@@ -186,6 +182,25 @@ function kotsadm_api_patch_prometheus() {
     cp "$DIR/addons/kotsadm/alpha/patches/api-prometheus.yaml" "$DIR/kustomize/kotsadm/api-prometheus.yaml"
 }
 
+function kotsadm_metadata_configmap() {
+    local src="$1"
+    local dst="$2"
+
+    # Create an empty application-metadata.yaml to keep kustomization happy
```
marccampbell: Is there a reason we should include empty files instead of not including the file in the kustomization.yaml?

addons/kotsadm/alpha/install.sh
```diff
@@ -186,6 +182,25 @@ function kotsadm_api_patch_prometheus() {
...
+    # There's a case where a slug is not specified, and thus no application metadata.
+    # In that case we won't create the ConfigMap for the metadata.
+    touch $dst/kotsadm-application-metadata.yaml
+
+    if [ "$AIRGAP" != "1" ] && [ -n "$KOTSADM_APPLICATION_SLUG" ]; then
+        # If slug exists, but there's no branding, then replicated.app will return nothing.
+        # (application.yaml will remain empty)
+        echo "Retrieving app metadata: url=$REPLICATED_APP_URL, slug=$KOTSADM_APPLICATION_SLUG"
```
marccampbell: Is this echo statement left here intentionally, or is this debug?

addons/kotsadm/alpha/install.sh
```diff
@@ -186,6 +182,25 @@ function kotsadm_api_patch_prometheus() {
...
+        curl $REPLICATED_APP_URL/metadata/$KOTSADM_APPLICATION_SLUG > "$src/application.yaml"
```
marccampbell: Will this work on airgap installs?

addons/kotsadm/alpha/install.sh
marccampbell: Why is this not part of the kustomization.yaml?

addons/kotsadm/alpha/install.sh
moonlight16: Yes, I left here intentionally.  I found it useful for debugging.

addons/kotsadm/alpha/install.sh
moonlight16: kotsadm-application-metadata.yaml is apart of the kustomization.yaml

addons/kotsadm/alpha/install.sh
moonlight16: I found kustomization will fail because its expecting kotsadm-application-metadata.yaml (see kustomization.yaml file).  So in the cases where there's no branding, I needed to keep kustomization happy with this empty file (as the comment implies).

addons/kotsadm/alpha/install.sh
moonlight16: The idea is to overwrite the file.  Although I'll go back and retest with airgap.

moonlight16: Ok, I addressed this in my latest commits.  Although I'm finding issues when building airgap installer, so unable to find a good way to test this path.

addons/kotsadm/alpha/install.sh
moonlight16: I decided to no longer use this empty file, instead i'm using insert_resources function to insert a kustomization resource for the application-metadata (if it exists).  See my latest commit.

addons/kotsadm/alpha/install.sh
moonlight16: I'm finding there's a separate issue in airgap bundling.  Airgap bundles are supposed to contain an application.yaml which contain the branding; I'm finding this isn't the case.  Its tedious to troubleshoot the issue, in any case, my latest commit in this PR ought to properly handle airgap if the application.yaml exists.

PR #241, opened 2020-02-19
Install kustomize binaries in kURL installations

moonlight16:


moonlight16:
It felt a bit excessive to build containers for krew and kustomize separately, so we could merge those into a single Dockerfile which curls all additional binaries required.  Thoughts?

bundles/kustomize/Dockerfile
```diff
@@ -0,0 +1,10 @@
+FROM ubuntu:18.04
+ENV KUSTOMIZE_VERSION 3.5.3
+
+RUN apt-get update
+RUN apt-get -y install curl apt-transport-https gnupg git
+
+RUN mkdir -p /kustomize
+WORKDIR /kustomize
+RUN curl -fsSLO https://github.com/kubernetes-sigs/kustomize/releases/download/kustomize%2Fv${KUSTOMIZE_VERSION}/kustomize_v${KUSTOMIZE_VERSION}_linux_amd64.tar.gz
+RUN mv kustomize_v${KUSTOMIZE_VERSION}_linux_amd64.tar.gz kustomize.tar.gz
```
laverya: if we're going to have a kustomize image, it should really be a multi-stage dockerfile
so instead of this run command, you'd have something like this:
```suggestion
RUN mv kustomize_v${KUSTOMIZE_VERSION}_linux_amd64.tar.gz kustomize.tar.gz

FROM scratch
COPY --from=0 /kustomize/kustomize.tar.gz /kustomize.tar.gz
```

moonlight16: @laverya thx but @areed had suggested a much simpler by using a Manifest file in the kURL dist process.  So I removed this code.  See latest changes.

PR #246, opened 2020-02-20
Wait for running weave-net pod before declaring weave add-on installed

### LINKED ISSUE #245: Install script hangs if weave add-on isn't deployed successfully
Currently, the default kurl install process attempts to deploy weave, and then continues after applying the weave yaml to the cluster, regardless of whether or not the weave deployment succeeds.

If there are problems with the weave deployment, no error is shown to the user, and the install proceeds to try to install the rook addon, which will likely hang during its `rook_ready_spinner` stage.  This is a problem, because it makes a weave problem look like a rook problem.

The weave add-on should use a similar pattern to the rook add-on, waiting for evidence that the weave deployment was successful, and exiting the install with an error if the deployment doesn't complete within a reasonable timeout period.

The simplest way to reproduce the weave failure is to use the `ip-alloc-range` parameter when running the kurl.sh install script, specifying an ip range that deliberately conflicts with reserved ips.

For example, to reproduce the problem on a GCP instance within the `us-west2` region, you could run ```curl https://kurl.sh/latest | sudo bash -s ip-alloc-range=10.168.0.1/16```

This problem reproduces with the default kurl.sh install config, but should also reproduce with any other kurl config that includes both weave and rook.

DevAndrewGeorge:
Fixes #245 in that the installer now hangs during the `weave` add-on install while waiting for `weave-net` pod to come up (as opposed to spuriously hanging during the `rook` install). The discussion is still live as to whether this PR will address the hang itself.

Minor proof:
```
⚙  Addon weave 2.5.2
serviceaccount/weave-net created
role.rbac.authorization.k8s.io/weave-net created
clusterrole.rbac.authorization.k8s.io/weave-net created
rolebinding.rbac.authorization.k8s.io/weave-net created
clusterrolebinding.rbac.authorization.k8s.io/weave-net created
secret/weave-passwd created
daemonset.apps/weave-net created
 [|]  
```

### DIFF
--- a/addons/weave/2.5.2/install.sh
+++ b/addons/weave/2.5.2/install.sh
@@ -22,6 +22,7 @@ function weave() {
     fi
 
     kubectl apply -k "$DIR/kustomize/weave/"
+    weave_ready_spinner
 }
 
 function weave_resource_secret() {
@@ -54,4 +55,8 @@ function weave_use_existing_network() {
         EXISTING_POD_CIDR=$(echo $weaveDev | awk '{ print $1 }')
         echo "Using existing weave network: $EXISTING_POD_CIDR"
     fi
+}
+
+function weave_ready_spinner() {
+    spinnerPodRunning kube-system weave-net
 }


addons/weave/2.5.2/install.sh
```diff
@@ -55,3 +56,7 @@ function weave_use_existing_network() {
         echo "Using existing weave network: $EXISTING_POD_CIDR"
     fi
 }
+
+function weave_ready_spinner() {
+    spinnerPodRunning kube-system weave-net
+}
```
DevAndrewGeorge: A one-line function in the event the number of checks grows in the future.

PR #253, opened 2020-02-24
Have script error out if weave deployment fails

### LINKED ISSUE #247: Weave install step should wait for the weave pods to be healthy, exiting the install with errors after a timeout period
Rather than spinning indefinitely, the weave install step should:

* wait for the weave pods to be healthy (not just 'running')
* wait a reasonable time (default to 120seconds)
* display the last few lines of error log output from the pods (`--all-containers=true` param may be helpful here, since weave pod has multiple containers)
* exit the install

DevAndrewGeorge:
Implements/fixes #247.

Checks all weave pods (as there may be multiple as it is a `DaemonSet`). The check does not pass until all [pod conditions are ready](https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/#pod-readiness-gate); this check seems to be more revealing than checking if a pod is running. I am open to hearing alternative methods for health checking.

These are the conditions in my test environment:
```json
status": {
        "conditions": [
          {
            "lastProbeTime": null,
            "lastTransitionTime": "2020-02-24T02:23:10Z",
            "status": "True",
            "type": "Initialized"
          },
          {
            "lastProbeTime": null,
            "lastTransitionTime": "2020-02-24T02:48:59Z",
            "message": "containers with unready status: [weave]",
            "reason": "ContainersNotReady",
            "status": "False",
            "type": "Ready"
          },
          {
            "lastProbeTime": null,
            "lastTransitionTime": "2020-02-24T02:48:59Z",
            "message": "containers with unready status: [weave]",
            "reason": "ContainersNotReady",
            "status": "False",
            "type": "ContainersReady"
          },
          {
            "lastProbeTime": null,
            "lastTransitionTime": "2020-02-24T02:23:10Z",
            "status": "True",
            "type": "PodScheduled"
          }
```

### DIFF
--- a/addons/weave/2.5.2/install.sh
+++ b/addons/weave/2.5.2/install.sh
@@ -57,6 +57,17 @@ function weave_use_existing_network() {
     fi
 }
 
+function weave_health_check() {
+    if [[ -z $(kubectl get pods -n kube-system -l name=weave-net -o jsonpath="{.items[*].status.conditions[?(@.status!='True')]}") ]]; then
+      return 0
+    else
+      return 1
+    fi
+}
+
 function weave_ready_spinner() {
-    spinnerPodRunning kube-system weave-net
+    if ! spinner_until 120 weave_health_check; then
+      kubectl logs -n kube-system -l name=weave-net --all-containers --tail 10
+      bail "The weave addon failed to deploy successfully."
+    fi
 }
--- a/addons/weave/2.5.2/install.sh
+++ b/addons/weave/2.5.2/install.sh
@@ -58,11 +58,10 @@ function weave_use_existing_network() {
 }
 
 function weave_health_check() {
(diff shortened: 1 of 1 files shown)


addons/weave/2.5.2/install.sh
```diff
@@ -57,6 +57,17 @@ function weave_use_existing_network() {
     fi
 }
 
+function weave_health_check() {
+    if [[ -z $(kubectl get pods -n kube-system -l name=weave-net -o jsonpath="{.items[*].status.conditions[?(@.status!='True')]}") ]]; then
```
areed: What about `kubectl get pods -n kube-system -l name=weave-net -o jsonpath="{range .items[*]}{range .status.conditions[*]}{ .type }={ .status }{'\n'}{end}{end} | grep Ready=False` to be more precise about what condition you're checking?

DevAndrewGeorge: Hm... that's actually  a very good idea. Didn't know `jsonpath` was that powerful.

PR #259, opened 2020-02-26
Accepting api token if jwt is not provided

jgruica:


web/src/util/jwt/index.ts
```diff
@@ -1,13 +1,29 @@
...
+  if (_.startsWith(auth, "Bearer")) {
+    const claims = jwt.verify(auth.split(" ").pop(), await(param("JWT_SIGNING_KEY", "/id/jwt_signing_key", true)));
+    return claims.team_id;
+  } else {
+    const fields = [
+      "teamid", "name", "access_token", "is_system_token", "last_active",
+      "read_only",
+    ];
+
+    const q = `select ${fields} from vendor_team_access_token where access_token = ?`;
```
areed: How about `select teamid, read_only from vendor_team_access_token where access_token` ?

web/src/util/jwt/index.ts
```diff
@@ -1,13 +1,29 @@
...
+    const v = [auth];
+    const pool = await getMysqlPool();
+    const result = await pool.query(q, v);
+
+    if (result.length === 0) {
```
areed: Could you return the empty string here so the caller can return a 401 to the client?

web/src/util/jwt/index.ts
```diff
@@ -1,13 +1,29 @@
...
+      throw new Error(`Token ${auth} not found`);
+    }
+
+    return result[0].teamid;
```
areed: Could you verify it's not a read_only token too?

PR #265, opened 2020-02-27
Don't fail kurl install if missing k8s assets

moonlight16:
Kustomize is packaged under an assets folder.  If that folder is missing, skip
installing kustomize and don't fail the kurl installation.

### DIFF
--- a/scripts/common/kubernetes.sh
+++ b/scripts/common/kubernetes.sh
@@ -280,6 +280,11 @@ function install_kustomize() {
         return 0
     fi
 
+    if [ ! -d $DIR/packages/kubernetes/${k8sVersion}/assets ];then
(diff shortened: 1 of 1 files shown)


moonlight16:
After side discussions, we decided to not move forward with this PR.  We'd rather have kurl installations fail as opposed to allow them to proceed.  For airgapped vendors, its difficult for them to install kustomize after installations.  So this installation failure at least flags down the missing binary from the start.

moonlight16:
This is a better solution to this issue:
https://github.com/replicatedhq/kURL/pull/271

This also has the advantage of flagging w/ an install or upgrade error if kustomize wasn't in the bundle.  So we would still not allow airgap bundle installs to proceed if kustomize was missing.

PR #275, opened 2020-03-02
only prompt to load images when missing

areed:
On cluster airgap installs check the list of images on each node's
status to ensure that all kubernetes, docker, and addon images are
available on that node. If a node is found to be missing an image, only
prompt the user to run the load-images task for that node, then
continue on to checking remaining remote nodes.

Addon manifest lines that declare images now have a 4th column
"no_remote_load". There will be no checks to ensure these images exist
on remote nodes. Currently there is a single image using this column,
the bcrypt image in the kotsadm addon which is never used in a pod.

### DIFF
--- a/addons/fluentd/1.7.4/elasticsearch/elasticsearch-statefulset.yaml
+++ b/addons/fluentd/1.7.4/elasticsearch/elasticsearch-statefulset.yaml
@@ -50,7 +50,7 @@ spec:
             value: "-Xms1024m -Xmx1024m"
       initContainers:
       - name: fix-permissions
-        image: busybox
+        image: busybox:1.31.1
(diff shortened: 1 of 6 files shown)


addons/fluentd/1.7.4/elasticsearch/elasticsearch-statefulset.yaml
```diff
@@ -50,20 +50,20 @@ spec:
             value: "-Xms1024m -Xmx1024m"
       initContainers:
       - name: fix-permissions
-        image: busybox
+        image: busybox:1.31.1
```
areed: without the tag implies `latest` which defaults to `imagePullPolicy: Always`

PR #277, opened 2020-03-03
Docker daemon.json options fix

mzaneri:
Change install script to only change exec opts in /etc/docker/daemon.json file. Also changes so that the change_cgroup_to_systemd function does not run for an existing docker install

### DIFF
--- a/scripts/common/docker.sh
+++ b/scripts/common/docker.sh
@@ -5,24 +5,23 @@ function change_cgroup_driver_to_systemd() {
     # to manage resources. For more info see the link below.
     # https://github.com/kubernetes/kubeadm/issues/1394#issuecomment-462878219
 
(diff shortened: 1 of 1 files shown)


scripts/common/docker.sh
```diff
@@ -9,35 +9,11 @@ function change_cgroup_driver_to_systemd() {
     	return
     fi
 
-    case $LSB_DIST in
-        ubuntu)
             cat > /etc/docker/daemon.json <<EOF
```
emosbaugh: can you check for the existence of the file and not overwrite it?

emosbaugh: what if `/etc/docker` doesnt exist

```
mkdir -p /etc/docker
```

emosbaugh: the indentation is strange

emosbaugh:
[ch23637]

PR #287, opened 2020-03-06
merge beta into master, resolving conflicts

laverya:


laverya:
Because making sure that there's a commit path from every release to master is a good way to make sure we don't accidentally leave something out in the future?

PR #310, opened 2020-03-20
ADD API for addon flags and descriptions

mzaneri:


web/src/installers/index.ts
```diff
@@ -28,20 +28,20 @@ export interface KubernetesConfig {
...
-    controlPlane: { type: "boolean", flag: "control-plane"},
-    certKey: { type: "string", flag: "cert-key" },
+    serviceCidrRange: { type: "string", flag: "service-cidr-range", description: "The size of the CIDR for Kubernetes (can be presented as just a number or with a preceding slash)" },
+    serviceCIDR: { type: "string", flag: "service-cidr", description: "This defines subnet for kubernetes" },
+    haCluster: { type: "boolean", flag: "ha", description: "Create the cluster as a high availability cluster (note that this needs a valid load balancer address and additional nodes to be a truly HA cluster)" },
+    masterAddress: { type: "string", flag: "kuberenetes-master-address", description: "The address of the internal Kubernetes API server, used during join scripts (read-only)" },
+    loadBalancerAddress: { type: "string", flag: "load-balancer-address", description: "Used for High Availability installs, indicates the address of the external load balancer" },
+    bootstrapToken: { type: "string", flag: "bootstrap-token", description: "A secret needed for new nodes to join an existing cluster (read-only)" },
```
areed: You can pass `bootstrap-token` to the install script on the first master. That will be used as the `kubeadm-token` that is passed to the join script for remote workers and masters. The `bootstrap-token` is usually generated by the install script but some people pre-generate it so they can automate bringing up a cluster. If you attempt to set the `kubeadm-token` flag on the join script and you hadn't passed the same value in the `bootstrap-token` flag to the install script the join will fail.

areed: All that to say it's not really `read-only`

web/src/installers/index.ts
```diff
@@ -101,12 +101,12 @@ export interface RookConfig {
...
 }
 
-const rookConfigSchema = {
+export const rookConfigSchema = {
   type: "object",
   properties: {
     version: { type: "string" },
-    storageClassName: { type: "string", flag: "storage-class-name" },
-    cephReplicaCount: { type: "number", flag: "ceph-replica-count" },
+    storageClassName: { type: "string", flag: "storage-class-name", description: "The number of replicas in the Rook Ceph pool" },
```
areed: this and the one below are swapped

web/src/installers/index.ts
```diff
@@ -235,14 +235,14 @@ export interface VeleroConfig {
...
   properties: {
     version: { type: "string" },
-    namespace: { type: "string", flag: "velero-namespace" },
-    localBucket: { type: "string", flag : "velero-local-bucket"},
-    disableCLI: { type: "boolean", flag: "velero-disable-cli" },
-    disableRestic: { type: "boolean", flag: "velero-disable-restic"},
+    namespace: { type: "string", flag: "velero-namespace", description: "The namespace to install velero into if not using the default"},
+    localBucket: { type: "string", flag : "velero-local-bucket", description: "Name of the bucket to create snapshots in the local object store"},
+    disableCLI: { type: "boolean", flag: "velero-disable-cli", description: "Don't install the velero CLI on the host" },
+    disableRestic: { type: "boolean", flag: "velero-disable-restic", description: "Don’t install the velero integration" },
```
areed: `Don't install the restic integration`

web/src/installers/index.ts
mzaneri: thanks that was really helpful, removing read-only tag

PR #314, opened 2020-03-25
wait for kotsadm and kotsadm-api deployments to be ready before continuing

laverya:


addons/kotsadm/1.14.0-beta.2/install.sh
```diff
@@ -344,3 +347,31 @@ function kotsadm_namespaces() {
         kubectl create ns "$NAMESPACE" 2>/dev/null || true
     done
 }
+
+function kotsadm_health_check() {
+    if [[ -n $(kubectl get pods -l app=kotsadm -o jsonpath="{range .items[*]}{range .status.conditions[*]}{ .type }={ .status }{'\n'}{end}{end}" | grep Ready=False) ]]; then
```
emosbaugh: What if there are no pods?

emosbaugh: I don’t think you need the bash style double bracket here

laverya: possibly; I copied this from the weave pod health check code

laverya: then we get the current 'no spinner' behavior, I suppose

laverya: but pod creation should follow the deployment creation very quickly

PR #322, opened 2020-03-30
Always configure IPVS

areed:
The platform installer used this to disable IPVS on Kubernetes 1.9, but
that's not relevant for kURL.


areed:
Ubuntu installs have been using IPVS because they already have the required modules. CentOS installer were in iptables mode. I didn't see any issues with switching a CentOS install from iptables to ipvs.

PR #330, opened 2020-04-01
rook block storage

areed:


addons/kotsadm/alpha/install.sh
```diff
@@ -384,7 +384,7 @@ function kotsadm_api_health_check() {
 }
 
 function kotsadm_api_ready_spinner() {
-    if ! spinner_until 180 kotsadm_api_health_check; then
+    if ! spinner_until 240 kotsadm_api_health_check; then
```
areed, line 387: I had a timeout at 180 seconds.

addons/velero/1.2.0/install.sh
```diff
@@ -63,8 +63,8 @@ function velero_kotsadm_local_backend() {
         return 0
     fi
 
-    if kubernetes_has_resource "$VELERO_NAMESPACE" backupstoragelocation kotsadm-velero-backend \
-        || kubernetes_has_resource "$VELERO_NAMESPACE" secret aws-credentials; then
+    if kubernetes_resource_exists "$VELERO_NAMESPACE" backupstoragelocation kotsadm-velero-backend \
```
areed, line 66: unrelated bug I noticed - the `kubernetes_has_resource` function is not defined

PR #346, opened 2020-04-08
ADD yaml merging of all fields including daemon.json conf

mzaneri:


kurl_util/cmd/installermerge/main.go
```diff
@@ -0,0 +1,233 @@
...
+		}
+
+		if key == "daemonConfig" {
+			mergedDockerConfig, _ := mergeDockerConfigData([]byte(oldVal.(string)), []byte(newVal.(string)))
+			mergedConfig[key] = string(mergedDockerConfig)
+			continue
+		}
+
+		oldValMap, isOldMap := oldVal.(map[interface{}]interface{})
+		newValMap, isNewMap := newVal.(map[interface{}]interface{})
```
divolgin, line 77: Do you ever hit this case?  YAML and JSON can't have keys that aren't strings.

kurl_util/cmd/installermerge/main.go
```diff
@@ -0,0 +1,233 @@
...
+		return nil, errors.Wrap(err, "failed to marshal merged config")
+	}
+
+	return mergedconfigdata, nil
+}
+
+func main() {
+	kurlscheme.AddToScheme(scheme.Scheme)
+
+	oldInstaller, _ := getInstallerConfigFromYaml("old.yaml")
```
divolgin, line 228: Shouldn't these files come from command line arguments?  And also, error handling?

mzaneri: the mergeConfigYamlData was really what were looking at, that will be done before its read for use in kurl

kurl_util/cmd/installermerge/main.go
mzaneri: yes, this is a quirk of YAML unmarshalling

PR #358, opened 2020-04-10
Prune images after building each package to conserve disk space

areed:


bin/upload-dist.sh
```diff
@@ -33,17 +33,24 @@ function list_all_packages() {
...
 }
 
+function upload() {
+    local package="$1"
+
+    make dist/$package
+    aws s3 cp dist/$package s3://$S3_BUCKET/staging/
+    make clean
+    if [ -n "$DOCKER_PRUNE" ]; then
+        docker system prune --all --force
```
divolgin, line 43: What's the reason for this?  This will be running on some ephemeral github server, right?

areed: The build is failing because it runs out of disk. The server only has 14GB total.

PR #368, opened 2020-04-14
Adding bin utils to airgap packages

divolgin:


.github/workflows/deploy-prod.yaml
```diff
@@ -14,15 +14,20 @@ jobs:
...
         echo "::add-path::$(go env GOPATH)/bin"
+        echo "::set-output name=VERSION_TAG::${GITHUB_REF/refs\/tags\//}"
       shell: bash
 
     - uses: actions/checkout@v2
 
-    - run: make -C web deps build
+    - run: make -C kurl_util build
+      env:
+        VERSION_TAG: ${{ env.VERSION_TAG }}
```
areed, line 28: How is this used in the make build command?

divolgin: @areed look in the `list-all-packages.sh` file

areed: I think the `VERSION_TAG` would need to be set in the jobs that call that script. The `build-upload-packages` job below calls that script and the cron-rebuild-packages workflow calls it too.

PR #373, opened 2020-04-15
Adding files to airgap bundles

divolgin:


bin/build-docker.sh
```diff
@@ -1,5 +1,7 @@
 #!/usr/bin/env bash
 
+# is this used?
```
divolgin, line 3: @areed i couldn't find any references to this file.

areed: It was used for kurl-util image but that has its own directory and makefile now

PR #381, opened 2020-04-16
ADD update to latest CRD version and fix airgap issue

mzaneri:


scripts/common/utilbinaries.sh
```diff
@@ -1,6 +1,6 @@
 
 function download_util_binaries() {
-    if [ "$AIRGAP" != "1" ]; then
+    if [ ! -d "./bin" ]; then
```
mzaneri, line 3: Because of different order of install.sh with the yaml changes, this function is run before the CRD is parsed out, so the airgrap variable isn't set yet. This is my fix to that, but I think we may want to delete the bin directory after these binaries are used so that we dont have issues on reruns of non-airgap installs

PR #385, opened 2020-04-18
Prod release builds common and kurl-utils-bin packages

areed:
The common package from staging contains replicated/kurl-util:alpha but
the prod common pakcage should have a tagged version. The kurl-utils-bin
package from staging is latest but prod needs the tagged version.


bin/list-all-packages.sh
```diff
@@ -15,12 +25,8 @@ function list_all_packages() {
...
     echo "docker-19.03.4.tar.gz"
     echo "common.tar.gz"
-    if [ -z "$VERSION_TAG" ]
-    then
-        echo "kurl-bin-utils-latest.tar.gz"
-    else
-        echo "kurl-bin-utils-${VERSION_TAG}.tar.gz"
-    fi
+    # Uncomment when this can be built from the last tagged release
+    # echo "$KURL_BIN_UTILS_FILE"
```
divolgin, line 29: This for cron builds only?

areed: For releasing to staging and prod also. I'm going to change this so it builds and let that 1 job fail on the cron prod workflow.

PR #403, opened 2020-04-24
use commonLabels instead of adding a label everywhere

laverya:


laverya:
converted to draft as this doesn't handle secrets

PR #438, opened 2020-05-05
Proposal for add-on categories

areed:


design/addon-categories.md
```diff
@@ -0,0 +1,132 @@
...
+- "Cluster Administration"
+- "Application Management"
+
+Each add-on can specify dependencies on other categories.
+
+Add-ons belonging to the same category conflict with each other.
+
+## Detailed Design
+
+Add a GET /add-ons endpoint that returns a list of add-ons along with their dependencies.
```
marccampbell, line 44: @GraysonNull will you add some detail here describing how this endpoint will be implemented?  Is there a database or a file, or where does the data come from and how can we update it?

design/addon-categories.md
```diff
@@ -0,0 +1,132 @@
...
+		requires: ["CNI Plugin", "Container Runtime"],
+	}
+]
+```
+
+## Alternatives Considered
+
+### Recommends
+
+Each add-on's `requires` list could be separated into `recommends` and `requires`.
```
GraysonNull, line 113: I think this is a good direction to go. Take in to consideration this design:

Screen Shot 2020-05-05 at 4 28 59 PM


I think there would be a lot of value in the "Works well with" section. To do this we could use the `recommends` field to grab items out of the suggested categories.

design/addon-categories.md
GraysonNull: Yeah....we need to figure that out. It seems pretty straight forward for us to generate a JSON file at build time and just serve it up all on the FE to avoid making more requests and the need for a new DB or database tables. My only concern with that is how do we add to it. Like is there a giant yaml file authors have to add to? Just feels brittle so I'm wondering if an API request to a database would be more stable over time as more add-ons are supported. 

I think a discussion with @areed and @mzaneri would be helpful in getting to a good solution for this.

design/addon-categories.md
areed: Moved to the design.

design/addon-categories.md
```diff
@@ -0,0 +1,140 @@
...
+	}
+]
+```
+
+Each addon root directory may have a categories.json file that defines an object representing the add-on.
+The addon/<version> directories do not have separate categories.json files.
+Add a Node script to bin/ that will read all the categories.json files, concatenate them into an array, and output the result as json.
+In the deploy-staging GitHub workflow, call the Node script, save the output to a tmp file, and upload the output to the S3 staging folder.
+Do the same for the prod script.
+Clients will then be able to GET the file from either https://kurl.sh/dist/add-ons.json or https://kurl-sh.s3.amazonaws.com/dist/add-ons.json.
```
GraysonNull, line 128: @areed How can we ensure that these files don't get cached causing kurl.sh to be missing newly added add-ons?

GraysonNull: I should also note that this script actually generates two files. One for the addons and one which is the supported versions for each of the addons. The supported versions is used on both the addons page and on the create installer page.

areed: S3 automatically adds ETag and Last-Modified headers when the files are uploaded.

PR #441, opened 2020-05-05
proposal for moving kurl.sh api to go

mzaneri:


marccampbell (dismissed):

design/go-api.md
```diff
@@ -0,0 +1,72 @@
...
+## Background
+
+Previously, kURL has used the backend typescript api to take a yaml config and make an install script with options and bash variables set for the install.
+With the recent changes to kURL such as the addition of the CRD, the current install script uses go binaries to parse and traverse the yaml tree to set options.
+
+We can greatly reduce code complexity from the no longer used parts of the typescript api and gain added type safety by moving this api to Go.
+
+## High-Level Design
+
+Endpoints
```
areed, line 23: The kurl.sh frontend needs the `GET /installer` endpoint to list available versions

design/go-api.md
```diff
@@ -0,0 +1,72 @@
...
+validateInstaller POST /installer/validate
+getBundle GET/bundle/{installerID}
+
+## Detailed Design
+
+postInstaller Endpoint:
+The function will take a yaml Installer and attempt to unmarshall it into an Installer CRD.
+If the unmarshalling fails, an error will be retuned. If the marshalling succeeds the version of an addon will be checked against existing versions.
+If the version is not found an error will be returned.
+Any version latest will be converted to the most recent version.
```
areed, line 40: This conversion needs to be done when the script is generated, not when the spec is posted.

design/go-api.md
```diff
@@ -0,0 +1,72 @@
...
+This endpoint will also handle JWT auth in order to validate the creater of the installer. 
+
+getInstallScript, getJoinScript, getUpgradeScript Endpoints:
+These will check the db for a specific installerID, and inject it into the script as INSTALLER_SPEC_YAML bash strings
+
+validateInstaller Endpoint:
+This endpoint will take a yaml file and do the steps similar to postInstaller Endpoint and return a response indicating if it is invalid or not.
+
+getBundle Endpoint:
+This is already served by a golang endpoint
```
areed, line 53: Currently it calls the typescript API to get the rendered scripts and the list of packages to push into the bundle.

design/go-api.md
```diff
@@ -0,0 +1,72 @@
...
+getInstallerYaml GET /installer/{installerID}
+validateInstaller POST /installer/validate
+getBundle GET/bundle/{installerID}
+
+## Detailed Design
+
+postInstaller Endpoint:
+The function will take a yaml Installer and attempt to unmarshall it into an Installer CRD.
+If the unmarshalling fails, an error will be retuned. If the marshalling succeeds the version of an addon will be checked against existing versions.
+If the version is not found an error will be returned.
```
areed, line 39: There are a couple other validations besides version. The podCidrRange and serviceCidrRange are checked by regex. Will those be ported over? We should add more validations like that in the future.

design/go-api.md
```diff
@@ -0,0 +1,72 @@
...
+This is already served by a golang endpoint
+
+## Legacy Considerations
+
+Currently, the way a url/installer hash is calculated is to take the oldstyle installflags and concat them together into a single string that is fed into a sha256 hash function, the output converted into a hexidecimal string, with the result truncated to 7 chars (there are roughly 268 million possibilities)
+
+Additionally, due to legacy considerations there are certain values that are added to the string before others.
+
+This result can be duplicated in Golang, but it probably make sense to change this to a similar format to that used in the parsing binaries for ease of use so things do not need to be updated by hand.
+However, this will mean that existing yaml will now hash to a different value.
```
areed, line 62: That's fine.

design/go-api.md
```diff
@@ -0,0 +1,72 @@
...
+If we keep all existing installerID hashes in the DB, we can maintain backwards compatibility.
+
+Additionally, to maintain backwards compatibility some shimming of old yaml spec where old was converted into new occurs.
```
areed, line 65: We can drop the migrateV1Beta1 function but I think we need to keep the legacyFieldConversion function around for a bit.

design/go-api.md
mzaneri: what is this used for currently?

design/go-api.md
mzaneri: Right I see those, added if there are any more we can them add too

design/go-api.md
mzaneri: So we store in the db as latest? Is that behavior we want instead of a concrate version?

design/go-api.md
mzaneri: noted

design/go-api.md
mzaneri: I've been thinking about this and I'm not sure the best course of action since they cannot be unmarshalled as part of the crd

design/go-api.md
areed: For the dropdowns on the homepage of kurl.sh

design/go-api.md
areed: Yes, when a new version of an add-on is released you're automatically updated if you specify latest.

design/go-api.md
areed: You could unmarshal into a map, fix the fields, then marshal back to yaml. Or use regexes.

design/go-api.md
mzaneri: I think we might want to move this to kurl.sh instead

PR #444, opened 2020-05-08
proposal to remove docker run commands from kURL

mzaneri:


design/remove-docker-deps.md
```diff
@@ -0,0 +1,53 @@
...
+
+- To utilize binaries from the kurl_util image to change config
+- To generate bcrypt hashes
+- To load images
+
+## Detailed Design
+
+Places to move functionality from kurl_util to go binaries:
+
+scripts/install.sh line 88
```
marccampbell, line 34: A link here would be helpful

design/remove-docker-deps.md
```diff
@@ -0,0 +1,53 @@
...
+scripts/common/upgrade.sh line 154
+scripts/common/common.sh 261
+
+Places to use a bcrypt binary instead of an image:
+
+addons/registry/2.7.1/install.sh line 65
+scripts/common/tasks 204
+addons/kotsadm/MANY/install.sh
+
+Places to handle loading of images:
```
areed, line 51: This will need to wait until containerd is added and then it will be either `docker load` or `ctr import`, right?

design/remove-docker-deps.md
```diff
@@ -0,0 +1,53 @@
...
+scripts/join.sh line 60
+scripts/common/kubernetes.sh lines 340, 360, 366, 401, 424, 430
+scripts/common/upgrade.sh line 154
+scripts/common/common.sh 261
+
+Places to use a bcrypt binary instead of an image:
+
+addons/registry/2.7.1/install.sh line 65
+scripts/common/tasks 204
+addons/kotsadm/MANY/install.sh
```
areed: Will old versions of kotsadm be updated and rebuilt?

design/remove-docker-deps.md
mzaneri: updated

design/remove-docker-deps.md
areed: If so, the kurl-util image can be removed from the common package, but it will need to be added to the kotsadm Manifest since kotsadm runs Jobs to generate join scripts with that image.

design/remove-docker-deps.md
mzaneri: yes this is the one place that I believe will need an if statement and either a docker load or the equivalent ctr command

design/remove-docker-deps.md
mzaneri: Yes thanks that would definitely need to be done. I was going to suggest removing the kurl-util image but that will definitely not work if kotsadm needs it.

emosbaugh:
Most if not all of this work has already been done. Closing

PR #459, opened 2020-05-13
Proxy design doc

areed:


design/proxy.md
```diff
@@ -0,0 +1,90 @@
...
+Proxy configuration uses three fields under the kurl section of the installer spec:
+
+```yaml
+apiVersion: cluster.kurl.sh/v1beta1
+kind: Installer
+metadata:
+  name: proxy
+spec:
+  kurl:
+    HTTPProxy: http://10.128.0.3:3128
```
marccampbell, line 39: How about `proxyServer` or `proxyAddress`?

design/proxy.md
```diff
@@ -0,0 +1,90 @@
...
+    additionalNoProxy: 10.128.04,10.128.0.5,10.128.0.6
```
marccampbell, line 40: Should this be array to make it easier to patch?

areed: @mzaneri Will array fields work with the spec and CRD?

design/proxy.md
areed: changed to proxyAddress

design/proxy.md
mzaneri: we have 2D arrays for IPTables et al cmds

mzaneri: I'm hesitant to change the type on this though, think its a good bit more work for not much gain

design/proxy.md
```diff
@@ -0,0 +1,90 @@
...
+## Goals
+
+- Proxy installs complete successfully.
+- Docker can pull images from remote and local registries.
+
+## Non Goals
+
+- Automatically detect proxy from environment.
+- Validate kotsadm add-on works with a proxy.
+- Support docker versions below 19.03.
```
mzaneri, line 14: outside of scope im sure, but has anyone asked for ipv6 support for anything kURL related?

design/proxy.md
mzaneri: Also maybe change the name to additionalProxyAddress

design/proxy.md
areed: no, I haven't heard any requests yet

Issue #264, opened 2020-02-27
`getNoProxyAddresses` doesn't exist


cmcarthur:
The latest kurl.sh references a function "getNoProxyAddresses" that does not exist. So, when a customer tries to specify a proxy, the install script fails. See command below highlighting the issue.

```bash
$ curl -sSL https://kurl.sh/latest | grep -n -5 'getNoProxyAddresses'
1827-        if [ -z "$PROXY_ADDRESS" ] && [ "$AIRGAP" != "1" ]; then
1828-            promptForProxy
1829-        fi
1830-
1831-        if [ -n "$PROXY_ADDRESS" ]; then
1832:            getNoProxyAddresses "$PRIVATE_ADDRESS" "$SERVICE_CIDR"
1833-        fi
1834-    fi
1835-
1836-    prompt_airgap_preload_images
1837-}
```

EDIT: example output here:

```
The installer will use network interface 'xxx' (with IP address 'x.x.x.x’)

Does this machine require a proxy to access the Internet? (y/N) y

Enter desired HTTP proxy address: https://proxy-name:8080

The installer will use the proxy at 'https://proxy-name:8080'

./script.sh: line 1832: getNoProxyAddresses: command not found
```

dexhorthy:
@laverya do you know if this has been adressed?

RatanVMistry:
I can see the same issue when I use http_proxy.
I downloaded the latest script and run ./latest http_proxy 
my VM have http_proxy, https_proxy and no_proxy set at ~/.bashrc 
 
I added some extra lines to debug 
![image](https://user-images.githubusercontent.com/20649157/80036463-9e178300-84a6-11ea-8fd5-7a3728999b04.png)

`The installer will use the proxy at 'http://172.16.1.232:8080' (imported from env var 'http_proxy')
The installer will use network interface 'ens5' (with IP address '172.16.2.198')
The installer will use the proxy at 'http://172.16.1.232:8080' (imported from env var 'http_proxy')
I am in Prompts 3rd
./latest: line 1914: getNoProxyAddresses: command not found`

dexhorthy:
@cmcarthur did you ever get a workaround in place for this? Or did you just go straight to airgap?

dexhorthy:
@areed I understand that the PR #463 fixes this -- do you have an ETA on release? Will this ship with a KOTS release or on a separate cadence?

areed:
This wil be in the 1.16 kots release.

[...]

PR #463, opened 2020-05-15
Http proxy

areed:


addons/kotsadm/alpha/tmpl-kotsadm-proxy.yaml
```diff
@@ -0,0 +1,14 @@
...
+kind: Deployment
+metadata:
+  name: kotsadm
+spec:
+  template:
+    spec:
+      containers:
+      - name: kotsadm
+        env:
+        - name: PROXY_ADDRESS
```
areed, line 11: I'd planned for this to be `HTTP_PROXY` but that broke in-cluster networking so we'll have to update kotsadm to manually lookup the proxy address from this env var and construct a client for outbound requests.

scripts/common/proxy.sh
```diff
@@ -1,25 +1,78 @@
 
+function proxy_bootstrap() {
```
areed, line 2: We need to know the proxy address to download the binaries but we don't have the binaries to parse the yaml spec so try to find it with grep.

scripts/common/proxy.sh
```diff
@@ -1,25 +1,78 @@
...
+    fi
+
+    bail "Failed to make outbound https request and no proxy is configured."
+}
+
 function configure_proxy() {
-    exportProxy
-    # kubeadm requires this in the environment to reach the K8s API server
-    export no_proxy="$NO_PROXY_ADDRESSES"
+    if [ "$NO_PROXY" = "1" ]; then
```
areed, line 29: This is after the binaries have property parsed the yaml spec and updated the relevant environment variables.

scripts/common/proxy.sh
```diff
@@ -1,25 +1,78 @@
...
+	# for curl to download packages
+	export https_proxy="$PROXY_ADDRESS"
+
+    if ! curl --silent --fail --connect-timeout 4 https://api.replicated.com/market/v1/echo/ip >/dev/null ; then
+        bail "Failed to make outbound request using proxy address $https_proxy"
+    fi
 }
 
-exportProxy() {
+function configure_no_proxy() {
```
areed, line 45: This has to be called after the addon pre-init functions since those can affect the pod and service cidrs.

scripts/install.sh
```diff
@@ -168,6 +170,11 @@ function outro() {
         dockerRegistryIP=" docker-registry-ip=$DOCKER_REGISTRY_IP"
     fi
 
+    local noProxyAddrs=""
+    if [ -n "$PROXY_ADDRESS" ]; then
+        noProxyAddrs=" additional-no-proxy-addresses=${SERVICE_CIDR},${POD_CIDR}"
```
areed, line 175: The join script has the yaml spec so it already has any additionalNoProxyAddresses defined in there but it doesn't know what CIDRs are in use. These will be appended to anything defined in the spec.

kurl_util/go.mod
```diff
@@ -6,7 +6,7 @@ require (
 	github.com/apparentlymart/go-cidr v1.0.1
 	github.com/coreos/etcd v3.3.15+incompatible // indirect
 	github.com/pkg/errors v0.8.1
-	github.com/replicatedhq/kurl v0.0.0-20200430165742-b6df5d5ede7e
+	github.com/replicatedhq/kurl v0.0.0-20200514170003-b9e4bf917dd9
```
areed, line 9: I'll have to follow up this PR with another with the updated kurlkinds code.

kurl_util/cmd/bashmerge/main.go
```diff
@@ -45,6 +45,7 @@ func getInstallerConfigFromYaml(yamlPath string) (*kurlv1beta1.Installer, error)
 
 func checkIfFlagHasValue(length int, flag string) bool {
 	shouldHaveLengthTwo := []string{
+		"additional-no-proxy-addresses",
```
mzaneri, line 48: Do we need this to be a bash flag? I suspect the answer is yes, but we're creeping back up in flags again.

areed: Yes, for the join script. I added a note to the design doc.

dexhorthy:
@areed does this fix https://github.com/replicatedhq/kURL/issues/264 ?

scripts/common/proxy.sh
mzaneri: This is also an issue with airgap, it currently will only work as a bash flag and not as yaml spec because the airgap stuff happens before the yaml is parsed. Open to suggestions to fixing this.

areed:
> @areed does this fix #264 ?

Yes, proxy installs will complete successfully.

scripts/common/docker.sh
```diff
@@ -193,91 +91,34 @@ information.\n\nhttps://help.replicated.com/docs/kb/developer-resources/devicema
...
-            yum versionlock ${1}-*
-            ;;
-        ubuntu)
-            apt-mark hold $1
-            ;;
-    esac
+function docker_get_host_packages_online() {
+    local version="$1"
+
+    if [ "$AIRGAP" != "1" ] && [ -n "$DIST_URL" ]; then
```
mzaneri, line 119: This is a great change

scripts/common/proxy.sh
areed: We could add an empty file to every airgap bundle named `airgap` and if that file exists it's an airgap install. Or, since the bin/ utils will already be there on airgap installs, we could try to use them to parse the yaml for the airgap flag and if all that succeeds it's airgap.

PR #468, opened 2020-05-18
ADD ability to use end user provided Fluentd.Conf file

mzaneri:


addons/fluentd/1.7.4/install.sh
```diff
@@ -36,6 +47,17 @@ fluentd() {
         cp -r "$logging_src/" "$logging_dst/"
         cp -r "$fluentd_standalone_src/" "$fluentd_dst/"
 
+        if [ -z "$FLUENTD_CONF_FILE" ]; then
+            FLUENTD_CONF_FILE="$fluentd_dst/fluentd.conf"
+            REMOVE_FLUENTD_CONF=1
+        fi
+
+        kubectl create configmap fluentdconf --from-file $FLUENTD_CONF_FILE --namespace logging -o yaml --dry-run=client > "$fluentd_dst/fluentd-fd-only-configmap.yaml"
```
areed, line 55: I don't think `dry-run=client` is compatible with Kubernetes 1.15

mzaneri: updating documentation to reflect

PR #488, opened 2020-05-30
Set kubernetes service cluster ip on kotsadm-api

areed:


addons/kotsadm/alpha/tmpl-kotsadm-api-proxy.yaml
```diff
@@ -11,4 +11,4 @@ spec:
         - name: HTTP_PROXY
           value: ${PROXY_ADDRESS}
         - name: NO_PROXY
-          value: ${NO_PROXY_ADDRESSES},${OBJECT_STORE_CLUSTER_IP}
+          value: ${NO_PROXY_ADDRESSES},${OBJECT_STORE_CLUSTER_IP},${KUBERNETES_CLUSTER_IP}
```
areed, line 14: There's graphql call that creates a Secret with this IP and ignores the service cidr

scripts/common/docker.sh
```diff
@@ -35,10 +35,6 @@ function install_docker() {
 
     if [ -z "$PRESERVE_DOCKER_CONFIG" ] && [ -n "$PROXY_ADDRESS" ]; then
         docker_configure_proxy
-        local dockerProxy=$(docker info 2>/dev/null | grep -i "HTTP Proxy:")
-        if ! echo "$dockerProxy" | grep -q "$PROXY_ADDRESS"; then
```
areed, line 39: This fails with an authenticating proxy because docker info redacts the credentials in the URL.

PR #513, opened 2020-06-09
REFACTOR remove htpasswd file creation from container into binary

mzaneri:


kurl_util/cmd/htpasswd/main.go
```diff
@@ -0,0 +1,26 @@
...
+func main() {
+	username := flag.String("u", "", "username")
+	password := flag.String("p", "", "password")
+	filePath := flag.String("f", "", "filePath")
+
+	flag.Parse()
+
+	if *username == "" || *password == "" {
+		flag.PrintDefaults()
+		os.Exit(-1)
```
areed, line 20: Why negative?

mzaneri: mistake, fixing thanks

PR #519, opened 2020-06-11
ADD containerd online support, centos 8

mzaneri:


scripts/common/common.sh
```diff
@@ -258,7 +271,11 @@ function get_shared() {
         rm common.tar.gz
     fi
     if [ -f shared/kurl-util.tar ]; then
-        docker load < shared/kurl-util.tar
+        if [ -n "$DOCKER_VERSION" ]; then
+            docker load < shared/kurl-util.tar
+        else
+            ctr -n=k8s.io images import shared/kurl-util.tar
```
areed, line 277: When is `-n=k8s.io` required?

mzaneri: I don't think it is actually, now reading through some documentation

mzaneri:
> Is airgap support coming later?

Later this week, containerd is currently in alpha status

PR #521, opened 2020-06-12
ADD airgap support for containerd on centos

mzaneri:


bundles/containerd-ubuntu1604/Dockerfile
```diff
@@ -0,0 +1,16 @@
+FROM ubuntu:18.04
```
areed, line 1: 16.04

PR #532, opened 2020-06-16
proxy with airgap

areed:
Configure proxy even when airgap. Do not make any outbound requests to test proxy in case it's an airgap install.

### DIFF
--- a/scripts/common/proxy.sh
+++ b/scripts/common/proxy.sh
@@ -6,7 +6,7 @@ function proxy_bootstrap() {
         printf "The installer will use the proxy at '%s' (imported from env var 'HTTP_PROXY')\n" "$ENV_PROXY_ADDRESS"
     elif [ -n "$http_proxy" ]; then
         ENV_PROXY_ADDRESS="$http_proxy"
-        export https_proxy="$HTTP_PROXY"
+        export https_proxy="$http_proxy"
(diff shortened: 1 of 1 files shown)


scripts/common/proxy.sh
```diff
@@ -1,14 +1,12 @@
 
 function proxy_bootstrap() {
-    if [ "$AIRGAP" = "1" ]; then
-        return
-    fi
-
     if [ -n "$HTTP_PROXY" ]; then
         ENV_PROXY_ADDRESS="$HTTP_PROXY"
+        export https_proxy="$HTTP_PROXY"
```
areed, line 5: This ensures the address can be used by curl to reach https urls

scripts/common/proxy.sh
```diff
@@ -1,14 +1,12 @@
...
         printf "The installer will use the proxy at '%s' (imported from env var 'HTTP_PROXY')\n" "$ENV_PROXY_ADDRESS"
     elif [ -n "$http_proxy" ]; then
         ENV_PROXY_ADDRESS="$http_proxy"
+        export https_proxy="$HTTP_PROXY"
```
salahalsaleh, line 9: should this be `$http_proxy` instead?

areed: It has to be https_proxy or curl ignores it for https urls

salahalsaleh: i meant `$http_proxy` instead of `$HTTP_PROXY`

PR #533, opened 2020-06-16
force client-go to v0.18.3 - kubevirt.io/client-go imports pre-modules version

laverya:
https://github.com/kubevirt/client-go/blob/e297527/go.mod#L46


marccampbell:
Lgtm

PR #539, opened 2020-06-16
Send support bundle and sonobuoy results to api

marccampbell:
This does a few things:

1. fix the support bundle collection
2. adds a quick sonobuoy conformance test to the output
3. removes the broken openebs spec and adds containerd with ubuntu (also broken)

### DIFF
--- a/testgrid/tgapi/pkg/handlers/instance_sonobouy.go
+++ b/testgrid/tgapi/pkg/handlers/instance_sonobouy.go
@@ -1,0 +1,22 @@
+package handlers
+
+import (
+	"fmt"
+	"io/ioutil"
+	"net/http"
+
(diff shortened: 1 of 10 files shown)


testgrid/tgapi/pkg/handlers/instance_sonobouy.go
```diff
@@ -0,0 +1,56 @@
...
+	/*
+		Plugin: e2e
+		Status: passed
+		Total: 4843
+		Passed: 1
+		Failed: 0
+		Skipped: 4842
+	*/
+
+	isSuccess := false
```
laverya, line 39: Why not just unmarshal the yaml, if the above is the format?

PR #554, opened 2020-06-19
deploy tgapi to prod from master

laverya:


marccampbell:
@laverya this looks good, but it's slightly misleading that the prod deployment happens from a file named deploy-staging now. I feel like this is going to make it more difficult to find later.

Can we rename the file or fix this somehow?

Issue #563, opened 2020-06-24
Support for configuring contour tls.minimum-protocol-version


emosbaugh:
Contour tls minimum protocol version defaults to 1.1. I would like to be able to configure this.

https://projectcontour.io/docs/master/configuration/

Contour exposes configuration in the contour ConfigMap. It would be helpful to be able to configure `tls.minimum-protocol-version` among other settings.

https://github.com/replicatedhq/kURL/blob/c91892617cbe3c2e19c1f62a1a81dd40b9896585/addons/contour/1.0.1/common.yaml#L7-L12

PR #569, opened 2020-06-25
Make contour.tlsMinimumProtocolVersion templatable

emosbaugh:


addons/contour/1.0.1/install.sh
```diff
@@ -16,6 +16,11 @@ function contour() {
 
     cp "$src/patches/service-patch.yaml" "$dst/"
 
+    if [ -z "$CONTOUR_TLS_MINIMUM_PROTOCOL_VERSION" ]; then
+        CONTOUR_TLS_MINIMUM_PROTOCOL_VERSION="1.2"
```
emosbaugh, line 20: this will default the version to 1.2. im not sure if this is desired or not

mzaneri: I think it is, 1.1- is being deprecated everywhere

emosbaugh:
Resolves #563

PR #584, opened 2020-06-30
Upgrade Contour to 1.6.0

aj-jester:
@mzaneri @areed 

Hey guys -- got a request to upgrade Contour to 1.6.0 to support `pathRewritePolicy` which was added in `1.1.0` https://projectcontour.io/announcing-contour-1.1/

I tested this with `kubectl apply -k` on a `kURL` cluster manually and it applied fine.

Couple of things:

* This requires 2 different Contour tags `1.6.0` and `latest`, can you please check the `Manifest` file to see if this has been properly defined?

* What is the best way to test this without merging? I made `1.6.0` _not_ the latest but is there a way to check if upgrade from `1.0.1` to `1.6.0` works? https://projectcontour.io/resources/upgrading/

### DIFF
--- a/addons/contour/1.6.0/Manifest
+++ b/addons/contour/1.6.0/Manifest
@@ -1,0 +1,3 @@
+image contour docker.io/projectcontour/contour:v1.6.0
+image contour docker.io/projectcontour/contour:latest
+image envoy docker.io/envoyproxy/envoy:v1.14.2
--- a/addons/contour/1.6.0/contour.yaml
+++ b/addons/contour/1.6.0/contour.yaml
@@ -1,0 +1,1430 @@
+---
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: contour
+  namespace: projectcontour
+---
+apiVersion: v1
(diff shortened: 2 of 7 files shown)


addons/contour/1.6.0/Manifest
```diff
@@ -0,0 +1,3 @@
+image contour docker.io/projectcontour/contour:v1.6.0
+image contour docker.io/projectcontour/contour:latest
```
laverya, line 2: why latest? nothing should use latest

aj-jester: @laverya Its because this is being used in a `Job`, see this comment.

https://github.com/replicatedhq/kURL/pull/584/files#diff-9d2f620150ba2477c9255aabc7acadd0R994-R997

marccampbell:
@aj-jester i think we need to skip straight to 1.7.0 due to https://github.com/projectcontour/contour/security/advisories/GHSA-mjp8-x484-pm3r

aj-jester:
For sure! Closing this one.

PR #592, opened 2020-07-06
add velero local bucket name and cluster ip to s3 secret

salahalsaleh:


addons/kotsadm/alpha/tmpl-secret-s3.yaml
```diff
@@ -9,3 +9,5 @@ stringData:
   endpoint: ${OBJECT_STORE_CLUSTER_HOST}
   access-key-id: ${OBJECT_STORE_ACCESS_KEY}
   secret-access-key: ${OBJECT_STORE_SECRET_KEY}
+  velero-local-bucket: ${VELERO_LOCAL_BUCKET}
+  cluster-ip: ${OBJECT_STORE_CLUSTER_IP}
```
divolgin, line 13: how about `object-store-cluster-ip`

salahalsaleh: i was hesitant, yep

PR #597, opened 2020-07-08
FIX broken re-runs of the install script on k8s 1.18.x

mzaneri:


scripts/install.sh
```diff
@@ -39,6 +39,9 @@ function init() {
         API_SERVICE_ADDRESS="$LOAD_BALANCER_ADDRESS:$LOAD_BALANCER_PORT"
     fi
 
+    #This is needed on k8s 1.18.x as $PRIVATE_ADDRESS is found to have a newline
+    PRIVATE_ADDRESS=$(echo "$PRIVATE_ADDRESS" | tr '\n' ' ')
```
areed, line 43: why replace with a space instead of the empty string?

mzaneri: tr: when not truncating set1, string2 must be non-empty

mzaneri: maybe I should also strip leading whitespace after too?

areed: yes all whitespace should be stripped

PR #600, opened 2020-07-13
Add airgap flag for airgapped kotsadm installs

divolgin:


addons/kotsadm/alpha/kotsadm-airgap.yaml
```diff
@@ -0,0 +1,12 @@
...
+metadata:
+  name: kotsadm
+spec:
+  template:
+    spec:
+      containers:
+      - name: kotsadm
+        env:
+        - name: DISABLE_OUTBOUND_CONNECTIONS
+          value: true
```
marccampbell, line 12: `"` this i think

PR #607, opened 2020-07-15
use a better 'password not regenerated' message

laverya:
one that sounds less like an error


laverya:
@markpundsack what about `You can log in with your existing password. If you need to reset it, run ${GREEN}kubectl kots reset-password default${NC}`?

markpundsack:
@laverya That works for me.

PR #610, opened 2020-07-21
FIX containerd local registry cert fix

mzaneri:


scripts/common/containerd.sh
```diff
@@ -50,3 +42,25 @@ function install_containerd() {
       configure_containerd
    fi
 }
+
+function add_registry_to_containerd_config() {
+   local docker_registry_ip="$1"
+
+   if ! grep -q "ca_file" /etc/containerd/config.toml; then
+      sed -i "/registry-1/a \        [plugins.cri.registry.mirrors."\""registry.kurl.svc.cluster.local""\"]\n          endpoint = [""\"$1:443""\"]\n      [plugins.cri.registry.configs]\n        [plugins.cri.registry.configs."\""registry.kurl.svc.cluster.local""\".tls]\n          ca_file = "\""/etc/kubernetes/pki/ca.crt"\""" /etc/containerd/config.toml
```
mzaneri, line 50: this is probably brittle, but works until containerd allows for live cert adding

areed, line 50: It would help to have a comment of how this is supposed to look before and after since it's hard to follow what's going on with all the quotes.

areed: Also, why does a mirror have to be configured? Is that so images can be pulled/pushed from inside the cluster using the service DNS?

mzaneri: Comment added, and that its my understanding of it, I used the IP for testing

PR #613, opened 2020-07-22
ADD support to restrict nodeport range to kuberenetes standard

mzaneri:


mzaneri:
> 1. Are there any changes needed to the typescript server?
> 2. This looks like it's reverting the changes made to the generated code to be compatible with K8s 1.18. Was this generated with the latest binaries?

1. Is addressed now, thank you

2. I have 1.18.2 Client and 1.18.4 server where this was generated, I also used the update clien-gen and controller-gen that marc added to the make file to get mroe recent versions of each.

PR #624, opened 2020-07-30
wait for kotsadm pods to be created before checking their status

divolgin:


divolgin:
@laverya that's a good point.  but I really don't want to wait until Terminating pods are completely removed because that can take a long time.  Or there can be a pod stuck in that state perpetually.   In case of upgrades, the UI will still be available.  This just takes care of new installs when UI is not yet accessible because 0 pods are running.

PR #627, opened 2020-08-04
Containerd registry ca

areed:
To set the CA for a registry containerd 1.3 is required. The .deb and .rpm packages don't have 1.3. This PR switches to installing containerd as an add-on by packaging the containerd and runc binaries along with a systemd service. Only the 1.3.7 add-on version configures the in-cluster registry.


addons/containerd/1.3.7/install.sh
```diff
@@ -0,0 +1,98 @@
...
+    if [ -s "/etc/containerd/config.toml" ]; then
+        return 0
+    fi
+
+    sleep 1
+
+    mkdir -p /etc/containerd
+    containerd config default > /etc/containerd/config.toml
+
+    sed -i '/systemd_cgroup/d' /etc/containerd/config.toml
```
areed, line 56: Rest of the file from this line on differs from 1.2.13.

scripts/install.sh
```diff
@@ -134,7 +133,9 @@ EOF
     kubectl cluster-info
     logSuccess "Cluster Initialized"
 
-    containerd_registry_init
+    if commandExists containerd_registry_init; then
```
areed, line 136: Undefined if using docker or containerd 1.2.13.

addons/containerd/1.3.7/install.sh
```diff
@@ -0,0 +1,98 @@
...
+    systemctl enable containerd.service
+    systemctl start containerd.service
+}
+
+function containerd_configure() {
+    if [ -s "/etc/containerd/config.toml" ]; then
+        return 0
+    fi
+
+    sleep 1
```
divolgin, line 51: is this needed?  should this be a loop that checks for some condition?

addons/registry/2.7.1/install.sh
```diff
@@ -78,8 +78,10 @@ function registry_docker_ca() {
         bail "Docker registry address required"
     fi
 
-    mkdir -p /etc/docker/certs.d/$DOCKER_REGISTRY_IP
-    ln -s --force /etc/kubernetes/pki/ca.crt /etc/docker/certs.d/$DOCKER_REGISTRY_IP/ca.crt
+    if [ -n "$DOCKER_VERSION" ]; then
```
divolgin, line 81: would `DOCKER_REGISTRY_IP` be a better check here?

addons/containerd/1.3.7/install.sh
areed: I was wondering the same. I'll remove it and we can discover in QA if it was needed.

addons/registry/2.7.1/install.sh
areed: That variable is set when using the registry with containerd too. I think `DOCKER_VERSION` is what we want here since that will only be set when using docker for the container runtime.

PR #639, opened 2020-08-12
Fix krew directory permissions

areed:


scripts/common/kubernetes.sh
```diff
@@ -275,10 +275,25 @@ function install_krew() {
     ./krew-linux_amd64 install --manifest=support-bundle.yaml --archive=support-bundle.tar.gz > /dev/null 2>&1
     popd
 
-    # Fixes permission issues with 'kubectl krew'
-    chmod -R 0777 /opt/replicated/krew/store
-    chmod -R a+rw /opt/replicated/krew
-    chmod -R a+rw /tmp/krew-downloads
+    # The local krew store is owned by root so commands such as `kubectl krew update` that write to
```
MikaelSmith, line 278: Why not just use the absolute path? Is there someone running `kubectl krew update` that doesn't know where it lives?

MikaelSmith: Modifying `/etc/sudoers` seems like an unnecessary complication.

MikaelSmith: I guess this is all so someone administering the system can update krew themselves in the normal fashion, but krew has to be usable by root (and thus you want it owned by root) because kurl uses it during setup.

PR #643, opened 2020-08-13
Fixing missing graphs in kotsadm

divolgin:


divolgin:
good point... that would explain why it was in a patch

PR #647, opened 2020-08-17
Add contour 1.7.0

areed:


addons/contour/1.7.0/contour.yaml
```diff
@@ -0,0 +1,1796 @@
...
+      labels:
+        app: "contour-certgen"
+    spec:
+      containers:
+      - name: contour
+        # This version is set to latest because Job specs are immutable;
+        # if we change this on each version, you can no longer upgrade
+        # just by applying the deployment YAML.
+        # See #2423, #2395, #2150, and #2030 for earlier questions about this.
+        image: docker.io/projectcontour/contour:latest
```
aj-jester, line 1391: @areed Do we need to add this to `Manifest`? How does Contour know to pull the `latest`?

areed: We are setting this is a patch to use the v1.7.0 tag

PR #650, opened 2020-08-18
match kotsadm-authstring secret format with kotsadm

divolgin:


addons/kotsadm/1.18.1/install.sh
```diff
@@ -124,11 +124,17 @@ function kotsadm_secret_authstring() {
     local AUTHSTRING=$(kubernetes_secret_value default kotsadm-authstring kotsadm-authstring)
 
     if [ -n "$AUTHSTRING" ]; then
-        return 0
-    else
-        AUTHSTRING=$(< /dev/urandom tr -dc A-Za-z0-9 | head -c16)
+        # These are the only two valid formats.  Regenerating token in other cases to fix existing installs.
+        if [ $AUTHSTRING == "Kots *" ]; then
```
areed, line 128: I think this would fail if it's a match since AUTHSTRING would be multi-word and isn't quoted.

divolgin: Indeed.  My test install was still running when I made the PR.  I'll update.

PR #656, opened 2020-08-26
Bootstrapping moved to Go API

divolgin:


addons/kotsadm/alpha/install.sh
```diff
@@ -55,14 +52,18 @@ function kotsadm() {
     kubectl delete deployment kotsadm-web || true; # replaced by 'kotsadm' deployment in 1.12.0
     kubectl delete service kotsadm-api || true; # replaced by 'kotsadm-api-node' service in 1.12.0
 
-    kotsadm_namespaces "$src" "$dst"
```
laverya, line 58: is this function no longer needed? what did it do?

divolgin: removed by accident.  added back.  thanks.

PR #661, opened 2020-09-02
set src variable when running velero join

laverya:
otherwise in online /assets is created, and in airgap it fails

(the normal `velero` function, run on first master, does set `src` - this copies that)
https://github.com/replicatedhq/kURL/pull/661/files#diff-a6784e5b3c7e2994442eeec2f29dbcfeL12

### DIFF
--- a/addons/velero/1.2.0/install.sh
+++ b/addons/velero/1.2.0/install.sh
@@ -39,6 +39,8 @@ function velero() {
 }
 
 function velero_join() {
+    local src="$DIR/addons/velero/1.2.0"
+
     velero_binary
 }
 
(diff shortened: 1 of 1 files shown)


areed (commented):
It looks like it needs to be set at the top of the velero_binary function

laverya:
maybe, but it's used elsewhere
I'd rather set it in the calling function in case we eventually add more calls there

areed:
What about passing it as an argument to velero_binary then? We don't intentionally rely on the behavior of local vars getting passed to callees anywhere in the project.

laverya:
That would make sense
Or a "use if set, otherwise provide default" if/else in the velero_binary function

areed (approved):

PR #662, opened 2020-09-02
rook 1.4.1: Add support for running ceph commands in ceph-operator

yasserg:
This is needed for ekc operator to work properly.

### DIFF
--- a/addons/rook/1.4.1/operator/patches/ceph-operator-init.yaml
+++ b/addons/rook/1.4.1/operator/patches/ceph-operator-init.yaml
@@ -1,0 +1,37 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: rook-ceph-operator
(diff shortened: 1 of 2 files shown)


addons/rook/1.4.1/operator/patches/ceph-operator-init.yaml
```diff
@@ -0,0 +1,37 @@
...
+                  key: ceph-secret
+          volumeMounts:
+            - mountPath: /etc/ceph
+              name: default-config-dir
+            - name: mon-endpoint-volume
+              mountPath: /etc/rook
+      volumes:
+        - name: mon-endpoint-volume
+          configMap:
+            name: rook-ceph-mon-endpoints
```
areed, line 34: If rook moves mons do the endpoints change? Do the ceph commands continue to work afterwards?

PR #674, opened 2020-09-10
Upgrading Weave add-on version to 2.7.0

paha:
RE: https://github.com/weaveworks/weave/releases

Adding missing params for weave-npc pod
`dnsPolicy: ClusterFirstWithHostNet`
and
`priorityClassName: system-node-critical`


areed (commented):
I see in the weave yaml it sets `dnsPolicy: ClusterFirstWithHostNet` and `priorityClassName: system-node-critical`. Should we include those in ours?

areed (approved):
LGTM. I'd wait to merge until after the release.

paha:
@areed thanks.
Yes, both should be present. Overall Weave CNI functionality should not be impacted by them, so it might be optional for us to update the previous versions. What do you think?

areed:
Agreed, no need to risk updating previous versions.

PR #677, opened 2020-09-15
Upgrade OpenEBS to v1.12.0

paha:


addons/openebs/1.12.0/Manifest
```diff
@@ -0,0 +1,15 @@
+image admission-server openebs/admission-server:1.6.0
```
areed, line 1: These images need to be updated

areed (approved):

PR #687, opened 2020-09-17
Include node join command expiry in command output

laverya:


divolgin (approved):

scripts/install.sh
```diff
@@ -214,6 +214,9 @@ function outro() {
         printf "These settings will automatically be applied on the next restart."
         printf "\n"
     fi
+    printf "\n"
+    printf "\n"
+    printf "Node join commands expire after two hours. You can rerun this script on any master node to generate fresh node join commands.\n"
```
markpundsack, line 219: Assuming it's printed right after the script.

```suggestion
    printf "Note: The above script expires in 2 hours. Rerun the kURL install script on any master node to generate a fresh script.\n"
```

laverya: currently the node join scripts are printed below this
but I get the idea

markpundsack: Is it *right* below it? If the scripts are too far removed from this message, it'll be harder to comprehend. They should be close together.

laverya: immediately below it
the output would look like:
```
Node join commands expire after two hours. You can rerun this script on any master node to generate fresh node join commands.

To add worker nodes to this installation, copy and unpack this bundle on your other nodes, and run the following:


${GREEN}    cat ./join.sh | sudo bash -s airgap kubernetes-master-address=${API_SERVICE_ADDRESS} kubeadm-token=${BOOTSTRAP_TOKEN} kubeadm-token-ca-hash=${KUBEADM_TOKEN_CA_HASH} kubernetes-version=${KUBERNETES_VERSION}${dockerRegistryIP}${noProxyAddrs}
```
etc

markpundsack: OK. I'd still prefer to be explicit with "the kURL install script" rather than the ambiguous "this script".

laverya: done

markpundsack (approved):

PR #693, opened 2020-09-17
test environment step by step instructions

paha:
Wondering if this has value and what is a good place for this help file. In this form, it is an extended version of the existing  https://kurl.sh/docs/community/core-contributions doc we have. I might add a link in that doc to this file as was previously suggested.

More time is needed to set up a more robust local testing. I was experimenting with Docker in docker, Kind + KubeVirt, and Qemu, but that is not complete yet. Also, was experimenting in decoupling API server component, and potentially object store, for both to run locally in containers, will get to that at some point.


test/dev.md
```diff
@@ -0,0 +1,176 @@
...
+    KURL_URL=
+    DIST_URL=
+    INSTALLER_ID=
+    REPLICATED_APP_URL=https://replicated.app
+    KURL_UTIL_IMAGE=replicated/kurl-util:alpha
+    KURL_BIN_UTILS_FILE=
+    INSTALLER_YAML="apiVersion: cluster.kurl.sh/v1beta1
+    kind: Installer
+    metadata:
+    name: latest
```
areed, line 36: need indentation

areed (approved):
I would remove everything else in the test directory if you're going to add this here. Nothing else in there works.

PR #704, opened 2020-09-22
add a task to generate new node join commands

laverya:


scripts/install.sh
```diff
@@ -216,7 +216,30 @@ function outro() {
...
+    local prefix="curl -sSL${proxyFlag} $KURL_URL/$INSTALLER_ID/"
+    if [ -z "$KURL_URL" ]; then
+        prefix="cat "
+    fi
+
+    if [ "$HA_CLUSTER" = "1" ]; then
+        printf "Master node join commands expire after two hours, and worker node join commands expire after 24 hours.\n"
+        printf "\n"
+        if [ "$AIRGAP" = "1" ]; then
+            printf "To generate new node join commands, run ${GREEN}cat ./tasks.sh | sudo bash -s join_token ha airgap${NC} on an existing master node.\n"
```
markpundsack, line 229: Curious, but why the inconsistent use of ${prefix}?

laverya: I copied it from the airgap vs online install scripts below this

areed (commented):
Could you get no proxy from local docker or somewhere else and pass that?

areed (approved):

PR #705, opened 2020-09-22
use the current weave tag to remove weave in reset script

laverya:
if the weave tag is not found, assume 2.7.0

### DIFF
--- a/scripts/common/tasks.sh
+++ b/scripts/common/tasks.sh
@@ -149,9 +149,12 @@ function weave_reset() {
     CONTAINER_IFNAME=ethwe
 
     if [ -z "$WEAVE_TAG" ]; then
-        # if we don't know the exact weave tag, use a sane default
-        export WEAVE_TAG="2.7.0"
-        printf "using default weave tag ${WEAVE_TAG}\n"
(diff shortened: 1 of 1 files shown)


areed (commented):
I would also add a fallback to `docker image ls | grep weaveexec` since that image is loaded anywhere the installer runs and kubectl will only be successful on primaries.

laverya:
I'll search for weaveworks/weave-npc - I don't think we have weaveexec by default

scripts/common/tasks.sh
```diff
@@ -145,7 +148,15 @@ function weave_reset() {
     DATAPATH=datapath
     CONTAINER_IFNAME=ethwe
 
-    WEAVE_TAG=${WEAVE_VERSION}
+    if [ -z "$WEAVE_TAG" ]; then
+        WEAVE_TAG=$(docker image ls | grep weaveworks/weave-npc | awk '{ print $2 }')
```
areed, line 152:
```suggestion
        WEAVE_TAG=$(docker image ls | grep weaveworks/weave-npc | awk '{ print $2 }' | head -1)
```
In case it's an upgrade

areed (approved):

PR #717, opened 2020-09-26
Workaround for nftables race condition

areed:
I had a CentOS 8 install where weave 2.7.0 was crashlooping with error `ipset v7.2: Set cannot be destroyed: it is in use by a kernel component`

[This workaround](https://github.com/weaveworks/weave/issues/3816#issuecomment-640474405) allowed weave to come up.

An alternative workaround is to configure weave to use legacy iptables instead of nftables.

### DIFF
--- a/addons/weave/2.6.5/daemonset.yaml
+++ b/addons/weave/2.6.5/daemonset.yaml
@@ -21,7 +21,9 @@ spec:
       containers:
         - name: weave
           command:
-            - /home/weave/launch.sh
+            - /bin/sh
+            - -c
+            - sed '/ipset destroy weave-kube-test$/ i sleep 1' /home/weave/launch.sh | /bin/sh
           env:
             - name: HOSTNAME
               valueFrom:
--- a/addons/weave/2.7.0/daemonset.yaml
+++ b/addons/weave/2.7.0/daemonset.yaml
@@ -21,7 +21,9 @@ spec:
       containers:
         - name: weave
           command:
-            - /home/weave/launch.sh
+            - /bin/sh
+            - -c
+            - sed '/ipset destroy weave-kube-test$/ i sleep 1' /home/weave/launch.sh | /bin/sh
           env:
             - name: HOSTNAME
               valueFrom:


divolgin (approved):

kavyabgowda:
**Even I'm facing the same issue after using the updated yaml.**

I was facing this issue when i used https://cloud.weave.works/k8s/net?k8s-version=Q2xpZW50IFZlcnNpb246IHZlcnNpb24uSW5mb3tNYWpvcjoiMSIsIE1pbm9yOiIxOSIsIEdpdFZlcnNpb246InYxLjE5LjQiLCBHaXRDb21taXQ6ImQzNjA0NTRjOWJjZDE2MzRjZjRjYzUyZDE4NjdhZjU0OTFkYzljNWYiLCBHaXRUcmVlU3RhdGU6ImNsZWFuIiwgQnVpbGREYXRlOiIyMDIwLTExLTExVDEzOjE3OjE3WiIsIEdvVmVyc2lvbjoiZ28xLjE1LjIiLCBDb21waWxlcjoiZ2MiLCBQbGF0Zm9ybToibGludXgvYW1kNjQifQpTZXJ2ZXIgVmVyc2lvbjogdmVyc2lvbi5JbmZve01ham9yOiIxIiwgTWlub3I6IjE5IiwgR2l0VmVyc2lvbjoidjEuMTkuNCIsIEdpdENvbW1pdDoiZDM2MDQ1NGM5YmNkMTYzNGNmNGNjNTJkMTg2N2FmNTQ5MWRjOWM1ZiIsIEdpdFRyZWVTdGF0ZToiY2xlYW4iLCBCdWlsZERhdGU6IjIwMjAtMTEtMTFUMTM6MDk6MTdaIiwgR29WZXJzaW9uOiJnbzEuMTUuMiIsIENvbXBpbGVyOiJnYyIsIFBsYXRmb3JtOiJsaW51eC9hbWQ2NCJ9Cg==
config
**ipset v7.2: Set cannot be destroyed: it is in use by a kernel component**

Before editing the deamonset file to include these changes only worker node weave pod was crashing. But now both master and worker is crashing after running 1 pod itself.

I tried everything in this issue thread. But nothing worked.

PR #721, opened 2020-09-29
initial GH action for testgrid runs

paha:


.github/workflows/daily-testgrid.yaml
```diff
@@ -0,0 +1,43 @@
...
+    working-directory: ./testgrid/tgrun
+
+jobs:
+  testgrid-init:
+    runs-on: ubuntu-latest
+    steps:
+    - name: Checkout
+      uses: actions/checkout@v2
+
+    - name: kubeconfig
```
marccampbell, line 17: Why is this needed?  How will this work when there are > testgrid servers/

.github/workflows/daily-testgrid.yaml
```diff
@@ -0,0 +1,43 @@
...
+        go-version: 1.15.1
+
+    - name: tgrun-build
+      run: make build
+
+    - name: testgrid-cleanup
+      env:
+        KUBECONFIG: "/tmp/config"
+      run:
+        kubectl delete vmi --all
```
marccampbell, line 35: This should be on the instance, not in the action. When we run this, we will need to have a dynamic set of more than 1 instance running these. It will be difficult to maintain this list in the action, but instead, the instance can delete them,.

.github/workflows/daily-testgrid.yaml
```diff
@@ -0,0 +1,43 @@
...
+
+    - name: testgrid-queue
+      env:
+        KUBECONFIG: "/tmp/config"
+        TESTGRID_API: "https://api.testgrid.staging.kurl.sh/"
+      
+      run: |
+        ./bin/tgrun queue --ref $(date "+%s") --testgrid-api $TESTGRID_API
```
marccampbell, line 43: What if we used the last 7 of the sha here?

.github/workflows/deploy-testgrid.yaml
```diff
@@ -16,6 +16,12 @@ jobs:
           --build-arg version=${GITHUB_SHA:0:7} \
           -t 923411875752.dkr.ecr.us-east-1.amazonaws.com/tgapi:${GITHUB_SHA:0:7} testgrid/tgapi
 
+    - run: |
+        docker build -f testgrid/tgrun/Dockerfile \
+          --build-arg version=${GITHUB_SHA:0:7} \
+          -t 923411875752.dkr.ecr.us-east-1.amazonaws.com/tgrun:latest \
+          -t 923411875752.dkr.ecr.us-east-1.amazonaws.com/tgrun:${GITHUB_SHA:0:7} testgrid/tgrun
```
marccampbell, line 23: testgrid should be public, can this be in Dockerhub instead of ecr?

marccampbell (changes requested):

.github/workflows/daily-testgrid.yaml
paha: Good points,
I was intending to add `vmi` cleanup to the *main loop* of `tgrun` left it here until then, but yes at this point it is confusing and multiple tg clusters would be hard to support with this model.

removing this in the next commit and will get clean up managed by `tgrun`.

.github/workflows/daily-testgrid.yaml
paha: Agreed, please see the previous comment on this.

.github/workflows/daily-testgrid.yaml
paha: Sounds good, the time/date reference is carried by other metadata. Changing in the next commit.

.github/workflows/deploy-testgrid.yaml
```diff
@@ -86,7 +87,26 @@ jobs:
...
+          docker build -f testgrid/tgrun/Dockerfile \
+            -t replicated/tgrun:latest \
+            -t replicated/tgrun:${GITHUB_SHA:0:7} replicated/tgrun
+
+      - name: tgrun-push
+        env:
+          DOCKERHUB_DOCKER_PASS: ${{ secrets.DOCKERHUB_DOCKER_PASS }}
+          DOCKERHUB_DOCKER_USER: ${{ secrets.DOCKERHUB_DOCKER_USER }}
+        run: |
+           echo "$DOCKERHUB_DOCKER_PASS" | docker login -u $DOCKERHUB_DOCKER_USER --password-stdin
```
marccampbell, line 108: https://github.com/replicatedhq/troubleshoot/blob/master/.github/workflows/build-test-deploy.yaml#L105

Better to use the docker login action for this please

.github/workflows/deploy-testgrid.yaml
```diff
@@ -86,7 +87,26 @@ jobs:
...
+           docker push replicated/tgrun
```
marccampbell, line 109: no tag?

marccampbell (changes requested):

testgrid/tgrun/deploy/tgrun-pod.yaml
```diff
@@ -0,0 +1,12 @@
+apiVersion: v1
+kind: Pod
+metadata:
+  name: tgrun
+  labels:
+    app: testgrid
+spec:
+  containers:
+  - name: tgrun
+    image: replicated/tgrun
```
marccampbell, line 10: there should be a tag on this, it's going to be very hard to reproduce and troubleshoot without a tag

testgrid/tgrun/deploy/tgrun-pod.yaml
```diff
@@ -0,0 +1,12 @@
+apiVersion: v1
+kind: Pod
```
marccampbell, line 2: why not a deployment instead of a pod?

.github/workflows/daily-testgrid.yaml
```diff
@@ -24,7 +24,7 @@ jobs:
 
     - name: testgrid-queue
       env:
-        TESTGRID_API: "https://api.testgrid.staging.kurl.sh/"
```
marccampbell: we should run these 7 days a week ^^ the cron spec is 1-5 but images might change over the weekend

laverya (approved):

PR #723, opened 2020-09-30
Containerd http proxy support

emosbaugh:


addons/containerd/1.3.7/install.sh
```diff
@@ -3,13 +3,16 @@
...
+    if [ "$SKIP_CONTAINERD_INSTALL" != "1" ]; then
+        containerd_binaries "$src"
+        containerd_configure
+        containerd_service "$src"
     fi
 
-    containerd_binaries "$src"
-    containerd_configure
-    containerd_service "$src"
+    # NOTE: this will not remove the proxy
```
emosbaugh, line 12: I copied the logic here from docker_configure_proxy. Is this desired or should we remove the proxy when not configured?

areed (approved):

PR #727, opened 2020-10-01
Configure crictl

### LINKED ISSUE #726: crictl not configured
When using the containerd addon, crictl is not configured and times out attempting to run any commands.

I fixed this by manually adding `/etc/crictl.yaml`
```
runtime-endpoint: /var/run/containerd/containerd.sock
image-endpoint: /var/run/containerd/containerd.sock
```

This seems like something the addon should setup.

emosbaugh:
Resolves #726

### DIFF
--- a/addons/containerd/1.3.7/crictl.yaml
+++ b/addons/containerd/1.3.7/crictl.yaml
@@ -1,0 +1,2 @@
+runtime-endpoint: /var/run/containerd/containerd.sock
+image-endpoint: /var/run/containerd/containerd.sock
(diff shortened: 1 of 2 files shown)


addons/containerd/1.3.7/crictl.yaml
```diff
@@ -0,0 +1,2 @@
+runtime-endpoint: /var/run/containerd/containerd.sock
```
areed, line 1: This file isn't used

areed (approved):

addons/containerd/1.3.7/crictl.yaml
emosbaugh: haha, oops

areed (approved):

PR #761, opened 2020-10-09
Convert non-HA to HA

areed:
Detect when adding a load balancer to a single or multi-node cluster and prompt user to run a task to update the kubelet's kubeconfig on all remote nodes to point at the new load balancer.

Also make the tasks.sh script a template like the other top level scripts.


scripts/install.sh
```diff
@@ -103,13 +100,19 @@ EOF
...
     # lead to a loss of quorum
     disable_rook_ceph_operator
+
     # since K8s 1.19.1 kubeconfigs point to local API server even in HA setup. When upgrading from
     # earlier versions and using a load balancer, kubeadm init will bail because the kubeconfigs
     # already exist pointing to the load balancer
     rm -f /etc/kubernetes/*.conf
+
+    # Regenerate api server cert in case load balancer address changed
+    rm -f /etc/kubernetes/pki/apiserver.crt /etc/kubernetes/pki/apiserver.key
```
emosbaugh, line 114: any reason to mv these rather than rm?

emosbaugh (approved):

scripts/install.sh
areed: good idea

PR #770, opened 2020-10-13
Cleanup function and cmd

paha:
- identifies PVs in Terminating state, and clears finalizers if the state is older then 12 hours

### DIFF
--- a/testgrid/tgrun/pkg/runner/loop.go
+++ b/testgrid/tgrun/pkg/runner/loop.go
@@ -172,7 +172,7 @@ func CleanUp() error {
 
 	clientset, err := GetClientset()
 	if err != nil {
-		return errors.Wrap(err, "failed to get clientset during cleanup")
+		return errors.Wrap(err, "failed to get clientset")
 	}
 
 	// PV finalizer kubernetes.io/pv-protection fails to clean up pv/pvc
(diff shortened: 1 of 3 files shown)


testgrid/tgrun/pkg/runner/loop.go
```diff
@@ -158,3 +164,33 @@ func homeDir() string {
...
+
+	// PV finalizer kubernetes.io/pv-protection fails to clean up pv/pvc
+	// Removing the finalizer on stale pvs
+	pvs, err := clientset.CoreV1().PersistentVolumes().List(ctx(), metav1.ListOptions{})
+	if err != nil {
+		return errors.Wrap(err, "failed to get pv list")
+	}
+
+	for _, pv := range pvs.Items {
+		// selecting persistent volumes marked for deletion over 12 hours ago.
```
laverya, line 186: Could we do something more like _1_ hour ago?

testgrid/tgrun/pkg/runner/loop.go
```diff
@@ -158,3 +164,33 @@ func homeDir() string {
...
 }
+
+// CleanUp deletes finalizers on 'stack' in Terminating pvs
+// localPath is cleared as PV completes termination.
+func CleanUp() error {
+	var ctx = context.TODO
+
+	clientset, err := GetClientset()
+	if err != nil {
+		return errors.Wrap(err, "failed to get clientset during cleanup")
```
emosbaugh, line 175: remove "during cleanup"

emosbaugh (approved):

PR #774, opened 2020-10-13
Testgrid ubuntu 20 support

emosbaugh:


testgrid/tgrun/pkg/instances/k8s1164.go
```diff
@@ -22,7 +22,7 @@ func init() {
 				Version: "1.0.1",
 			},
 			Docker: &kurlv1beta1.Docker{
-				Version: "19.03.4",
+				Version: "19.03.10",
```
laverya, line 25: can we leave this with old docker? Or make a new test with old docker specifically?

laverya (approved):

testgrid/tgrun/pkg/instances/k8s1164.go
emosbaugh: i think we should make a test with old docker specifically. ill add that one

PR #785, opened 2020-10-14
Fix load balancer address change detection

areed:
Re-running with the `ha` flag and existing load-balancer address was prompting user to update address on remote nodes.

### DIFF
--- a/scripts/install.sh
+++ b/scripts/install.sh
@@ -32,7 +32,7 @@ function init() {
         API_SERVICE_ADDRESS="$LOAD_BALANCER_ADDRESS:$LOAD_BALANCER_PORT"
     fi
 
-    OLD_LOAD_BALANCER_ADDRESS=$(kubernetes_load_balancer_address)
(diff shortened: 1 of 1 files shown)


scripts/install.sh
```diff
@@ -32,7 +32,7 @@ function init() {
         API_SERVICE_ADDRESS="$LOAD_BALANCER_ADDRESS:$LOAD_BALANCER_PORT"
     fi
 
-    OLD_LOAD_BALANCER_ADDRESS=$(kubernetes_load_balancer_address)
+    local oldLoadBalancerAddress=$(kubernetes_load_balancer_address)
```
areed, line 35: Read in https://github.com/replicatedhq/kURL/pull/785/files#diff-e076e43cb817ecc8b3c2ed18cd58baa50544ecc66293ed6e61129c43ca6a09ccR139

emosbaugh (approved):

PR #801, opened 2020-10-21
Change load balancer on HA cluster

areed:


addons/ekco/0.6.0/regen-cert-pod.yaml
```diff
@@ -0,0 +1,23 @@
+apiVersion: v1
```
areed, line 1: This file is new

laverya (approved):

PR #809, opened 2020-10-23
Configuration option for nameserver and preflight check for loopback nameserver

areed:


paha (approved):

scripts/common/preflights.sh
```diff
@@ -233,3 +234,15 @@ function kotsadm_prerelease() {
         fi
     fi
 }
+
+function host_nameservers_reachable() {
+    if [ "$AIRGAP" = 1 ]; then
+        return 0
+    fi
+    if ! discover_non_loopback_nameservers; then
+        printf "\n${YELLOW}No nameservers configured on the host are accessible by pods. All DNS queries outside the cluster will fail. Continue?${NC} "
```
marccampbell, line 243: @areed can you help me understand how i'd want to answer this?  if i'm installing an application (the primary kURL use case today), in an airgapped environment, what does this mean?  If I see this, should I continue or abort?

scripts/common/preflights.sh
```diff
@@ -233,3 +234,15 @@ function kotsadm_prerelease() {
...
+        if ! confirmN; then
+            bail "\nAt least one nameserver must be accessible on a non-loopback address"
```
marccampbell, line 245: "accessible".  Does this mean accessibility or configured?

marccampbell:
Is it possible to install a local DNS server on a server, configure resolv.conf to point to that and have the local DNS server be a cache through to upstream? Would this incorrectly warn that it's not set up properly?

Would a better solution be to test for the condition that we are looking for? This currently tests for the configuration and makes assumptions about what that means, when it might be easier or more reliable to test that DNS resolution works, using one of the known-open, documented Replicated endpoints?

scripts/common/preflights.sh
areed: In an airgapped environment this preflight is skipped. In an online environment you should abort and fix your nameserver.

scripts/common/preflights.sh
areed: The address in resolv.conf must be non-loopback. It's about what address the client uses not what address the nameserver binds.

areed:
> Is it possible to install a local DNS server on a server, configure resolv.conf to point to that and have the local DNS server be a cache through to upstream? Would this incorrectly warn that it's not set up properly?

Yes, you can run dnsmasq but if that's your only nameserver in resolv.conf it has to be on the host IP, not loopback.

areed:
> Would a better solution be to test for the condition that we are looking for? This currently tests for the configuration and makes assumptions about what that means, when it might be easier or more reliable to test that DNS resolution works, using one of the known-open, documented Replicated endpoints?

We could do that but then it wouldn't be a preflight, we'd have to wait for kubernetes to be installed

laverya (approved):

testgrid/tgrun/pkg/instances/k8s119_nameserver.go
```diff
@@ -0,0 +1,30 @@
...
+		types.Instance{
+			InstallerSpec: types.InstallerSpec{
+				Kubernetes: kurlv1beta1.Kubernetes{
+					Version: "1.19.3",
+				},
+				Weave: &kurlv1beta1.Weave{
+					Version: "2.6.4",
+				},
+				Rook: &kurlv1beta1.Rook{
+					Version: "1.0.4",
```
laverya, line 19: do we need rook here?
just want to keep tests minimal (and thus theoretically faster), and rook 1.0.4 is definitely tested elsewhere

laverya (approved):

testgrid/tgrun/pkg/instances/k8s119_nameserver.go
areed: No, I'll remove

PR #810, opened 2020-10-23
Add Metrics server 0.3.7

nikogura:
I attempted to follow the instructions here: https://kurl.sh/docs/add-on-author/

Apologies in advance if I missed something.

The file components.yaml comes from https://github.com/kubernetes-sigs/metrics-server/releases/tag/v0.3.7

Thank you!

### DIFF
--- a/addons/metrics-server/0.3.7/components.yaml
+++ b/addons/metrics-server/0.3.7/components.yaml
@@ -1,0 +1,150 @@
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: system:aggregated-metrics-reader
+  labels:
+    rbac.authorization.k8s.io/aggregate-to-view: "true"
+    rbac.authorization.k8s.io/aggregate-to-edit: "true"
+    rbac.authorization.k8s.io/aggregate-to-admin: "true"
+rules:
+- apiGroups: ["metrics.k8s.io"]
+  resources: ["pods", "nodes"]
+  verbs: ["get", "list", "watch"]
+---
+apiVersion: rbac.authorization.k8s.io/v1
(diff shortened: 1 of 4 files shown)


jesserobbins:
@grantmiller I would recommend modifying your cla-generator configuration. It would be better to expand it to allow for CCLAs with individual contributor authorizations to avoid us having to provide this kind of note. ;-) In this case, Orion Labs, Inc. is authorizing its employee @nikogura to make the contribution contained in this specific pull request and to agree to the CLA for this commit.

laverya (approved):

addons/metrics-server/0.3.7/components.yaml
```diff
@@ -0,0 +1,150 @@
...
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: Role
+  name: extension-apiserver-authentication-reader
+subjects:
+- kind: ServiceAccount
+  name: metrics-server
+  namespace: kube-system
+---
+apiVersion: apiregistration.k8s.io/v1beta1
```
areed, line 42: Could this use v1? It's supported in K8s 1.15.

nikogura: That's a good question.  What you see here is directly from the Metrics Server release.

How far do you want to go in maintaining a downstream fork, vs incorporating their release verbatim?

areed: We try to change as little as possible.  In this case I think it's worth modifying because v1beta1 is deprecated in 1.19 so it would print warnings during installation.

nikogura: Sounds good to me.  Updated.

areed (approved):

PR #816, opened 2020-10-28
include appslug in vanity url installers

laverya:


web/src/controllers/InstallerAPI.ts
```diff
@@ -261,3 +231,64 @@ export class Installers {
...
+  let i: Installer;
+  try {
+    i = Installer.parse(request.body, teamID);
+  } catch(error) {
+    response.status(400);
+    return { error };
+  }
+  i.id = id;
+
+  if (i.spec.kotsadm && (i.spec.kotsadm.applicationSlug == undefined || i.spec.kotsadm.applicationSlug == "")) {
```
salahalsaleh, line 277:
```suggestion
  if (i.spec.kotsadm && !i.spec.kotsadm.applicationSlug) {
```

salahalsaleh (approved):
lgtm, added a suggestion

web/src/controllers/InstallerAPI.ts
laverya: I'll use this if I go this way instead of fixing things upstream

PR #859, opened 2020-11-04
always apply secrets in order to update labels/annotations

salahalsaleh:


addons/kotsadm/alpha/install.sh
```diff
@@ -123,33 +121,23 @@ function kotsadm_secret_cluster_token() {
...
 
-    if [ -n "$AUTHSTRING" ]; then
-        # These are the only two valid formats.  Regenerating token in other cases to fix existing installs.
-        if [[ "$AUTHSTRING" =~ ^'Kots ' ]]; then
-            return 0
-        fi
-        if [[ "$AUTHSTRING" =~ ^'Bearer ' ]]; then
-            return 0
-        fi
+    if [[ -z "$AUTHSTRING" || (! "$AUTHSTRING" =~ ^'Kots ' && ! "$AUTHSTRING" =~ ^'Bearer ') ]]; then
```
areed, line 124: I'm not sure about these order of operations. Is there another way to write this?

salahalsaleh: sure, did test this will multiple scenarios FWIW

areed (approved):

areed (approved):

PR #869, opened 2020-11-06
automatically include 'ha' flag if multimaster

laverya:


scripts/common/utilbinaries.sh
```diff
@@ -209,3 +214,10 @@ function apply_iptables_config() {
     fi
     CONFIGURE_IPTABLES_SCRIPT=$CONFIGURE_IPTABLES_SCRIPT $BIN_SYSTEM_CONFIG -c iptables -e -y $MERGED_YAML_SPEC
 }
+
+function is_ha() {
+    local master_count=$(kubectl get node --selector='node-role.kubernetes.io/master' | grep 'master' | wc -l) #get nodes with the 'master' role, and then search for 'master' to remove the column labels row
```
emosbaugh, line 219: this doesnt seem like it works for second master... what are you trying to solve for here?

laverya: This is trying to determine if you're rerunning install.sh on an existing multimaster cluster, but forgot the `ha` flag

emosbaugh (approved):

PR #905, opened 2020-11-13
Bail if Docker unsupported OS

emosbaugh:


laverya (approved):

laverya (approved):

laverya:
Before we merge this - docker is working on [centos 8.2 in testgrid](https://testgrid.kurl.sh/run/6ca1707-onmerge-staging?kurlLogsInstanceId=cvqars), and we should investigate if it's actually broken in reality

And if we should perhaps just log something instead of failing, or require a prompt

scripts/common/preflights.sh
```diff
@@ -81,6 +82,28 @@ checkDockerK8sVersion()
...
+
+    case "$LSB_DIST" in
+    centos|rhel)
+        if [[ "$DIST_VERSION" =~ ^8 ]]; then
+            logWarn "Docker is not supported on ${LSB_DIST} ${DIST_VERSION}."
+            logWarn "The containerd addon is recommended. https://kurl.sh/docs/add-ons/containerd"
+            if ! commandExists "docker" ; then
+                printf "${YELLOW}Continue? ${NC}" 1>&2
+                if ! confirmY "-t 30"; then
+                    exit 1
```
laverya, line 99: does confirmY bail if you don't respond? Or does it continue instead?

laverya (approved):

scripts/common/preflights.sh
laverya: continues by default

PR #927, opened 2020-11-19
tgrun to provide feedback on reasons of not scheduling new jobs

paha:


testgrid/tgrun/pkg/runner/loop.go
```diff
@@ -120,6 +120,7 @@ func canScheduleNewVM() (bool, error) {
 	// if there are pending pods, hold off until there are no longer pending pods
 	for _, pod := range pods.Items {
 		if pod.Status.Phase == v1.PodPending {
+			fmt.Printf("Pod %s in (%s namespace) is in pending state. New jobs won't be scheduled.", pod.Name, Namespace)
```
laverya, line 123: I'm worried - this happens in a loop
will we fill up the logs with this every 5s or so? The majority of the lifetime of a testgrid instance will be spent in this state, after all (or at least a large fraction)

paha: It is a judgment call, I felt it is worth it. The loop is every 15 seconds, the message only will occur when there are pending jobs/things in a queue to run, so it will not be very often. The benefit is that the service provides feedback on the reasons the queue is stack.

DanStough:
Closing because this is stale.

PR #932, opened 2020-11-21
Upgrade add-ons on remote nodes

areed:
This ensures that remote nodes are kept up to date with the installer when run on the first primary by prompting the user to run the `upgrade.sh` script on all remotes nodes when there may be changes. Any add-on that defines a `_join` function has will cause the prompt for the upgrade script to be shown. Additionally, a proxy will trigger the upgrade prompt since docker or containerd may need to be reconfigured.

The upgrade script now runs all `_join` functions from add-ons and ensures docker and containerd are configured to use a proxy and to push/pull images from the in-cluster registry.

### DIFF
--- a/addons/containerd/1.3.7/install.sh
+++ b/addons/containerd/1.3.7/install.sh
@@ -1,5 +1,4 @@
 
-
 function containerd_install() {
     local src="$DIR/addons/containerd/1.3.7"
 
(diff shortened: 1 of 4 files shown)


laverya (approved):
better than not updating other nodes, but I'd love it if we could only show the 'run on other nodes' prompt if a has-host-components addon or proxy info changed

PR #934, opened 2020-11-23
Use --disable-eviction on upgrade when supported

### LINKED ISSUE #817: Upgrade script gets stuck when pods use eviction policies
We have eviction policies in place for pods, including some - postgres - that aren't supposed to be evicted.
```
evicting pod "postgres-0"
error when evicting pod "postgres-0" (will retry after 5s): Cannot evict pod as it would violate the pod's disruption budget.
...
```

[Kubernetes 1.18](https://v1-18.docs.kubernetes.io/docs/setup/release/notes/) added `disable-eviction` as an option for `kubectl drain`. When upgrading to 1.18 or newer, `kubectl drain` should use that option to ignore PodDisruptionBudgets.

MikaelSmith:
When draining nodes for upgrade, uses `--disable-eviction=true` to evict pods that would otherwise not be allowed due to a PodDisruptionBudget. Allows upgrades to continue cleanly while using PodDisruptionBudgets for normal quality of service guarantees.

Only works with kubectl 1.18+, so it's only used when upgrading a Kubernetes 1.18 install.

Fixes #817.

### DIFF
--- a/scripts/common/kubernetes.sh
+++ b/scripts/common/kubernetes.sh
@@ -188,10 +188,16 @@ function spinner_kubernetes_api_stable() {
 }
 
 function kubernetes_drain() {
+    local evict=""
+    # Check whether existing tools support bypassing eviction protection
+    if [ "$KUBERNETES_CURRENT_VERSION_MINOR" -ge "18" ]; then
+        evict=--disable-eviction=true
+    fi
+
     kubectl drain "$1" \
         --delete-local-data \
         --ignore-daemonsets \
-        --force \
+        --force $evict \
         --grace-period=30 \
         --timeout=300s \
         --pod-selector 'app notin (rook-ceph-mon,rook-ceph-osd,rook-ceph-osd-prepare,rook-ceph-operator,rook-ceph-agent),k8s-app!=kube-dns' || true


scripts/common/kubernetes.sh
```diff
@@ -188,10 +188,16 @@ function spinner_kubernetes_api_stable() {
 }
 
 function kubernetes_drain() {
-    kubectl drain "$1" \
+    local evict=""
+    # Check whether existing tools support bypassing eviction protection
```
MikaelSmith, line 192: Would it hurt to switch the order for upgrades to first install the new kubectl version?

MikaelSmith: Not sure if there's a specific reason to perform drain and `kubeadm upgrade` with the old version.

MikaelSmith: https://v1-18.docs.kubernetes.io/docs/tasks/administer-cluster/kubeadm/kubeadm-upgrade/#upgrade-the-first-control-plane-node actually specifically suggests upgrading kubeadm before using it to upgrade the cluster.

areed: We do upgrade kubeadm binary before running `kubeadm upgrade` but we don't upgrade kubectl until after.

MikaelSmith: Ok. I must've missed that part.

areed (commented):
It would be great to continue using eviction as the default case. What do you think about something like this to only disable_eviction on single-primary clusters that have a poddisruptionbudget?

```
function disable_eviction() {
  if kubectl client version < 1.18 => return false

  if kubectl get poddisruptionbudgets -A --no-headers | wc -l  == 0 => return false

  if is_ha => false

  return true
}

MikaelSmith:
I'm not really clear on how evict and delete differ in practice. My reading is that both will attempt graceful termination, but with `delete` it will force shutdown after the grace period while eviction won't. I can add your check, but not sure it's an important distinction here.

MikaelSmith:
Also https://kubernetes.io/docs/tasks/administer-cluster/safely-drain-node/. As far as I can tell, the logic you wrote is essentially consistent with the difference between drain and cordon+delete. And honoring `is_ha` wouldn't help my particular problem, since I have a single postgres instance that can't be safely evicted while maintaining service.

areed:
I see what you mean, but I don't want to break the case where someone had a pdb for 3 pods and is counting on that to keep the app running during upgrades. How about first try the drain with timeout, then if that fails do the drain with evicition disabled?

MikaelSmith:
Yeah, that seems like it could work.

areed:
> would that cause a problem for the upgrade if some pods are left running?

No, I think the drain is recommended to keep apps available, especially if something goes wrong. I haven't seen the control plane have any problems with some pods running during upgrades.

MikaelSmith:
Then maybe it's better to just leave it as-is, and we can document that upgrades might take a bit while this step finishes.

areed:
We could lower the timeout from 5 minutes to 2 or 3 to make it a bit less painful.

PR #945, opened 2020-12-01
Record each time an install script is downloaded

areed:


web/src/controllers/Scripts.ts
```diff
@@ -48,6 +53,22 @@ export class Installers {
     }
     installer = installer.resolve();
 
+    try {
+      let clientIP = request.get("True-Client-IP");
```
divolgin, line 57: How about using this here https://www.npmjs.com/package/request-ip

areed: good idea

divolgin (approved):

PR #951, opened 2020-12-04
Skip SElinux add-on on Ubuntu unless selinux is installed

areed:


laverya (approved):
Only extra thing I'd love is a message like "not installing selinux" or something

PR #954, opened 2020-12-07
Include registry data in DR snapshots

divolgin:


addons/kotsadm/alpha/tmpl-secret-s3.yaml
```diff
@@ -2,9 +2,6 @@ apiVersion: v1
 kind: Secret
 metadata:
   name: kotsadm-s3
-  labels:
```
salahalsaleh, line 5: we should remove these in the install script explicitly too

salahalsaleh: unless we always apply the secret

addons/registry/2.7.1/kustomization.yaml
```diff
@@ -1,4 +1,3 @@
 namespace: kurl
 
 resources:
```
salahalsaleh, line 3: is this key still needed?

scripts/install.sh
```diff
@@ -379,6 +379,7 @@ function main() {
     addon_for_each addon_load
     init
     apply_installer_crd
+    create_registry_service
```
areed, line 382: won't be defined when registry add-on isn't included in your spec

addons/registry/2.7.1/kustomization.yaml
divolgin: not sure... this is how I tested :)

areed (commented):
I don't understand why the KURL_REGISTRY_IP is an option for the user and why it's a bash flag instead of part of the yaml spec.

divolgin:
@areed KURL_REGISTRY_IP is not something that can ever be defined by a vendor, and it's really only applicable when restoring another kURL cluster.  So I made it a command line flag only, at least for now.

areed (approved):

PR #979, opened 2020-12-15
add centos 7.4 to testgrid

laverya:
it's the earliest centos version listed as supported

Tested [here](https://testgrid.kurl.sh/run/laverya%20centos%207.4) as the PR was made - 1708 was the first ISO to register as "centos 7.4" and not "7.3"


areed (approved):

laverya:
It appears that all the 7.4 installs are timing out
IMO this means we should look into fixing things, but since [we list 7.4 as supported](https://kurl.sh/docs/install-with-kurl/system-requirements) into the grid it goes

PR #980, opened 2020-12-16
collect support bundles on ctrl+c or addon failure

laverya:


markpundsack:
Hmm... this sends it on ctrl-c as well? That might be unexpected and take a while to generate and send. Might need to only do it on failure. But, I hope we're only *offering* to send and not sending without consent. So asking on ctrl-c is weird.

laverya:
The idea with asking for it on ctrl+c is that there's a few forever-spinners that someone might give up on, and we want to find out about that @markpundsack

areed (approved):

PR #982, opened 2020-12-17
report that installs failed due to ctrl+c even before completing k8s install

laverya:


areed (approved):
Would it be better to test for a cluster in case this is a re-run of the install script?

laverya:
I agree that it would be better to test if this was a rerun, but I don't know if that should go in a handler like this

There's more rerun tracking stuff in the pipeline anyways

PR #986, opened 2020-12-18
add extra timeout for centos 7.4 runs

laverya:


laverya:
easier to just add extra timeout for all runs instead

PR #1005, opened 2021-01-06
contour 1.10.1 and 1.11.0

laverya:


addons/contour/1.10.1/contour.yaml
```diff
@@ -0,0 +1,1794 @@
...
+        - --xds-address=0.0.0.0
+        - --xds-port=8001
+        - --envoy-service-http-port=80
+        - --envoy-service-https-port=443
+        - --contour-cafile=/certs/ca.crt
+        - --contour-cert-file=/certs/tls.crt
+        - --contour-key-file=/certs/tls.key
+        - --config-path=/config/contour.yaml
+        command: ["contour"]
+        image: docker.io/projectcontour/contour:v1.10.1
```
areed, line 1607: This breaks docker-ee 1.13. Could the docker.io prefix be removed for all images?

areed (approved):

addons/contour/1.10.1/contour.yaml
laverya: Looking
I don't believe we did that for previous versions of contour though, and we don't support docker-ee at all anyways with kurl

laverya: it's the same in previous contour versions, so I don't think we need to support this yet: https://github.com/replicatedhq/kURL/blob/bd58c691bc777704f010d96e34088719977e078b/addons/contour/1.7.0/contour.yaml#L1612

https://github.com/replicatedhq/kURL/blob/470d2142b5929fe3626c236daa40682d80b58159/addons/contour/1.0.1/deployment.yaml#L190

markpundsack:
@laverya @DanStough Were these changes tested in Testgrid?

laverya:
they were, see: https://testgrid.kurl.sh/run/laverya%202021.01.06%20contour%201.10.1%20+%201.11.0%20staging

PR #1014, opened 2021-01-11
autogeneration of contour addon versions

laverya:


laverya:
see https://github.com/replicatedhq/kURL/pull/1015 for example PR

addons/contour/template/base/tmpl-configmap.yaml
```diff
@@ -0,0 +1,88 @@
+
```
areed, line 1: Is this file used?

areed (approved):

addons/contour/template/base/tmpl-configmap.yaml
laverya: No, actually
I can fix that

PR #1018, opened 2021-01-12
docs: testgrid branch builds proposal

DanStough:
Proposal for using TestGrid on Kurl branches. To be honest, considering the limitations here and the near-term utility of it, I'm thinking the time is better spent coming up with a strategy for supporting 3rd party addons, but I guess that's up to the team. 

Thanks to @laverya for working through these ideas with me. The installer spec was all him 🥇 .


design/tg-branches.md
```diff
@@ -0,0 +1,111 @@
...
+1. kurl.sh will be modified to allow any version for an addon in the spec with the `upstream_override` enabled.
+1. `tgrun` will be updated
+   1. Since this proposal will only really test updated addons, a flag to run only a single "latest" kurl installer against all OSes.
+   1. Flags to specify addon version and download url
+1. Modify install.sh to use `upstream_override` bucket path from the kurl spec.
+   1. Possibly modify `yamltobash` utils (and others?)
+1. Add a github action to run on a PR creation/update.
+   1. Do a diff and see if any of the addon dirs have been modified.
+   1. Push packages to `<s3>/pr` for modified dirs (wait?)
+   1. `tgrun queue --ref=<short commit>-pr-<#> --latest-only --version=contour=1.12.0 --upstream_override=<s3 url in previous step>`
```
areed, line 51: How does tgrun know which addons to override if there's a single flag `--upstram_override`. Wouldn't you need an upstream override flag for each addon?

areed (approved):

emosbaugh (approved):
Looks good!

design/tg-branches.md
emosbaugh: Could you make tgrun accept a spec?

DanStough: @areed good look. I was thinking it would look like the `version` flag: `--upstream_override=contour=<s3 url>`
Maybe better names would be `version_override` and `s3_override`?

@emosbaugh I think thank could work. Like a `kubectl patch`, right? I think the only thing I don't like about that is then I think you need a spec file hanging the repo around for convenience, but maybe there is a better.

design/tg-branches.md
```diff
@@ -0,0 +1,111 @@
...
+1. The Kurl installer spec will be updated to allow installers to be generated with `upsteam_override`.
+1. kurl.sh will be modified to allow any version for an addon in the spec with the `upstream_override` enabled.
+1. `tgrun` will be updated
+   1. Since this proposal will only really test updated addons, a flag to run only a single "latest" kurl installer against all OSes.
+   1. Flags to specify addon version and download url
+1. Modify install.sh to use `upstream_override` bucket path from the kurl spec.
+   1. Possibly modify `yamltobash` utils (and others?)
+1. Add a github action to run on a PR creation/update.
+   1. Do a diff and see if any of the addon dirs have been modified.
+   1. Push packages to `<s3>/pr` for modified dirs (wait?)
```
markpundsack, line 50: Why do things need to be pushed to s3 as opposed to pulled from GitHub directly?

design/tg-branches.md
```diff
@@ -0,0 +1,111 @@
...
+   1. `tgrun queue --ref=<short commit>-pr-<#> --latest-only --version=contour=1.12.0 --upstream_override=<s3 url in previous step>`
+   1. (Optional) Poll results and/or add TestGrid url in PR.
+
+### Limitations
```
markpundsack, line 54: `latest` won't work for all addons, and only some are included by default.

design/tg-branches.md
```diff
@@ -0,0 +1,111 @@
...
+Initial testing will be challenging considering it requires changes to a lot of the components. 
+Proposing to break this up into these PRs to mitigate risk and manage code review:
+- Change kurl.sh web to adopt new kurl spec, run test suite.
+- Modify `tgrun` with proposed changes, test that kurl.sh install fail for the new spec format. 
+- Thorough manual testing for changes to install.sh that pull from the repository.
+- Attempt an addon upgrade and see that the GH action produces the desired affect.
+
+Regressions should be detected  quickly as this will run as part of the PR process on the repo. 
+
+## Alternatives Considered
```
markpundsack, line 96: Could it be possible to specify an installer ID to run for a given PR? Perhaps by using a convention of branch naming where you can deduce the installer ID? This could get around addons that aren't in `latest`, as well as helping with new addons.

design/tg-branches.md
DanStough: I think you could clone the git repo and run the addons from there, but IMHO I think it's a smaller code shift to work with the S3 handling already built into kurl install.sh and the scripting in the repo for pushing packages.

design/tg-branches.md
DanStough: This is a really good point I didn't think of. When I look at the current addons though, none of them are mutually exclusive (except for the CRI).  In that case I see the override flags just appending to the default latest installer, which will generate a non-latest id for the installer.

design/tg-branches.md
DanStough: If I'm understanding how everything works, I think you could easily make a `tgrun --installer-id=<id>` flag to test a specific installer against all OSes, which could be independently valuable in the future for testing an arbitrary installer instead of the just the ones coded into `tgrun`. 

Not sure this would help with PRs though. To have an installer ID, the web api has to receive a spec. So someone would need to submit the spec manually (defeats the purpose), or you generate the ID based on a spec sent automatically as part of the PR (the current proposal using `tgrun`).

New addons will always be a problem because install.sh needs to be updated to handle the new addon, but can't be rendered until a new version of the web api is pushed.

PR #1019, opened 2021-01-13
feat: add s3 bucket override to installer spec

DanStough:
PR 1 of 3-ish.

Added the the new field `s3Override` field to the kurl web api per design doc.

Pretty sure this will break install.sh if anyone tried to use the new field, but considering it's not really for general, use I'd rather have the review now and have someone push back on merging.

### DIFF
--- a/web/src/installers/index.ts
+++ b/web/src/installers/index.ts
@@ -17,6 +17,7 @@ interface ErrorResponse {
 
 export interface KubernetesConfig {
   version: string;
+  s3Override?: string; 
   serviceCidrRange?: string;
   serviceCIDR?: string;
   HACluster?: boolean;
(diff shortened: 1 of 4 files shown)


web/src/installers/index.ts
```diff
@@ -10,13 +10,15 @@ import { MysqlWrapper } from "../util/services/mysql";
 import { instrumented } from "monkit";
 import { logger } from "../logger";
 import { Forbidden } from "../server/errors";
+import { consoleReporter } from "replicated-lint/dist/cmdutil/reporters";
```
areed, line 13: Is this needed?

web/src/test/installers.ts
```diff
@@ -506,6 +538,14 @@ spec:
           expect(out).to.equal(undefined);
         });
       });
+
+      describe("unknown versions w/ overrides", () => {
+        it("=> void", () => {
+          const out = Installer.parse(overrideUnknownVersion).validate();
+          console.log(out)
```
areed, line 545: Was this for dev only?

areed (approved):

web/src/installers/index.ts
DanStough: Nope, removing.

web/src/test/installers.ts
DanStough: Yeah, also removing.

PR #1020, opened 2021-01-14
Containerd auto

areed:
[Example PR](https://github.com/replicatedhq/kURL/pull/1023)

Includes a new type for the Manifest, `dockerout`, which takes a dockerfile, output dir, and version. The bin/save-manifest-assets.sh script will build a contextless image from the Dockerfile and run the default command and include everything in the `/out` directory of the container in the addon package.


addons/containerd/template/base/Manifest
```diff
@@ -0,0 +1,8 @@
+image pause k8s.gcr.io/pause:3.1
+image pause k8s.gcr.io/pause:3.2
```
laverya, line 2: why two pause versions?

areed: This should be 3.1 for containerd 1.3 and 3.2 for 1.4. I'll update.

Containerd overrides the pod sandbox image so depending on the version of Kubernetes installed, the Kubernetes airgap package may not contain the pause version that containerd actually uses.

laverya (approved):

PR #1027, opened 2021-01-15
Generate containerd versions

areed:
Also add containerd versions to testgrid and enable manually triggering the workflow.

### DIFF
--- a/web/src/installers/index.ts
+++ b/web/src/installers/index.ts
@@ -17,7 +17,7 @@ interface ErrorResponse {
 
 export interface KubernetesConfig {
   version: string;
-  s3Override?: string; 
+  s3Override?: string;
   serviceCidrRange?: string;
   serviceCIDR?: string;
   HACluster?: boolean;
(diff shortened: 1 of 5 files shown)


web/src/installers/index.ts
```diff
@@ -594,10 +594,7 @@ export class Installer {
       "19.03.4",
       "18.09.8",
     ],
-    containerd: [
-      "1.3.7",
-      "1.2.13",
-    ],
+    containerd: [ "1.3.7", "1.2.13"], // cron-containerd-update
```
emosbaugh, line 597: why no 1.4.3 here?

emosbaugh (approved):

web/src/installers/index.ts
laverya: And 1.3.9

laverya (approved):

web/src/installers/index.ts
areed: I was leaving those for the automated PR to fill in to confirm it's working.

PR #1028, opened 2021-01-15
feat(installer): add s3 override to install tooling

DanStough:
Adding support to the actual install script for using the s3 override property. I test against a new version of the metrics server that I'll use to benchmark the PR process in the end. Also made sure it worked with an existing installer ID with a valid DIST_URL.


scripts/common/addon.sh
```diff
@@ -105,6 +110,17 @@ function addon_load() {
     load_images $DIR/addons/$name/$version/images
 }
 
+function addon_fetch() {
+    local url=$1
+
+    local archiveName=$(basename $url)
```
emosbaugh, line 116: Do we prefer snake or camel case? Seems to be a mix in this pr

emosbaugh (approved):

scripts/common/addon.sh
DanStough: Hmm. Looking at the other scripts, locals are camel case and functions are snake case. I'll change my PR to be consistent. IMHO I would make everything snake case - looks like that's the [Google Style guide ](https://google.github.io/styleguide/shellguide.html#s7-naming-conventions) as well.

PR #1030, opened 2021-01-15
Fix containerd version order

areed:


addons/containerd/template/script.sh
```diff
@@ -27,19 +27,19 @@ function find_common_versions() {
     docker build -t ubuntu18 -f Dockerfile.ubuntu18 .
     docker build -t ubuntu20 -f Dockerfile.ubuntu20 .
 
-    CENTOS7_VERSIONS=($(docker run --rm -i centos7 yum list --showduplicates containerd.io | grep -Eo '1\.[[:digit:]]+\.[[:digit:]]+' | grep -vE '1\.[012]\.' | sort | uniq))
+    CENTOS7_VERSIONS=($(docker run --rm -i centos7 yum list --showduplicates containerd.io | grep -Eo '1\.[[:digit:]]+\.[[:digit:]]+' | grep -vE '1\.[012]\.' | sort | uniq | tac))
```
laverya, line 30: why not just `sort -r`?

laverya (approved):

addons/containerd/template/script.sh
areed: Updated

PR #1037, opened 2021-01-18
Feat/tgrun default overrides

DanStough:
Pivoted a little from the original plan to use CLI flags for everything to just using a `--spec` flag at the CLI per @emosbaugh's suggestion. If you want to test the `tgrun` changes:

```bash
./bin/tgrun queue --staging --spec "$(cat ./hack/test-spec.yaml)" --ref user-test
# OR
./bin/tgrun queue --staging --latest-only --ref user-test
```

When this gets merged I have a PR for metrics-server that I will make to test out the action.

Lessons learned:
- [StringToString flag type doesn't play well with Viper AutoEnv](https://github.com/spf13/viper/issues/911) - not worth the effort now IMHO
- As a GH Actions newbie: the checkout action only checks out one commit -> makes it hard to do a diff in this case.

### DIFF
--- a/testgrid/tgrun/pkg/scheduler/scheduler.go
+++ b/testgrid/tgrun/pkg/scheduler/scheduler.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"encoding/json"
 	"fmt"
+	"gopkg.in/yaml.v2"
 	"io/ioutil"
 	"math/rand"
 	"net/http"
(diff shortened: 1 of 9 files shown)


testgrid/tgrun/pkg/scheduler/scheduler.go
```diff
@@ -107,11 +113,42 @@ func Run(schedulerOptions types.SchedulerOptions) error {
...
+	fmt.Printf("Started tests on %d specs across %d images\n", len(kurlPlans), len(operatingSystems))
 
 	return nil
 }
 
+func getKurlPlans(schedulerOptions types.SchedulerOptions) ([]types.Instance, error) {
+	var kurlPlans []types.Instance
+
+	// Custom Kurl Spec takes precedence
+	if schedulerOptions.Spec != "" {
```
emosbaugh, line 125: is tgapi open? last time i checked i think it was. i worry that this is a security issue as it may open us up to arbitrary specs being run. right now we dont have anything like addon `script` or `helm` but if we did wouldnt this be an issue? @laverya wanna weigh in here?

emosbaugh (approved):

PR #1039, opened 2021-01-19
Testing PR CI

DanStough:
Updating Metrics server to see if TestGrid runs.

### DIFF
--- a/addons/metrics-server/0.4.1/kubelet-insecure-tls.yaml
+++ b/addons/metrics-server/0.4.1/kubelet-insecure-tls.yaml
@@ -1,0 +1,16 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: metrics-server
+  namespace: kube-system
+spec:
+  template:
+    spec:
+      containers:
+      - name: metrics-server
+        args:
+          - --cert-dir=/tmp
+          - --secure-port=4443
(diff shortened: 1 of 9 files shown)


addons/metrics-server/0.4.1/kubelet-insecure-tls.yaml
```diff
@@ -0,0 +1,16 @@
...
+  template:
+    spec:
+      containers:
+      - name: metrics-server
+        args:
+          - --cert-dir=/tmp
+          - --secure-port=4443
+          - --kubelet-insecure-tls
+          - --kubelet-preferred-address-types=InternalIP,ExternalIP,Hostname
+          - --kubelet-use-node-status-port
```
DanStough, line 16: These lines are new.

laverya:
the testgrid spec appears to be the following:
```
{
    "kind": "Installer",
    "apiVersion": "cluster.kurl.sh/v1beta1",
    "metadata": {
        "name": "test",
        "creationTimestamp": null
    },
    "spec": {
        "kubernetes": {
            "version": "latest"
        }
    }
}
```
(and is the same for both runs)

So I think there's something going on here - maybe the testgrid CLI needs to be updated to use the new spec?

DanStough:
@laverya good find. The new spec should be gettting passed into the CLI but I'll have to debug.

PR #1041, opened 2021-01-20
feat(addon): add http config for contour v1.X

DanStough:
Adding options to define NodePorts for contour HTTP and HTTPS. Backported only the v1.X versions. Also added to the automation template.


laverya (approved):
I would like to see this added to one of the contour testgrid runs, but lgtm

DanStough:
Good call. I saw there were a couple contour 1.7 tests, so I modified one to 1.11 and added these overrides to it.

PR #1042, opened 2021-01-21
testgrid metrics

laverya:


testgrid/tgapi/pkg/handlers/healthz.go
```diff
@@ -13,5 +17,17 @@ func Healthz(w http.ResponseWriter, r *http.Request) {
 		IsAlive: true,
 	}
 
+	pending, running, timedOut, err := testinstance.GetTestStats()
+	if err != nil {
+		fmt.Printf("error getting test stats: %s", err)
+		return // being unable to get test stats indicates unhealthiness
```
emosbaugh, line 23: should you be returning some other status here?

emosbaugh (changes requested):

testgrid/tgapi/pkg/handlers/healthz.go
laverya: yeah, I'll have that do a 500

laverya: (it still works for "this pod is not healthy" with just the return, but it should be more explicit)

emosbaugh (approved):

PR #1052, opened 2021-01-25
Alpha support for RKE2 kubernetes distribution

emosbaugh:


Makefile
```diff
@@ -170,6 +170,7 @@ build/install.sh:
 		cat $$script >> tmp/install.sh ; \
 	done
 	sed -n '/# Magic end/,$$p' scripts/install.sh | sed '1d' >> tmp/install.sh
+	sed -i 's/^DISABLE_REPORTING=.*/DISABLE_REPORTING=1/' tmp/install.sh # disable reporting
```
areed, line 173: This is going to disable reporting for prod.

areed: You could put `DISABLE_REPORTING=1` in the Manifest and then delete the whole line in `make build/templates/install.tmpl`

areed (approved):

DanStough:
> Testgrid Run(s) Executing @

I need to fix this so you don't get an empty message. There are no templates so it didn't run anything.

PR #1053, opened 2021-01-26
RKE2 installer spec

emosbaugh:


DanStough (changes requested):
AFAIK, I think we need to generate the new installer CRD yaml and commit that back as well. I think `make generate` should add it: https://github.com/replicatedhq/kURL/blob/e6d3abb98aa7e218118972b50ea8e807150409bb/kurlkinds/README.md
  
There is crd here as well, but I couldn't really find where we use it. I guess it can be ignored. https://github.com/replicatedhq/kURL/blob/1f4329c31d36b5184343c40f323c0baad34511d8/web/crds/cluster_v1beta1_installer.yaml

emosbaugh:
@DanStough updated

DanStough (approved):

PR #1055, opened 2021-01-26
feat: initial addon testgrid templates

DanStough:
Adding basic templates for most addons.

### DIFF
--- a/addons/cert-manager/template/testgrid/k8s-docker.yaml
+++ b/addons/cert-manager/template/testgrid/k8s-docker.yaml
@@ -1,0 +1,9 @@
+kubernetes:
+  version: "latest"
+weave:
+  version: "latest"
+docker:
+  version: "latest"
+certManager:
+  version: "__testver__"
+  s3Override: "__testdist__"
--- a/addons/collectd/template/testgrid/k8s-docker.yaml
+++ b/addons/collectd/template/testgrid/k8s-docker.yaml
@@ -1,0 +1,9 @@
+kubernetes:
+  version: "latest"
+weave:
+  version: "latest"
+docker:
+  version: "latest"
+collectd:
+  version: "__testver__"
+  s3Override: "__testdist__"
--- a/addons/contour/template/testgrid/k8s-docker.yaml
+++ b/addons/contour/template/testgrid/k8s-docker.yaml
@@ -1,0 +1,9 @@
+kubernetes:
+  version: "latest"
+weave:
+  version: "latest"
+docker:
+  version: "latest"
+contour:
+  version: "__testver__"
+  s3Override: "__testdist__"
--- a/addons/openebs/template/testgrid/k8s-docker.yaml
+++ b/addons/openebs/template/testgrid/k8s-docker.yaml
@@ -1,0 +1,11 @@
+kubernetes:
+  version: "latest"
+weave:
+  version: "latest"
+docker:
+  version: "latest"
+minio:
+  version: "latest"
+openebs:
+  version: "__testver__"
+  s3Override: "__testdist__"
--- a/addons/rook/template/testgrid/k8s-docker.yaml
+++ b/addons/rook/template/testgrid/k8s-docker.yaml
@@ -1,0 +1,11 @@
+kubernetes:
+  version: "latest"
+weave:
+  version: "latest"
+docker:
+  version: "latest"
+minio:
+  version: "latest"
+rook:
+  version: "__testver__"
+  s3Override: "__testdist__"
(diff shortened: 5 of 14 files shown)


addons/cert-manager/template/testgrid/k8s-docker.yaml
```diff
@@ -0,0 +1,11 @@
+kubernetes:
+  version: "latest"
+weave:
+  version: "latest"
+rook:
+  version: "latest"
```
laverya, line 6: rook isn't required here

addons/collectd/template/testgrid/k8s-docker.yaml
```diff
@@ -0,0 +1,11 @@
+kubernetes:
+  version: "latest"
+weave:
+  version: "latest"
+rook:
+  version: "latest"
```
laverya, line 6: or here

addons/contour/template/testgrid/k8s-docker.yaml
```diff
@@ -0,0 +1,11 @@
+kubernetes:
+  version: "latest"
+weave:
+  version: "latest"
+rook:
+  version: "latest"
```
laverya, line 6: or here

addons/openebs/template/testgrid/k8s-docker.yaml
```diff
@@ -0,0 +1,9 @@
+kubernetes:
+  version: "latest"
+weave:
+  version: "latest"
+docker:
+  version: "latest"
+openebs:
+  version: "__testver__"
+  s3Override: "__testdist__"
```
laverya, line 9: IMO we should also include minio or registry here to use a PVC and ensure openebs is working

addons/rook/template/testgrid/k8s-docker.yaml
```diff
@@ -0,0 +1,9 @@
+kubernetes:
+  version: "latest"
+weave:
+  version: "latest"
+docker:
+  version: "latest"
+rook:
+  version: "__testver__"
+  s3Override: "__testdist__"
```
laverya, line 9: IMO we should also include minio, kotsadm or registry here to use a PVC and ensure rook is working

addons/velero/template/testgrid/k8s-docker.yaml
```diff
@@ -0,0 +1,11 @@
+kubernetes:
+  version: "latest"
+weave:
+  version: "latest"
+rook:
+  version: "latest"
```
laverya, line 6: I don't believe rook is required here, but I'm not positive

addons/weave/template/testgrid/k8s-docker.yaml
```diff
@@ -0,0 +1,9 @@
+kubernetes:
+  version: "latest"
+rook:
+  version: "latest"
```
laverya, line 4: rook isn't required here

addons/cert-manager/template/testgrid/k8s-docker.yaml
DanStough: I left Rook (or some other storage class) in all of them because when the TestGrid runner executes the part of the script that needs to create a troubleshooting bundle, it was dying without storage for some reason. Part of the reason for moving to these tempalte. Is there a better way around this?

laverya: this one: https://github.com/replicatedhq/kURL/blob/db70ebca6bf8eca87aae75a515189de69947d507/testgrid/tgrun/pkg/instances/k8s119_minimal_containerd137.go#L9

laverya (approved):
I think we should merge this and deal with pruning test cases later

addons/openebs/template/testgrid/k8s-docker.yaml
DanStough: adding minio

addons/rook/template/testgrid/k8s-docker.yaml
DanStough: adding minio

addons/cert-manager/template/testgrid/k8s-docker.yaml
DanStough: You called it: https://testgrid.kurl.sh/run/certManger-wo-rook

DanStough: removing rook from the mentioned addons.

addons/velero/template/testgrid/k8s-docker.yaml
DanStough: I would guess that you're right but leaving this one since it makes sense for these to be deployed together.

PR #1061, opened 2021-01-26
Minio addon hostpath option

salahalsaleh:


addons/minio/2020-01-25T02-50-51Z/install.sh
```diff
@@ -23,6 +33,13 @@ function minio() {
 }
 
 function minio_creds() {
+    if [ -n "$MINIO_HOSTPATH" ]; then
```
divolgin, line 36: Is this block safe to run on updates to existing working cluster?

divolgin (approved):

addons/minio/2020-01-25T02-50-51Z/install.sh
salahalsaleh: This will only execute if a hostPath option is provided, and will delete that `.minio.sys` directory from that hostPath location. So it should be safe. Any particular edge cases you have in mind?

divolgin: I see it also scales the deployment down.  What scales it back up?  And is there a case where we would want to keep the directory?

divolgin: maybe add a comment explaining this logic?

salahalsaleh: this will scale it back up https://github.com/replicatedhq/kURL/pull/1061/files#diff-93b1b6f9d511cb0f57d148dc9967819f594ea155dadadba8182cc43c2914b123R7

salahalsaleh: explanation added

addons/minio/2020-01-25T02-50-51Z/install.sh
```diff
@@ -23,6 +33,18 @@ function minio() {
...
 
 function minio_creds() {
+    # in the case of using a "hostPath", minio will generate a ".minio.sys" directory
+    # that is used to control access to that hostPath. this can be a problem when that hostpath
+    # is a directory in a shared file system (an NFS mount for example) and the installer is being
+    # run on a fresh instance, because new credentials will be generated and minio won't be able access the old data,
+    # so we make sure that ".minio.sys" is regenerated from the current minio credentials.
+    if [ -n "$MINIO_HOSTPATH" ]; then
+        # reset minio server access creds
+        rm -rf "$MINIO_HOSTPATH/.minio.sys"
```
markpundsack, line 43: @salahalsaleh Is this going to remove that directory every time kurl install is re-run? That seems a bit heavy-handed and dangerous.

Are we positive there are no negative consequences for wiping out this directory? What if someone had set up Minio manually on that same path, and used encryption? I know we don't support encryption in the kURL addon, but the NFS mount point could easily be outside kURL control.

Is it possible to detect if the .minio.sys directory is compatible with the current instance, and only delete if necessary? Perhaps with user confirmation as well?

salahalsaleh: technically, only removing the `config` directory inside `.minio.sys` is required, and both are regenerated if minio starts up and couldn't find them. i can update the code to only delete the `config` directory.

in case another minio instance was set up manually with server side encryption, deleting this directory will prevent that minio instance from accessing the data (we don't delete the actual data).

user confirmation logic sounds like a good idea.

addons/minio/2020-01-25T02-50-51Z/install.sh
```diff
@@ -25,16 +35,84 @@ function minio() {
...
+    if [ ! -d "$MINIO_CONFIG_PATH" ]; then
+        return 0
+    fi
+
+    printf "\n${YELLOW}The $MINIO_HOSTPATH directory will be re-configured to be used by this minio instance. Would you like to continue?${NC} "
+    if ! confirmN "-t 120"; then
+        bail "\nWill not re-configure $MINIO_HOSTPATH."
+    fi
+
+    rm -rf "$MINIO_CONFIG_PATH"
```
divolgin, line 107: Since this will render minio data useless, how about `rm -rf $KURL_DIR` and then make that warning more dire and color it red?

divolgin (approved):

divolgin (approved):

PR #1064, opened 2021-01-27
RKE2 fixes

emosbaugh:


scripts/common/rke2.sh
```diff
@@ -507,14 +507,14 @@ rke_install() {
...
 
-    # TODO(dan): moved from init
-    exportKubeconfig
-
     # For Kubectl and Rke2 binaries 
-    echo "export PATH=\$PATH::/var/lib/rancher/rke2/bin/" | sudo tee -a /etc/profile > /dev/null
-    export PATH=$PATH:/var/lib/rancher/rke2/bin/
+    echo "export PATH=\$PATH:/var/lib/rancher/rke2/bin" | tee -a /etc/profile > /dev/null
+    export PATH=$PATH:/var/lib/rancher/rke2/bin
     export KUBECONFIG=/etc/rancher/rke2/rke2.yaml   # TODO(dan): move this to exportKubeconfig?
```
DanStough, line 513: Do we need this line anymore since you fixed exportKubeconfig with the disto interface?

DanStough (approved):

scripts/common/rke2.sh
emosbaugh: i think we still do. there is still something similar in kubeadm install

https://github.com/replicatedhq/kURL/blob/master/scripts/install.sh#L375

PR #1065, opened 2021-01-27
RKE2 bundles and airgap support

emosbaugh:


scripts/common/plugins.sh
```diff
@@ -0,0 +1,46 @@
+export KUBECTL_PLUGINS_PATH=/usr/local/bin
```
emosbaugh, line 1: this was all moved from kubernetes.sh

areed (approved):

PR #1072, opened 2021-01-28
Do not download kubernetes packages twice

emosbaugh:
I tried to prevent changing any other code paths

### DIFF
--- a/scripts/common/kubernetes.sh
+++ b/scripts/common/kubernetes.sh
@@ -1,18 +1,17 @@
-export KUBECTL_PLUGINS_PATH=/usr/local/bin
 
 function kubernetes_host() {
     kubernetes_load_ipvs_modules
 
     kubernetes_sysctl_config
 
+    kubernetes_install_host_packages "$KUBERNETES_VERSION"
+
     # For online always download the kubernetes.tar.gz bundle.
     # Regardless if host packages are already installed, we always inspect for newer versions
     # and/or re-install any missing or corrupted packages.
-    if [ "$AIRGAP" != "1" ] && [ -n "$DIST_URL" ]; then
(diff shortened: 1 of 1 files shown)


scripts/common/kubernetes.sh
```diff
@@ -1,19 +1,18 @@
-export KUBECTL_PLUGINS_PATH=/usr/local/bin
 
 function kubernetes_host() {
     kubernetes_load_ipvs_modules
 
     kubernetes_sysctl_config
 
+    kubernetes_install_host_packages "$KUBERNETES_VERSION"
```
areed, line 7: Why was this moved to before `kubernetes_get_host_packages_online` ?

emosbaugh: kubernetes_install_host_packages calls kubernetes_get_host_packages_online at the top of the function. kubernetes_install_host_packages is called in other places so im trying to avoid making changes to it.

areed: In that case the block beginning on 12 should be removed. That will never be true.

emosbaugh: it will be if `kubernetes_host_commands_ok`

```
# k8sVersion is an argument because this may be used to install step versions of K8s during an upgrade
# to the target version
function kubernetes_install_host_packages() {
    k8sVersion=$1
    logStep "Install kubelet, kubeadm, kubectl and cni host packages"
        return
    fi
    if [ "$AIRGAP" != "1" ] && [ -n "$DIST_URL" ]; then
        kubernetes_get_host_packages_online "$k8sVersion"
    fi
```

areed: And that's needed to get the images that are loaded right after

PR #1078, opened 2021-01-29
Kubernetes 1.20.2

areed:


scripts/upgrade.sh
```diff
@@ -52,39 +52,35 @@ maybe_upgrade() {
 
         upgrade_kubeadm "$KUBERNETES_VERSION"
 
-        case "$KUBERNETES_TARGET_VERSION_MINOR" in
```
areed, line 55: This was legacy code from platform when kubeadm api changed in 1.15.

laverya (approved):

PR #1091, opened 2021-01-30
Include docker and kubernetes in generated versions

areed:
Also prefer order defined for add-on versions on the Installer class since
that determines latest. We do not want weave 2.7.0 above 2.6.5, for
example, since 2.6.5 is latest and 2.7.0 has bugs.

### DIFF
--- a/generate-addons.js
+++ b/generate-addons.js
@@ -28,6 +28,7 @@ const uploadFile = (file) => {
   });
 }
 
+const preferredVersions = JSON.parse(fs.readFileSync("./web/versions.json"));
 let addons = [];
 let supportedVersions = new Object();
 const specDir = "./addons";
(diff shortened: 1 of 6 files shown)


web/package.json
```diff
@@ -45,7 +45,7 @@
     "nyc": "^11.4.1",
     "snyk": "^1.41.1",
     "start-server-and-test": "^1.9.0",
-    "ts-node": "^4.0.2",
+		"ts-node": "^4.0.2",
```
salahalsaleh, line 48: indentation

areed: fixed

generate-addons.js
```diff
@@ -68,11 +69,20 @@ fs.readdir(specDir, (err, files) => {
...
+      if (preferredVersions[file]) {
+        sv = preferredVersions[file];
+      }
       sv.unshift("latest");
       supportedVersions[file] = sv
     }
   });
 
+  supportedVersions.kubernetes = preferredVersions.kubernetes;
+  preferredVersions.kubernetes.unshift("latest");
```
salahalsaleh, line 81: i don't get why doing `unshift` at this point, `preferredVersions.kubernetes` is not used after this line. unless you meant to use `supportedVersions`?

salahalsaleh: i guess modifying `preferredVersions` would also modify `supportedVersions`, but it's less clear this way

areed: I'll change to `supportedVersions`

salahalsaleh (approved):

salahalsaleh (approved):

PR #1092, opened 2021-02-01
Install node dependencies for generate-versions.ts

areed:


.github/workflows/deploy-staging.yaml
```diff
@@ -53,6 +53,10 @@ jobs:
     runs-on: ubuntu-18.04
     steps:
     - uses: actions/checkout@v2
+    - uses: actions/setup-node@v1
```
DanStough, line 56: any reason not to use @v2 here?

DanStough (approved):

.github/workflows/deploy-staging.yaml
areed: no, I'll update

PR #1096, opened 2021-02-01
Testgrid api queue endpoint authentication

emosbaugh:


.github/workflows/daily-testgrid.yaml
```diff
@@ -25,13 +25,14 @@ jobs:
 
     - name: testgrid-queue-prod
       env:
-        TESTGRID_API: "https://api.testgrid.kurl.sh/"
```
emosbaugh, line 28: this should be `TESTGRID_TESTGRID_API`

laverya (approved):

PR #1101, opened 2021-02-02
docs: helm addon proposal

DanStough:


design/helm.md
```diff
@@ -0,0 +1,145 @@
...
+    version: "latest"
+  helmReleases:
+    # Basic Idea, Option 1
+    - release: nginx-ingress
+      namespace: kube-system    # Optional
+      chart: 
+        name: bitnami/nginx
+        repo: https://charts.bitnami.com/bitnami 
+        version: 8.5.1          # Optional - uses latest if not provided
+      values: |                 # Optional
```
emosbaugh, line 70: it would be awesome if you somehow didnt have to include this in your spec. i cant really think of how. possibly a url but that would not work for airgap

design/helm.md
```diff
@@ -0,0 +1,145 @@
...
+        repo: https://charts.bitnami.com/bitnami 
+      values:
+        postgresqlUsername: dan
+        postgresqlPassword: opensesame
+    # Option 3 (Config Url)
+    - release: postgres-auth
+      chart: 
+        name: bitnami/postgresql
+        repo: https://charts.bitnami.com/bitnami 
+      values: https://gist.githubusercontent.com/DanStough/b4376b7a6aa5b73c6a647bff504212df/raw/829680f4a474720dc36b7a963d2d691305b8cdfb/postgres-auth.yaml
```
emosbaugh, line 89: i see :). How will this work with airgap?

design/helm.md
```diff
@@ -0,0 +1,145 @@
...
+For airgapped the chart and artifacts need to be packaged into a bundle. The builder should perform the following:
+1. `helm pull` the repo
+1. `helm template` to render the chart with the provided values
+1. Generate an adhoc `manifest` file by parsing `image` specs from the rendered yaml.
+1. Generate a bundle including the chart + containers
+
+In the airgapped case, install.sh will perform the following:
+1. Unpack the helm bundle on initialization
+1. Load docker containers as part of the initialization
+1. Unarchive the chart and install with the installer spec values during the install phase
```
emosbaugh, line 40: Download the values yaml from url if necessary...

design/helm.md
```diff
@@ -0,0 +1,145 @@
...
+
+For airgapped the chart and artifacts need to be packaged into a bundle. The builder should perform the following:
+1. `helm pull` the repo
+1. `helm template` to render the chart with the provided values
+1. Generate an adhoc `manifest` file by parsing `image` specs from the rendered yaml.
+1. Generate a bundle including the chart + containers
+
+In the airgapped case, install.sh will perform the following:
+1. Unpack the helm bundle on initialization
+1. Load docker containers as part of the initialization
```
emosbaugh, line 39: images not containers right?

emosbaugh: what if it includes an operator? is there a common way to specify images in a helm chart? if not could you add an `additionalImages` property for airgap purposes?

design/helm.md
emosbaugh, line 89: id opt for allow url or inline. in that case maybe two separate properties? `values` and `valuesUrl`?

design/helm.md
```diff
@@ -0,0 +1,145 @@
...
+Kurl will bundle the Helm binary into the common tools bundle already delivering krew and kubectl plugins. The install.sh script will perform the following in the online case:
+1. Unarchive Helm and move into the `$PATH` and `/usr/bin`
+1. For each release, check if the release exists. 
+1. If the release exists, perform a `helm upgrade`.
+1. If the release doesn't exist, perform a `helm install`
+
+For airgapped the chart and artifacts need to be packaged into a bundle. The builder should perform the following:
+1. `helm pull` the repo
+1. `helm template` to render the chart with the provided values
+1. Generate an adhoc `manifest` file by parsing `image` specs from the rendered yaml.
```
emosbaugh, line 34: what if it includes an operator? is there a common way to specify images in a helm chart? if not could you add an additionalImages property for airgap purposes?

design/helm.md
```diff
@@ -0,0 +1,145 @@
...
+
+Kurl will bundle the Helm binary into the common tools bundle already delivering krew and kubectl plugins. The install.sh script will perform the following in the online case:
+1. Unarchive Helm and move into the `$PATH` and `/usr/bin`
+1. For each release, check if the release exists. 
+1. If the release exists, perform a `helm upgrade`.
+1. If the release doesn't exist, perform a `helm install`
+
+For airgapped the chart and artifacts need to be packaged into a bundle. The builder should perform the following:
+1. `helm pull` the repo
+1. `helm template` to render the chart with the provided values
```
emosbaugh, line 33: could i include an `installer-spec-file` as the end customer to override values?

design/helm.md
```diff
@@ -0,0 +1,145 @@
...
+      chart: 
+        name: bitnami/postgresql
+        repo: https://charts.bitnami.com/bitnami 
+      values: https://gist.githubusercontent.com/DanStough/b4376b7a6aa5b73c6a647bff504212df/raw/829680f4a474720dc36b7a963d2d691305b8cdfb/postgres-auth.yaml
+  ...
+```
+
+### Assumptions
+1. Authenticated Registries are NOT supported by Kurl
+1. The join and upgrade functionality of Kurl is transparent to Helm installs.
```
emosbaugh, line 95: join calls addon_join which calls load_images

design/helm.md
DanStough: I think this one was resolved below. I think we still need to support this option so we don't need to answer "what url" and "where do I put my file".

design/helm.md
```diff
@@ -0,0 +1,145 @@
+# Helm Addons
+
+To be an addon in a Kurl cluster, they must be contributed into the repo and maintained across new breaking changes and new versions.
+This proposal will give users the flexibility to specify their own addons through a helm chart support in the Kurl installer spec.
+
+
+## Goals
+- Allow any Helm chart to be specified as part of a Kurl spec.
```
diamonwiggins, line 8: Will we support helm2 or just helm3? Would imagine the latter?

design/helm.md
```diff
@@ -0,0 +1,145 @@
...
+spec: 
+  kubernetes: 
+    version: "latest"
+  weave: 
+    version: "latest"
+  containerd: 
+    version: "latest"
+  helm:                 # Optional Future Capability
+    version: "latest"
+  helmReleases:
```
diamonwiggins, line 62: With whatever spec we use, should the releases be nested under `helm` so someone doesn't slip up and try to install a HelmRelease without installing helm itself?

design/helm.md
DanStough: Curl the file and save into a directory with some kind of sematic name like <release>-value.yaml

design/helm.md
DanStough: Yep! will fix

design/helm.md
DanStough: Agree.

design/helm.md
DanStough: That's a really god question. AFAIK there isn't a way to specify this kind of relationship in helm since the operator doesn't need to declare it in a manifest or as a dependency.  `additionalImages` would be a good way to deal with this. I'll add an example.

design/helm.md
```diff
@@ -0,0 +1,145 @@
...
+To be an addon in a Kurl cluster, they must be contributed into the repo and maintained across new breaking changes and new versions.
+This proposal will give users the flexibility to specify their own addons through a helm chart support in the Kurl installer spec.
+
+
+## Goals
+- Allow any Helm chart to be specified as part of a Kurl spec.
+
+## Non Goals
+
+- Verify authenticity, trustworthyness and "goodness" of helm addons.
```
markpundsack, line 12: We may not need to verify things directly, but we may need to allow vendor and/or end user to verify things haven't been tampered with. e.g. when a vendor publishes a spec with a certain helm chart, capture fingerprints so the enterprise can later make sure it's the *same* content.

design/helm.md
```diff
@@ -0,0 +1,145 @@
...
+Helm is a popular tool for managing deployment configuration that allows application/addon developers or the community to maintain a consistent interface for deployment through a chart.
+By allowing Kurl users to specifify a chart and its values, we can allow customization of kutl with new addons while delegating addons support and maintance to its community.
+
+## High-Level Design
+
+Kurl will bundle the Helm binary into the common tools bundle already delivering krew and kubectl plugins. The install.sh script will perform the following in the online case:
+1. Unarchive Helm and move into the `$PATH` and `/usr/bin`
+1. For each release, check if the release exists. 
+1. If the release exists, perform a `helm upgrade`.
+1. If the release doesn't exist, perform a `helm install`
```
markpundsack, line 29: You can just use `helm upgrade --install` for both cases.

design/helm.md
DanStough: Noted!

design/helm.md
DanStough: Yeah, I think at this point supporting Helm 2 would just be sending the wrong message to new customers, plus they stopped providing any security updates.

design/helm.md
DanStough: Great! I had no idea about this flag.

design/helm.md
```diff
@@ -0,0 +1,145 @@
...
+  helm:                 # Optional Future Capability
+    version: "latest"
+  helmReleases:
+    # Basic Idea, Option 1
+    - release: nginx-ingress
+      namespace: kube-system    # Optional
+      chart: 
+        name: bitnami/nginx
+        repo: https://charts.bitnami.com/bitnami 
+        version: 8.5.1          # Optional - uses latest if not provided
```
markpundsack, line 69: Using latest could be problematic, particularly with airgap since airgap bundle would be created at a certain point in time. Could/would we detect upstream changes and create new airgap bundles proactively?

design/helm.md
```diff
@@ -0,0 +1,145 @@
...
+metadata: 
+  name: "latest"
+spec: 
+  kubernetes: 
+    version: "latest"
+  weave: 
+    version: "latest"
+  containerd: 
+    version: "latest"
+  helm:                 # Optional Future Capability
```
markpundsack, line 60: Random thought, but we could have a `helmfile` plugin (https://github.com/roboll/helmfile) and just provide configuration to it as a single "addon". Not saying that's better in any sense, but it's pushing the abstraction down one layer and leveraging even more open source.

design/helm.md
DanStough: I'm wondering if we even give folks the option here now that I think about it. We just use the latest Helm 3 until we've identified that maybe there are some versions that have problems, in which case we will release versions that have been tested.

design/helm.md
markpundsack: Why wouldn't you want to include this in your spec? Isn't that kind of the whole point? Are you saying that we would want to have some opinionated defaults? The point is to leverage the opinions of the underlying addon provider as much as possible, thus leveraging their helm chart defaults.

markpundsack: Or are you just saying it should be specified in some other file? I like the simplicity and clarity of having a single file.

design/helm.md
DanStough: I added this to the bundler section.

design/helm.md
emosbaugh: my thought was allowing for opinionated defaults

design/helm.md
DanStough: I agree it's problematic but we could give consumers the option the same as other kurl addons to be consistent. We'll have to hash airgap bundles anyway since any values change could generate a new bundle. The hash + the helm manifest will help us identify the installed version.

As for proactively building bundles, I imagine it's on demand now and it would be a big change to monitor upstream containers/images, but it could be done. I don't really know how the bundling system works yet TBH.

design/helm.md
DanStough: That seems like a viable option. I'm mostly neutral in opinion. I guess consumers will have to learn our syntax or the helmfile syntax which is the same burden. I kind of like having a simpler facade to helm that we manage and having one less moving part to deliver.  If you have strong feelings I think it's not that much of a change.

I'll at least identify it in the alternatives section for now.

design/helm.md
DanStough: Sure, we will definitely need to hash the chart, containers and values to make the airgap bundles. Maybe one new point here is that a hash would always have to include the values (as opposed to a URL as in option 3 of the spec).

design/helm.md
```diff
@@ -0,0 +1,145 @@
...
+
+### Design Decisions
+1. Re-install a chart: chart installs are not idempotent, so this will be treated as an upgrade.
+1. State between executions - the only place we want to keep a 
+1. Waiting - the (initial) default behavior of kurl will be to wait for resources to create. This will ensure that the cluster is in a known state before the user continues with the next set of installation (kotadm)
+1. RKE2 support: has built in helm chart support by putting the Rancher Helm CR in the directory `/var/lib/rancher/rke2/server/manifests/`, but we will initially manage helm support through the CLI to have one code path for all distros.
+ 
+### Limitations
+1. If someone changes the name of a helm 
+1. If a helm release cannot be upgraded, it will need to be fixed manually (no option to use --force)
```
markpundsack, line 105: Helmfile has an interesting comment about being able to periodically sync and enforce Helm charts match declarative configuration. I wonder if we should consider such a thing.

design/helm.md
DanStough: I was thinking of just having an external file here since the values can get crazy or we maybe there is a case where the installer is the same but the values are different per install. K3OS does the same thing with `k3os.install.config_url`

design/helm.md
DanStough: I think you'd need to move the periodic sync to a chron job in k8s (I didn't see anything about setting a timer or interval for helmfile) which could be done. I have this nagging feeling we are creating two sources of truth though, the installer and declarative helmfile (that I'm assuming we export) and that the two could get out of sync if people are trying to manage them locally.

design/helm.md
markpundsack: So interesting, but perhaps decoupleable case. As in, replacing our opinionated addons with versions installed via Helm makes sense. But I see that as different from the current experimentation which is about giving vendors control. It does make sense to solve both with the same tooling though, so I see where the question comes from.

design/helm.md
markpundsack: Yeah, I forgot we already have that problem with airgap bundles and `latest`.

design/helm.md
markpundsack: The intent here would definitely be to **not** have folks manipulate the helmfile after installation.

design/helm.md
DanStough: Per our conversation, I'm not sure if we do (not) need this ability with helm and if it's desirable to have users change this from the install command vs. in the installer. I think your point, which I agree with, is that users were doing this to get around inflexible addons, and now they can modify values directly in the installer to make similar changes.

DanStough:
I added updates to this doc based on using Helmfile if anyone would like to re-reviewing. I think the only controversial thing might be shipping rendered yaml in the airgap case instead of archived charts.

design/helm.md
markpundsack: My preference is to eventually let end-users provide their own kurl installer spec, which would then let them specify any Helm chart modifications too. I'm not sure we need to go half-way here and let end-users augment vendor-provided specs.

emosbaugh: @DanStough i want to clarify what "users" in your message means. i am assume you are referring to the vendor

DanStough:
I think this served it's purpose, this feature has been in beta for some time, and we need to decide if we want to remove or promote.

PR #1102, opened 2021-02-02
Install kurl dependencies into /var/lib/kurl

emosbaugh:


test/dev.md
```diff
@@ -58,7 +58,7 @@ We welcome contributions to kURL. We appreciate your time and help.
 cd kurl
 # validate your expected changes in install.sh|upgrade.sh and|or addons|packages
 # run installation
-sudo ./install.sh
+sudo ./install.sh kurl-install-directory=.
```
emosbaugh, line 61: this is a change to dev env. is there a better way to do this?

areed: You could change [this](https://github.com/replicatedhq/kURL/blob/master/bin/watchrsync.js#L28) to `${remote}:/var/lib/kurl`

areed (approved):

test/dev.md
emosbaugh: I explored that route. i think this is preferable because if i change that i have to add a bunch of sudos to those tar commands to download packages

areed (approved):

PR #1103, opened 2021-02-02
feat(kurlkinds): add helm spec

DanStough:


emosbaugh (approved):
i feel like environment could be a nice property to pass to the helmfile. im not sure there may be more but this is a good start.

PR #1104, opened 2021-02-02
Do not suppress curl output when downloading large packages

emosbaugh:


scripts/common/addon.sh
```diff
@@ -95,8 +95,8 @@ function addon_fetch() {
 
     local archiveName=$(basename $url)
 
-    echo "Fetching $archiveName"   
-    curl -sSLO "$url"
+    echo "Fetching $archiveName"
+    curl -SLO "$url"
```
areed, line 99: `S` is only relevant with `s` according to the man page

scripts/common/utilbinaries.sh
```diff
@@ -1,7 +1,7 @@
 
 function download_util_binaries() {
     if [ -z "$AIRGAP" ] && [ -n "$DIST_URL" ]; then
-        curl -Ss -L $DIST_URL/$KURL_BIN_UTILS_FILE | tar zx
+        curl -sSL $DIST_URL/$KURL_BIN_UTILS_FILE | tar zx
```
areed, line 4: Why is this one different?

areed (approved):

scripts/common/utilbinaries.sh
emosbaugh: oops. thanks

areed (approved):

PR #1112, opened 2021-02-03
Install kurl dependencies into /var/lib/kurl

emosbaugh:


scripts/common/utilbinaries.sh
```diff
@@ -95,6 +97,11 @@ function get_patch_yaml() {
                 ;;
             kubernetes-version)
                 ;;
+            kurl-install-directory)
```
emosbaugh, line 100: if this is set should i print the prompt with the same in the join script? the load images task?

areed: Yes, that makes sense

areed (approved):

scripts/tasks.sh
```diff
@@ -272,15 +304,6 @@ function join_token() {
     # get the kurl url and installer id from the kurl-config configmap
     local kurl_url=$(kubectl -n kube-system get cm kurl-config -ojsonpath='{ .data.kurl_url }')
     local installer_id=$(kubectl -n kube-system get cm kurl-config -ojsonpath='{ .data.installer_id }')
-    local service_cidr=$(kubectl -n kube-system get cm kurl-config -ojsonpath='{ .data.service_cidr }')
-    local pod_cidr=$(kubectl -n kube-system get cm kurl-config -ojsonpath='{ .data.pod_cidry }')
```
emosbaugh, line 276: this has an extra `y` `.data.pod_cidry`

areed: That would create problems on proxy installs with non-default pod-cidrs

areed (approved):

PR #1113, opened 2021-02-04
Feat/helmfile online

DanStough:


scripts/common/helm.sh
```diff
@@ -0,0 +1,41 @@
...
+    helmfile --file helmfile-tmp.yaml sync  # || report_helm_failure  #TODO (dan): add reporting
+
+    rm helmfile-tmp.yaml
+
+    # TODO (dan): add reporting for helm
+    # report_helm_success
+}
+
+function helm_load() {
+    if [ "$AIRGAP" ]; then
```
emosbaugh, line 36: i think this should be `if [ "$AIRGAP" = "1" ]; then`

emosbaugh (changes requested):
looks good. one small thing

scripts/common/helm.sh
DanStough: Yes will fix.

diamonwiggins (approved):

emosbaugh (approved):

PR #1118, opened 2021-02-04
add a containerd k8s upgrade test

laverya:


testgrid/tgrun/pkg/instances/k8s1184_1202_containerd.go
```diff
@@ -0,0 +1,41 @@
...
+			UpgradeSpec: &types.InstallerSpec{
+				Kubernetes: kurlv1beta1.Kubernetes{
+					Version: "1.20.2",
+				},
+				Weave: &kurlv1beta1.Weave{
+					Version: "2.6.5",
+				},
+				Contour: &kurlv1beta1.Contour{
+					Version: "latest",
+				},
```
DanStough, line 34: Pin these versions to the pre-upgrade value since the other test you edited is already running the same upgrade config?

testgrid/tgrun/pkg/instances/k8s1184_1202_containerd.go
```diff
@@ -0,0 +1,41 @@
...
+				Containerd: &kurlv1beta1.Containerd{
+					Version: "1.4.3",
```
DanStough, line 36: "Latest" containerd would be another, orthogonal test, right?

DanStough (approved):

testgrid/tgrun/pkg/instances/k8s1184_1202_containerd.go
laverya: The other test is running the same config... except for containerd, which is the variable I want to isolate here.

testgrid/tgrun/pkg/instances/k8s1184_1202_containerd.go
laverya: Yeah
1.4.3 is currently the latest though

testgrid/tgrun/pkg/instances/k8s1184_1202_containerd.go
DanStough: Yeah I guess I don't see why either test is concerned with contour and weave upgrades which would create a false positive for kubernetes issues.

laverya: weave is required for everything - it's the only CNI currently supported on kurl.sh. Contour is there as a "we should install at least _one_ non-critical addon to make sure things are stable" thing.

DanStough: Understood. My comment was more about upgrading those versions, but maybe you're saying that this is a typical customer configuration we should test, or that those versions are the only one's compatible with these versions of k8s. 

It's a minor point - I'm fine with including it. I just have a hard time saying if a test fails, this is the one thing that broke it without looking at the logs. It seems like the existing daily's some times test more than one thing at a time.

PR #1120, opened 2021-02-04
Join bail if kubernetes is already installed

emosbaugh:


scripts/common/kubernetes.sh
```diff
@@ -549,3 +549,31 @@ function kubernetes_pod_completed() {
...
+}
+
+function kubernetes_is_join_node() {
+    if cat /opt/replicated/kubeadm.conf 2>/dev/null | grep -q 'kind: JoinConfiguration'; then
+        return 0
+    fi
+    return 1
+}
+
+function kubernetes_is_installed() {
```
emosbaugh, line 568: is there a better way to detect this?

scripts/common/kubernetes.sh
```diff
@@ -549,3 +549,31 @@ function kubernetes_pod_completed() {
 
     return 1
 }
+
+function kubernetes_is_current_cluster() {
+    local api_service_address="$1"
+    if cat /opt/replicated/kubeadm.conf 2>/dev/null | grep -q "${api_service_address}"; then
```
areed, line 555: It's possible they've changed their load balancer address and run the `set-kubeconfig-server` task. How about checking `/etc/kubernetes/kubelet.conf` instead?

scripts/common/kubernetes.sh
areed: I can't think of one

scripts/common/kubernetes.sh
emosbaugh: what about on secondary nodes? ```$ ls /etc/kubernetes/
manifests  pki```

emosbaugh: ill check both

areed (approved):

PR #1182, opened 2021-02-16
allow airgap testgrid access to all local ips and tgapi

laverya:


testgrid/tgrun/pkg/runner/runner.go
```diff
@@ -265,15 +265,30 @@ export KUBECONFIG=/etc/kubernetes/admin.conf
...
 iptables -A OUTPUT -p tcp -d 127.0.0.1 -j ACCEPT # accept comms to localhost
-iptables -A OUTPUT -p tcp -s 10.0.0.0/8 -j ACCEPT # accept comms to internal kubernetes IPs
+iptables -A OUTPUT -p tcp -s 10.0.0.0/8 -j ACCEPT # accept comms to internal IPs
+iptables -A OUTPUT -p tcp -s 172.16.0.0/12 -j ACCEPT # accept comms to internal IPs
+iptables -A OUTPUT -p tcp -s 192.168.0.0/16 -j ACCEPT # accept comms to internal IPs
 iptables -A OUTPUT -p tcp -j REJECT # reject comms to other IPs
 
+# test the lack of internet access
+echo "testing disabled internet"
+curl -v --connect-timeout 5 --max-time 5 "https://www.google.com"
```
divolgin, line 290: I've seen google.com resolve from GCE instances even when nothing else was accessible.

divolgin (approved):

testgrid/tgrun/pkg/runner/runner.go
laverya: It's not a GCE instance, thankfully
(and everything except our specific set of IPs is added to the droplist above)

PR #1183, opened 2021-02-16
explicitly fail test run if it has internet or if tarball fails to download

laverya:


testgrid/tgrun/pkg/runner/runner.go
```diff
@@ -287,10 +287,23 @@ iptables -A OUTPUT -p tcp -j REJECT # reject comms to other IPs
 
 # test the lack of internet access
 echo "testing disabled internet"
-curl -v --connect-timeout 5 --max-time 5 "https://www.google.com"
+curl -v --connect-timeout 5 --max-time 5 "http://www.google.com"
+CURL_EXIT_STATUS=$?
+if [ $TAR_EXIT_STATUS -eq 0 ]; then
```
DanStough, line 292: should be curl_exit_status

testgrid/tgrun/pkg/runner/runner.go
```diff
@@ -360,10 +373,23 @@ iptables -A OUTPUT -p tcp -j REJECT # reject comms to other IPs
...
```
DanStough, line 378: should be curl_exit status here too

DanStough (changes requested):

DanStough (approved):

PR #1187, opened 2021-02-16
improve testgrid after-failure reporting

laverya:
also allow access to the local IP address

### DIFF
--- a/testgrid/tgrun/pkg/runner/runner.go
+++ b/testgrid/tgrun/pkg/runner/runner.go
@@ -451,6 +451,14 @@ if [ $KURL_EXIT_STATUS -eq 0 ]; then
     echo "completed kurl run"
 else
     echo "failed kurl run with exit status $KURL_EXIT_STATUS"
+
+    echo "kubelet status"
+    systemctl status kubelet
+    echo "kubelet journalctl"
+    journalctl -xeu kubelet
+    echo "docker containers"
+    docker ps -a
+
     curl -s -X POST -d "{\"success\": false}" $TESTGRID_APIENDPOINT/v1/instance/$TEST_ID/finish
 fi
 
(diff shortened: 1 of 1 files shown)


testgrid/tgrun/pkg/runner/runner.go
```diff
@@ -275,6 +275,22 @@ for i in $APIENDPOINT_IPS; do
 	iptables -A OUTPUT -p tcp -d $i -j ACCEPT # accept comms to testgrid API IPs
 done
 
+# allow access to the local IP(s)
+_count=0
+_regex="^[[:digit:]]+: ([^[:space:]]+)[[:space:]]+[[:alnum:]]+ ([[:digit:].]+)"
+while read -r _line; do
+	[[ $_line =~ $_regex ]]
+	if [ "${BASH_REMATCH[1]}" != "lo" ] && [ "${BASH_REMATCH[1]}" != "kube-ipvs0" ] && [ "${BASH_REMATCH[1]}" != "docker0" ] && [ "${BASH_REMATCH[1]}" != "weave" ]; then
```
DanStough, line 283: ```[ "${BASH_REMATCH[1]}" != "docker0" ] && [ "${BASH_REMATCH[1]}" != "weave" ]```
It'll work for containerd and rancher stuff too right 😬 😅. I don't think this is your goal at the moment though.

DanStough (approved):

testgrid/tgrun/pkg/runner/runner.go
laverya: 1. I copied this from https://github.com/replicatedhq/kURL/blob/e73e48e899c4056a3c4c589acf144bef07b26af6/scripts/common/prompts.sh#L262
2. this runs before the install script, and thus should work just fine for those too (and adding more allowed addresses won't hurt, to be honest)

PR #1188, opened 2021-02-16
install bind-utils for airgapped testgrid runs that lack dig

laverya:


testgrid/tgrun/pkg/runner/runner.go
```diff
@@ -374,6 +375,7 @@ echo "upgrading installation"
 curl -L -o upgrade.tar.gz KURL_UPGRADE_URL
 
 # get the list of testgrid API IPs
+command -v dig >/dev/null 2>&1 || { yum -y install bind-utils; }
```
DanStough, line 378: Is there ever a case when you would need to install it for the upgrade and it wasn't installed initially?

DanStough (approved):

testgrid/tgrun/pkg/runner/runner.go
laverya: if the upgrade is airgapped and the initial install isn't

laverya: (but generally I've been trying to keep the code the same between the two sections whenever possible)

PR #1192, opened 2021-02-17
fix: don't run ctrd setup if you have docker installed

DanStough:


emosbaugh (approved):

areed (commented):
Proxy installs would have the same bug

PR #1201, opened 2021-02-17
fix: don't run containerd install if docker exists

DanStough:


areed (approved):
This works but it would be better to fix [higher up the stack](https://github.com/replicatedhq/kURL/blob/master/scripts/common/common.sh#L289).

Ideally you'd still be able to update proxy for docker with our script.

DanStough:
Makes sense. Let's see if this looks better.

I debated making this it's own special case, but we're already following this code path for re-installs on the same machine.

areed (approved):

PR #1204, opened 2021-02-18
feat: experimental k3s support

DanStough:
Tested with Registry, Kotsadm, Velero, Minio and OpenEBS + local rancher provisioner (default). Registry in kotsadm works fine with airgap install. Only thing untested is doing a backup/restore. Will try tomorrow.

### DIFF
--- a/bundles/k3s-rhel8/Dockerfile
+++ b/bundles/k3s-rhel8/Dockerfile
@@ -1,0 +1,21 @@
+# THIS ONLY DOWNLOADS THE SELINUX POLICY, not the k3s cli
+FROM centos:8
+
+RUN yum install -y gettext # envsubst
+RUN yum install -y yum-utils # yumdownloader
+
+RUN mkdir -p /packages/archives
+
+ADD ./rancher-k3s-common.repo.tmpl /tmp/
+
+ENV rpm_site=rpm.rancher.io
+ENV k3s_rpm_channel=stable
+ENV maj_ver=8
+
+ARG K3S_VERSION
+
+RUN envsubst < /tmp/rancher-k3s-common.repo.tmpl > /etc/yum.repos.d/rancher-k3s-common.repo
+RUN yumdownloader --resolve --destdir=/packages/archives -y \
+    k3s-selinux
+
+CMD cp -r /packages/archives/* /out/
--- a/scripts/common/k3s.sh
+++ b/scripts/common/k3s.sh
@@ -1,8 +1,8 @@
-# Kurl Specific RKE Install
+# Kurl Specific K3S Install
 
-RKE2_SHOULD_RESTART=
+K3S_SHOULD_RESTART=
 
-function rke2_init() {
+function k3s_init() {
 #     logStep "Initialize Kubernetes"
 
 #     kubernetes_maybe_generate_bootstrap_token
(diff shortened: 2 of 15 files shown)


bundles/k3s-rhel8/Dockerfile
```diff
@@ -0,0 +1,20 @@
...
+
+ADD ./rancher-k3s-common.repo.tmpl /tmp/
+
+ENV rpm_site=rpm.rancher.io
+ENV k3s_rpm_channel=stable
+ENV maj_ver=8
+
+ARG K3S_VERSION
+
+RUN mkdir -p /packages/archives
```
emosbaugh, line 15: how come this isnt in rhel7?

scripts/common/k3s.sh
```diff
@@ -0,0 +1,597 @@
...
+
+    logStep "Load K3S images"
+
+    mkdir -p /var/lib/rancher/k3s/agent/images
+    gunzip -c $DIR/packages/k3s/${k3s_version}/assets/k3s-images.linux-amd64.tar.gz > /var/lib/rancher/k3s/agent/images/k3s-images.linux-amd64.tar
+
+    logSuccess "K3S images loaded"
+}
+
+k3s_server_setup_systemd_service() {
```
emosbaugh, line 521: function

scripts/common/k3s.sh
```diff
@@ -0,0 +1,597 @@
...
+EOF
+    
+    systemctl daemon-reload
+    systemctl start k3s-server.service 
+    systemctl enable k3s-server.service 
+
+    logSuccess "K3S Service created"
+}
+
+k3s_create_env_file() {
```
emosbaugh, line 571: function

scripts/common/k3s.sh
```diff
@@ -0,0 +1,597 @@
...
+    local fileK3sEnv=/etc/systemd/system/k3s-server.env
+    echo "Creating environment file ${fileK3sEnv}"
+    UMASK=$(umask)
+    umask 0377
+    env | grep '^K3S_' | tee ${fileK3sEnv} >/dev/null
+    env | egrep -i '^(NO|HTTP|HTTPS)_PROXY' | tee -a ${fileK3sEnv} >/dev/null
+    umask $UMASK
+}
+
+k3s_create_symlinks() {
```
emosbaugh, line 581: function

scripts/common/k3s.sh
```diff
@@ -0,0 +1,597 @@
...
+    local binDir=/usr/local/bin
+
+    for cmd in kubectl crictl ctr; do
```
emosbaugh, line 584: this is cool. would be nice for rke2

scripts/common/k3s.sh
```diff
@@ -0,0 +1,597 @@
...
+    # trap k8s_ctrl_c SIGINT # trap ctrl+c (SIGINT) and handle it by asking for a support bundle - only do this after k8s is installed
+    ${K8S_DISTRO}_addon_for_each addon_load
+    # init                              # See next line
+    k3s_init                            # TODO(dan): A mix of Kubeadm stuff and general setup.
+    apply_installer_crd
+    type create_registry_service &> /dev/null && create_registry_service # this function is in an optional addon and may be missing
+    ${K8S_DISTRO}_addon_for_each addon_install
+    # post_init                          # TODO(dan): more kubeadm token setup
+    # outro                              # See next line
+    k3s_outro                            # TODO(dan): modified this to remove kubeadm stuff
```
emosbaugh, line 411: missing package_cleanup

scripts/common/k3s.sh
```diff
@@ -0,0 +1,597 @@
...
+    fi
+
+    kubelet --version | grep -q "$(echo $k3s_version | sed "s/-/+/")"
+}
+
+function k3s_get_host_packages_online() {
+    local k3s_version="$1"
+
+    echo "Fetching k3s-${k3s_version}.tar.gz"
+    curl -LO "$DIST_URL/k3s-${k3s_version}.tar.gz"
```
emosbaugh, line 504: use package_download

bundles/k3s-rhel8/Dockerfile
DanStough: It is, [just in a slightly different order](https://github.com/replicatedhq/kURL/blob/feat/exp-k3s-support/bundles/k3s-rhel7/Dockerfile#L6), which is an artifact of the copy-paste from RKE2. I can make it uniform though.

scripts/common/k3s.sh
DanStough: Seems like the architecture may be different here AFAIK. RKE has actual binaries but[ K3S bundles other tools under their CLI.](https://github.com/k3s-io/k3s/blob/master/pkg/kubectl/main.go) Not sure if it would make as much sense.

I think there is a question if we should install our own tools or not. I don't know the right answer but my first thought it that deferring to rancher's tools will prevent incompat. issues.

emosbaugh: Got it

DanStough:
Ready for re-review.

emosbaugh (approved):

PR #1206, opened 2021-02-19
feat: add k3s to kurl web api

DanStough:
Add K3S to the web api.

Started breaking out tests into separate files.

### DIFF
--- a/web/src/installers/index.ts
+++ b/web/src/installers/index.ts
@@ -63,6 +63,17 @@ export const rke2ConfigSchema = {
   },
 }
 
+export interface K3SConfig {
+  version: string;
+}
+
+export const k3sConfigSchema = {
+  type: "object",
+  properties: {
+    version: { type: "string" },
+  },
+}
+
 export interface DockerConfig {
   version: string;
   s3Override?: string;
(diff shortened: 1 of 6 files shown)


web/src/installers/index.ts
```diff
@@ -1143,6 +1160,14 @@ export class Installer {
     if (this.spec.kubernetes && this.spec.rke2) {
       return { error: { message: `This spec contains both kubeadm and rke2, please specifiy only one Kubernetes distribution` } };
     }
+    if (this.spec.k3s) {
+      if (!Installer.hasVersion("k3s", this.spec.k3s.version) && !this.hasS3Override("k3s")) {
+        return { error: { message: `K3S version ${_.escape(this.spec.k3s.version)} is not supported` } };
+      }
+    }
+    if (this.spec.kubernetes && this.spec.k3s) {
```
emosbaugh, line 1168: well now dont you need to check all three?

emosbaugh (approved):
one comment but i think this is still mergeable without fixing

web/src/installers/index.ts
DanStough: Yeah, I guess all possible combinations which means I need to probably do something different here. Going to merge this as is and think about the best way to do it.

PR #1211, opened 2021-02-19
fix(web): escape quotes in installer

DanStough:


laverya (approved):

DanStough:
Shows the current failing case: https://testgrid.kurl.sh/run/docker-daemon-conf

testgrid/tgrun/pkg/instances/k8s1164_docker19034.go
```diff
@@ -26,7 +26,8 @@ func init() {
 					Version: "1.7.0",
 				},
 				Docker: &kurlv1beta1.Docker{
-					Version: "19.03.4",
+					Version:      "19.03.4",
+					DaemonConfig: "{\n  \"exec-opts\": [\"native.cgroupdriver=systemd\"],\n  \"default-address-pools\": [ {\"base\":\"10.69.0.0/16\",\"size\":16} ]\n}",
```
laverya, line 30: I would suggest using a multiline string here instead of \n

web/src/util/services/templates.ts
```diff
@@ -84,6 +84,6 @@ function manifestFromInstaller(i: Installer, kurlURL: string, replicatedAppURL:
     KURL_UTIL_IMAGE: kurlUtilImage,
     KURL_BIN_UTILS_FILE: kurlBinUtils,
     STEP_VERSIONS: `(${Installer.latestMinors().join(" ")})`,
-    INSTALLER_YAML: i.toYAML(),
+    INSTALLER_YAML: i.toYAML().replace(/\"/g, "\\\""),
```
laverya, line 87: doesn't this have the possibility of breaking valid yaml?

laverya (changes requested):

web/src/util/services/templates.ts
laverya: Or is there a "bash string escape" function we could be using instead?

laverya: also - shouldn't we be escaping double-quotes? (Like, `abc: "abc"` should become `abc: \"abc\"`)

if we want this to be a fully-general solution, that is

DanStough: This was for double quotes, are you thinking we should do single quotes as well?

laverya: Discussed elsewhere - we should make a "bashEscape" function that replaces `\` with `\\` and then `"` with `\"`

testgrid/tgrun/pkg/instances/k8s1164_docker19034.go
```diff
@@ -27,6 +27,10 @@ func init() {
 				},
 				Docker: &kurlv1beta1.Docker{
 					Version: "19.03.4",
+					DaemonConfig: `{
+	"exec-opts": ["native.cgroupdriver=systemd"],
+	"default-address-pools": [ {"base":"10.69.0.0/16","size":16} ]
```
DanStough, line 32: I got rid of the escapes here since this is how customers would post the specs.

web/src/util/services/templates.ts
```diff
@@ -75,6 +75,10 @@ interface Manifest {
   INSTALLER_YAML: string;
 }
 
+function bashStringEscape( unescaped :string): string {
+  return unescaped.replace(/[!"'\\]/g, "\\\$&");
```
DanStough, line 79: I tried using the soln from https://qntm.org/bash, but it was escaping whitespace characters and unreadable. I also tried an expanded set of characters including $, but this also seemed like overkill and I'm not sure it would be the desired behavior.

laverya: Would it be possible to get a test for this?

laverya: (if we even have a test framework here...)

laverya (approved):

web/src/util/services/templates.ts
DanStough: You want it, you got :D 

Not doing a anthing crazy but at least there's a placeholder file now for folks to add service tests.

laverya (approved):

PR #1227, opened 2021-02-22
add a newline to the end of common.sh to fix tasks.sh

laverya:
also run tasks.sh in testgrid to catch this in the future


laverya:
from the tasks.sh returned by staging today:
```
        journalctl -xe
        bail "Could not successfully restart systemd service $serviceName"
    fi

    echo "Service $serviceName restarted."
}# Gather any additional information required from the user that could not be discovered and was not
# passed with a flag

function prompts() {
    if [ -z "$PRIVATE_ADDRESS" ]; then
        promptForPrivateIp
```

DanStough (approved):

PR #1233, opened 2021-02-24
openebs 2.6.0

laverya:
Upgrades still WIP, but fresh installs work

### DIFF
--- a/addons/openebs/2.6.0/install.sh
+++ b/addons/openebs/2.6.0/install.sh
@@ -1,0 +1,265 @@
+function openebs_pre_init() {
+    if [ -z "$OPENEBS_NAMESPACE" ]; then
+        OPENEBS_NAMESPACE=openebs
+    fi
+    if [ -z "$OPENEBS_LOCALPV_STORAGE_CLASS" ]; then
+        OPENEBS_LOCALPV_STORAGE_CLASS=openebs-localpv
+    fi
+    if [ -z "$OPENEBS_CSTOR_STORAGE_CLASS" ]; then
+        OPENEBS_CSTOR_STORAGE_CLASS=openebs-cstor
+    fi
+    if [ -z "$OPENEBS_CSTOR_TARGET_REPLICATION" ]; then
+        OPENEBS_CSTOR_TARGET_REPLICATION="3"
+    fi
+}
+
+function openebs() {
+    local src="$DIR/addons/openebs/$OPENEBS_VERSION"
+    local dst="$DIR/kustomize/openebs"
+
+    render_yaml_file "$src/tmpl-kustomization.yaml" > "$dst/kustomization.yaml"
+    render_yaml_file "$src/tmpl-namespace.yaml" > "$dst/namespace.yaml"
+    cp "$src/operator.yaml" "$dst/"
+
+    openebs_upgrade
+
+    if [ "$OPENEBS_LOCALPV" = "1" ]; then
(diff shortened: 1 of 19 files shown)


addons/openebs/2.6.0/operator.yaml
```diff
@@ -0,0 +1,791 @@
...
+# 4. set OPENEBS_IO_CREATE_DEFAULT_STORAGE_CONFIG to "false"
+# This manifest deploys the OpenEBS control plane components, with associated CRs & RBAC rules
+# NOTE: On GKE, deploy the openebs-operator.yaml in admin context
+
+# Create Maya Service Account
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: openebs-maya-operator
+  namespace: openebs
```
DanStough, line 14: Namespace -> I would expect all these values to be the template value from the kurl spec. If I compare to 1.12, all the ns fields were removed in 1.12.

addons/openebs/2.6.0/operator.yaml
```diff
@@ -0,0 +1,791 @@
...
+        - name: OPENEBS_IO_VOLUME_MONITOR_IMAGE
+          value: "openebs/m-exporter:2.6.0"
+        - name: OPENEBS_IO_CSTOR_POOL_EXPORTER_IMAGE
+          value: "openebs/m-exporter:2.6.0"
+        - name: OPENEBS_IO_HELPER_IMAGE
+          value: "openebs/linux-utils:2.6.0"
+        # OPENEBS_IO_ENABLE_ANALYTICS if set to true sends anonymous usage
+        # events to Google Analytics
+        - name: OPENEBS_IO_ENABLE_ANALYTICS
+          value: "true"
```
DanStough, line 209: Should be false like 1.12

addons/openebs/2.6.0/install.sh
```diff
@@ -0,0 +1,265 @@
...
+
+        kubectl apply -k "$dst/"
+
+        if ! openebs_cstor_pool_count "$OPENEBS_CSTOR_MAX_POOLS"; then
+            if [ "$OPENEBS_CSTOR_MAX_POOLS" = "1" ]; then
+                printf "${GREEN}Waiting for 1 disk${NC}\n"
+            else
+                printf "${GREEN}Waiting for ${OPENEBS_CSTOR_MAX_POOLS} disks${NC}\n"
+            fi
+            spinner_until 300 openebs_cstor_pool_count "$OPENEBS_CSTOR_MAX_POOLS"
```
DanStough, line 78: Don't we usually wait for someone to attach the disk in Rook? Not opposed to this just a behavior change.

laverya: Yeah, but this is what we've done so far with openebs

DanStough: Wasn't this indefinitely waiting though: https://github.com/replicatedhq/kURL/blob/f64d562cdb582974f1967bf9eaa6e8cc53377721/addons/openebs/1.12.0/install.sh#L86

Again, I wouldn't let this hold up me approving the PR.

laverya: oh, that's what you meant
this just makes it so that if nothing attaches the disk within 5 minutes it's a failure

I can move it back to 'forever' if you want

DanStough: NM, It's cool. In the words of Big Krit `4eva Is a Mighty Long Time`.

DanStough (approved):

PR #1235, opened 2021-02-25
Built-in host preflights

emosbaugh:
Screen Shot 2021-02-25 at 12 31 16 PM


pkg/preflight/assets/host-preflights.yaml
```diff
@@ -0,0 +1,37 @@
+apiVersion: troubleshoot.sh/v1beta2
```
emosbaugh, line 1: this is the builtin preflight spec

.github/workflows/build-test.yaml
```diff
@@ -14,7 +14,7 @@ jobs:
     steps:
       - uses: actions/setup-go@v2.1.3
         with:
-          go-version: "1.14.2"
```
laverya, line 17: maybe make these 1.16.0 too?

laverya (approved):

markpundsack:
I hope you can differentiate between warnings and errors. Should we treat them differently? Like warnings default to Y, and errors default to N?

Does kURL already have an "unattended" mode of some kind? I'd argue warnings should not stop unattended installs, whereas errors should, unless you've also specified something to install even with errors.

markpundsack:
How does this work with the existing kurl preflights? IIRC, there's something later on about running preflights. It might be confusing to have two different preflights run.

emosbaugh:
@markpundsack 

Failures will exit the install unless `ignore-preflights` is set.

Warnings will prompt to continue unless `ignore-preflight-warnings` or `ignore-preflights` is set. As of now, unattended installs will need to use the `ignore-preflight-warnings` flag.

I can detect no tty and continue automatically on warnings. Do you want to discuss the desired behavior or would you like me to make this change?

Existing kurl preflights will run before host preflights. Linked is a list of preflights that currently run.
https://github.com/replicatedhq/kURL/blob/master/scripts/common/preflights.sh#L3-L13

emosbaugh:
I've made it so that it will detect no tty and continue on warnings.

kurl_util/go.mod
```diff
@@ -20,4 +20,4 @@ require (
 
 replace github.com/replicatedhq/kurl => ../
 
-replace github.com/replicatedhq/kurl/kurl_util => ../kurl_util
+replace github.com/replicatedhq/kurl/kurlkinds => ../kurlkinds/
```
emosbaugh, line 23: @DanStough this will fix the vendoring issue

emosbaugh:
The build will fail until I can vendor in an updated troubleshoot.

pkg/cli/preflight.go
```diff
@@ -0,0 +1,154 @@
...
+		RunE: func(cmd *cobra.Command, args []string) error {
+			installerSpec, err := installer.RetrieveSpec(cli.GetFS(), args[0])
+			if err != nil {
+				return errors.Wrap(err, "retrieve installer spec")
+			}
+
+			builtin := preflight.Builtin()
+			data := installer.TemplateData{
+				Installer: *installerSpec,
+				IsPrimary: viper.GetBool("is-primary"),
```
emosbaugh, line 37: I dont love all these flags but i cant think of a better way to do it. Open to suggestions.

emosbaugh:
NOTES:
- change warning prompt to default yes
- change preflight failure to prompt with default no
- suggest the flag in the failure message

markpundsack:
Some subtleties:
1. In the screenshot, it shows a warning for CPU cores saying 4 are required. Does that mean the machine doesn't have 4, but it's just a warning and not an error? Or does the machine have more than 4, but below a recommended amount? Either way, we should be clear.
2. For disk space, we say `/var/lib/kubelet` has space available, but that directly may not actually exist yet. Even if it did, directories don't really have disk free, the disks where the directories exist have space. Is there a way we could phrase it more accurately? Like, the disk where `/var/lib/kubelet` will be installed has space available?
3. The first three warnings are disturbing, but especially the message about failing to get unit file state. Seems completely meaningless and ignorable. Can we hide it?

emosbaugh:
> * In the screenshot, it shows a warning for CPU cores saying 4 are required. Does that mean the machine doesn't have 4, but it's just a warning and not an error? Or does the machine have more than 4, but below a recommended amount? Either way, we should be clear.

based on the latest code change it will be an error, not a warning.
there is only a failure threshold at 4 and no warning threshold.

> * For disk space, we say `/var/lib/kubelet` has space available, but that directly may not actually exist yet. Even if it did, directories don't really have disk free, the disks where the directories exist have space. Is there a way we could phrase it more accurately? Like, the disk where `/var/lib/kubelet` will be installed has space available?

i will rephrase

> 3\. The first three warnings are disturbing, but especially the message about failing to get unit file state. Seems completely meaningless and ignorable. Can we hide it?

we have an internal issue 28767 related to this message. ive prioritized it

pkg/preflight/assets/host-preflights.yaml
```diff
@@ -0,0 +1,66 @@
...
+    - diskUsage:
+        checkName: "Ephemeral Disk Usage"
+        collectorName: "Ephemeral Disk Usage"
+        outcomes:
+          - warn:
+              when: "used/total > 70%"
+              message: /var/lib/kubelet is more than 70% full
+          - fail:
+              when: "total < 30Gi"
+              message: /var/lib/kubelet has less than 30Gi of total space
```
areed, line 45:
```suggestion
          - fail:
              when: "total < 30Gi"
              message: /var/lib/kubelet has less than 30Gi of total space
          - fail:
              when: "used/total > 80%"
              message: /var/lib/kubelet is more than 80% full
          - warn:
              when: "used/total > 60%"
              message: /var/lib/kubelet is more than 60% full
```

pkg/preflight/assets/host-preflights.yaml
```diff
@@ -0,0 +1,66 @@
...
+          - pass:
+              message: /var/lib/kubelet has at least 30Gi disk space available
+    - tcpLoadBalancer:
+        checkName: "Kubernetes API Server Load Balancer"
+        collectorName: "Kubernetes API Server Load Balancer"
+        exclude: '{{ and .Installer.Spec.Kubernetes.LoadBalancerAddress .IsPrimary (not .IsJoin) (not .IsUpgrade) | not }}'
+        outcomes:
+          - warn: # fail
```
areed, line 53: What do these fail comments mean?

pkg/preflight/assets/host-preflights.yaml
```diff
@@ -0,0 +1,66 @@
...
+              when: "connection-refused"
+              message: Connection to {{ .Installer.Spec.Kubernetes.LoadBalancerAddress }} via load balancer was refused.
+          - warn: # fail
+              when: "connection-timeout"
+              message: Timed out connecting to {{ .Installer.Spec.Kubernetes.LoadBalancerAddress }} via load balancer. Check your firewall.
+          - warn: # fail
+              when: "error"
+              message: Unexpected port status
```
areed, line 61:
```suggestion
              message: Unexpected port status
          - warn:
              when: "address-in-use"
              message: Port 6443 is unavailable
```

scripts/common/prompts.sh
```diff
@@ -200,6 +200,8 @@ promptForLoadBalancerAddress() {
     if [ -z "$LOAD_BALANCER_PORT" ]; then
         LOAD_BALANCER_PORT=6443
     fi
+
+    $BIN_BASHTOYAML -c $MERGED_YAML_SPEC -f "load-balancer-address=${LOAD_BALANCER_ADDRESS}:${LOAD_BALANCER_PORT}"
```
areed, line 204: This means the merged yaml spec will always have a load balancer address and the tcp load balancer check will always run, even on non-ha clusters, right?

areed (approved):

pkg/preflight/assets/host-preflights.yaml
emosbaugh: i guess if i was confident in this check i would change those to fail

scripts/common/prompts.sh
emosbaugh: we only prompt for a load balancer in ha clusters

https://github.com/replicatedhq/kURL/blob/28855945e6db7addf65fe5eeab47899fe3c102d5/scripts/common/prompts.sh#L12-L14

```bash
    if [ "$HA_CLUSTER" = "1" ]; then
        promptForLoadBalancerAddress
    fi
```

emosbaugh: To clarify, we template the preflight spec with the merged yaml spec. If we dont merge in values from cli flags and prompts then its not possible to base conditional checks based on user input.

PR #1238, opened 2021-02-25
openebs 2.6.0 upgrade support

laverya:


testgrid/tgrun/pkg/scheduler/types/types.go
```diff
@@ -35,8 +35,8 @@ type OperatingSystemImage struct {
 }
 
 type Instance struct {
-	InstallerSpec    InstallerSpec
-	UpgradeSpec      *InstallerSpec
+	InstallerSpec    InstallerSpec  `json:"installerSpec" yaml:"installerSpec"`
+	UpgradeSpec      *InstallerSpec `json:"upgradeSpec,omitempty" yaml:"upgradeSpec,omitempty"`
 	UnsupportedOSIDs []string
```
DanStough, line 40: I think it would be nice to have support for this too, then we could add some tests for RKE, but not essential in this PR.

addons/openebs/2.6.0/install.sh
```diff
@@ -253,13 +259,247 @@ function openebs_upgrade() {
...
+function openebs_spc_to_cspc() {
+    # upgrade job from https://github.com/openebs/upgrade/blob/v2.6.0/examples/migrate/spc-migration.yaml
+
+    local pools
+    pools=$(kubectl get spc --template '{{range .items}}{{.metadata.name}}{{"\n"}}{{end}}')
+
+    for pool in $pools; do
+        local spcCurrentVer=$(kubectl get spc $pool -o jsonpath='{.versionDetails.status.current}')
+        logSubstep "Upgrading $pool from SPC to CSPC"
+        local out_file=/tmp/openebs-spc-cspc-$pool.yaml
```
DanStough, line 303: Should we put this in the kurl install directory (three affected files here)? No strong opinion.

addons/openebs/2.6.0/install.sh
```diff
@@ -253,13 +259,247 @@ function openebs_upgrade() {
...
+    logSubstep "OpenEBS batch job to upgrade cstor volumes completed."
+}
+
+# returns true when a job has completed
+# TODO: remove before merging
+function job_is_completed() {
+  local namespace="$1"
+  local jobName="$2"
+  kubectl get jobs -n "$namespace" "$jobName" | grep -q '1/1'
+}
```
DanStough, line 505: Guessing you can delete this now the `common.sh` is the permanent home.

DanStough (approved):

testgrid/tgrun/pkg/scheduler/types/types.go
laverya: what do you mean?

DanStough: `UnsupportedOSIDs`

addons/openebs/2.6.0/install.sh
DanStough: NM, I just realized these files are tmp not var.

addons/openebs/2.6.0/install.sh
laverya: not if I want it to keep working in testgrid

testgrid/tgrun/pkg/scheduler/types/types.go
laverya: ohhhh
yeah, I can add that

DanStough (approved):

PR #1243, opened 2021-02-26
feat(longhorn): add alpha longhorn addon

DanStough:


addons/longhorn/1.1.0/ui-deployment.yaml
```diff
@@ -0,0 +1,31 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  labels:
+    app: longhorn-ui
+  name: longhorn-ui
+  namespace: longhorn-system
+spec:
+  replicas: 1
```
laverya, line 9: should we leave this scaled to 0 by default, or is it actually useful?

Also, does it have auth? If it doesn't have auth, definitely scale to 0 by default

laverya (approved):

addons/longhorn/1.1.0/ui-deployment.yaml
DanStough: Yeah it doesn't have auth. In this case should the number of replicas be in the kurl spec, or would we expect folks only to do this in isolated cases.

And yes, I think it's very useful :)

laverya (approved):

PR #1255, opened 2021-03-03
fix(contour): remove image prefix for docker ee

DanStough:
`docker.io` is in a lot of the bundle Manifests, but usually not in the specs.

I kept the job change out of 1.13.1 since it's only a couple days old and I don't want to carry this change into all future version of contour.


addons/contour/1.10.1/contour.yaml
```diff
@@ -1351,7 +1351,7 @@ rules:
 apiVersion: batch/v1
 kind: Job
 metadata:
-  name: contour-certgen-v1.10.1
+  name: contour-certgen-v1.10.1-1
```
DanStough, line 1354: Changed these to prevent immutability errors with the new image name when running this against an existing cluster.

addons/contour/1.10.1/patches/job-image.yaml
```diff
@@ -2,12 +2,19 @@
...
       containers:
       - name: contour
-        image: docker.io/projectcontour/contour:v1.10.1
+        image: projectcontour/contour:v1.10.1
         imagePullPolicy: IfNotPresent
+        command:
+        - contour
+        - certgen
+        - --kube
+        - --incluster
```
DanStough, line 18: This removes the `overwrite` flag to prevent any issues of making a new cert while contour pods are already running.

DanStough:
Summary of new changes:
- Moved batch jobs out of `contour.yaml` and into `batch.yaml`
- Consolidated the job-patch into the new file
- Added a section that only applies the (new) batch job if it's not already detected for that version.
- Blocked Contour automated PRs for 1.13.1

areed (approved):

PR #1260, opened 2021-03-04
Any primary

areed:
Fixes the bug that requires all kurl upgrades be initiated on the first primary. The problem was that a config file was being passed to `kubeadm upgrade` and the existing file on primaries other than the first contained a JoinConfiguration.

The solution was to stop passing any config file to `kubeadm upgrade`. Kubeadm prints out a warning to not use a config file for upgrades. When no config is passed the existing config is used that is stored in the cluster.

### DIFF
--- a/bundles/helm/Dockerfile
+++ b/bundles/helm/Dockerfile
@@ -1,6 +1,5 @@
 FROM ubuntu:18.04
-RUN apt-get update
-RUN apt-get -y install curl
+RUN apt-get update && apt-get -y install curl
 
 RUN mkdir -p /helm
 WORKDIR /helm
--- a/scripts/common/kubernetes.sh
+++ b/scripts/common/kubernetes.sh
@@ -354,7 +354,7 @@ function discover_service_subnet() {
         excluded="$excluded,${PRIVATE_ADDRESS}/16"
     fi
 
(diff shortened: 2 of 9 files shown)


bundles/helm/Dockerfile
```diff
@@ -1,6 +1,5 @@
 FROM ubuntu:18.04
-RUN apt-get update
-RUN apt-get -y install curl
+RUN apt-get update && apt-get -y install curl
```
areed, line 2: Not an optimization, it wasn't building for me until I combined into a single line

scripts/common/kubernetes.sh
```diff
@@ -354,7 +354,7 @@ function discover_service_subnet() {
         excluded="$excluded,${PRIVATE_ADDRESS}/16"
     fi
 
-    EXISTING_SERVICE_CIDR=$(kubeadm config view 2>/dev/null | grep serviceSubnet | awk '{ print $2 }')
+    EXISTING_SERVICE_CIDR=$(maybe kubeadm_cluster_configuration | grep serviceSubnet | awk '{ print $2 }')
```
areed, line 357: Get rid of deprecation warnings when using `kubeadm config view`

emosbaugh (approved):

PR #1263, opened 2021-03-04
Host preflight warnings

emosbaugh:


pkg/preflight/assets/host-preflights.yaml
```diff
@@ -20,17 +20,23 @@ spec:
...
         checkName: "Number of CPUs"
         outcomes:
-          - fail:
+          - warn:
               when: "count < 4"
-              message: At least 4 CPU cores are required
+              message: At least 4 CPU cores are recommended
+          - fail:
+              when: "count < 2"
+              message: At least 2 CPU cores are required
```
markpundsack, line 28:
```suggestion
              message: At least 2 CPU cores are required, and 4 CPU cores are recommended
```

pkg/preflight/assets/host-preflights.yaml
```diff
@@ -20,17 +20,23 @@ spec:
...
         checkName: "Amount of Memory"
         outcomes:
-          - fail:
+          - warn:
               when: "< 8Gi"
-              message: At least 8Gi of memory is required
+              message: At least 8Gi of memory is recommended
+          - fail:
+              when: "< 4Gi"
+              message: At least 4Gi of memory is required
```
markpundsack, line 39:
```suggestion
              message: At least 4Gi of memory is required, and 8Gi of memory is recommended
```

pkg/preflight/assets/host-preflights.yaml
markpundsack: It's a poor experience to be told a minimum, then reconfigure the machine to meet the minimum, only to be then told a higher recommended number, so just give them both at the same time.

My only question is, will it be clear to the user that they're below both? Does it say how many cores we detect?

pkg/preflight/builtin_test.go
```diff
@@ -48,17 +48,23 @@ spec:
...
         checkName: "Number of CPUs"
         outcomes:
-          - fail:
+          - warn:
               when: "count < 4"
-              message: At least 4 CPU cores are required
+              message: At least 4 CPU cores are recommended
+          - fail:
+              when: "count < 2"
+              message: At least 2 CPU cores are required
```
markpundsack, line 56:
```suggestion
              message: At least 2 CPU cores are required, and 4 CPU cores are recommended
```

pkg/preflight/builtin_test.go
```diff
@@ -48,17 +48,23 @@ spec:
...
         checkName: "Amount of Memory"
         outcomes:
-          - fail:
+          - warn:
               when: "< 8Gi"
-              message: At least 8Gi of memory is required
+              message: At least 8Gi of memory is recommended
+          - fail:
+              when: "< 4Gi"
+              message: At least 4Gi of memory is required
```
markpundsack, line 67:
```suggestion
              message: At least 4Gi of memory is required, and 8Gi of memory is recommended
```

markpundsack: Could also do with with parens: `(8Gi memory recommended)`

areed (approved):

PR #1264, opened 2021-03-04
fix(scripts): check tty for preflights

DanStough:


scripts/common/preflights.sh
```diff
@@ -359,9 +359,17 @@ function host_preflights() {
     if [ "${PREFLIGHT_IGNORE}" = "1" ]; then
         "${DIR}"/bin/kurl host preflight "${MERGED_YAML_SPEC}" ${opts} || true
     else
-        if ! "${DIR}"/bin/kurl host preflight "${MERGED_YAML_SPEC}" ${opts} </dev/tty ; then
-            printf "${RED}Host preflights have failures. Do you want to proceed anyway? ${NC} "
-            if ! confirmN "-t 30"; then
+        # interactive terminal
+        if [ -t 1 ] ; then
```
emosbaugh, line 363: i think this should be `if [ -t 0 ]` for stdin

emosbaugh: also should remove the check on line 345 or change to 0
https://github.com/replicatedhq/kURL/pull/1264/files#diff-6777059f94085a391eecf73c216ffd180069451ac33f4d92cbe8ce2660cf514eR345

emosbaugh (changes requested):

emosbaugh (approved):

PR #1268, opened 2021-03-05
CI: Fix Auto PR Names and Block Contour 1.13.1 Update

DanStough:


areed (approved):
If you add `on.workflow_dispatch: {}` then you can trigger from the GitHub UI to test.

PR #1276, opened 2021-03-06
Automate velero PR

areed:
[Example PR](https://github.com/replicatedhq/kURL/pull/1277)

### DIFF
--- a/addons/velero/template/base/install.sh
+++ b/addons/velero/template/base/install.sh
@@ -1,0 +1,115 @@
+
+function velero_pre_init() {
+    if [ -z "$VELERO_NAMESPACE" ]; then
+        VELERO_NAMESPACE=velero
+    fi
+    if [ -z "$VELERO_LOCAL_BUCKET" ]; then
+        VELERO_LOCAL_BUCKET=velero
+    fi
+}
+
+function velero() {
+    local src="$DIR/addons/velero/$VELERO_VERSION"
+    local dst="$DIR/kustomize/velero"
+
+    cp "$src/kustomization.yaml" "$dst/"
+
+    velero_binary
+
+    velero_install "$src" "$dst"
+
+    velero_patch_restic_privilege "$src" "$dst"
+
+    kubectl apply -k "$dst"
+
+    kubectl label -n default --overwrite service/kubernetes velero.io/exclude-from-backup=true
+}
+
+function velero_join() {
+    velero_binary
+}
+
+function velero_install() {
+    local src="$1"
+    local dst="$2"
+
+    # Pre-apply CRDs since kustomize reorders resources. Grep to strip out sailboat emoji.
+    $src/assets/velero-v${VELERO_VERSION}-linux-amd64/velero install --crds-only | grep -v 'Velero is installed'
+
+    local resticArg="--use-restic"
+    if [ "$VELERO_DISABLE_RESTIC" = "1" ]; then
+        resticArg=""
+    fi
+
+    local bslArgs="--no-default-backup-location"
+    if ! kubernetes_resource_exists "$VELERO_NAMESPACE" backupstoragelocation default; then
+        bslArgs="--provider aws --bucket $VELERO_LOCAL_BUCKET --backup-location-config region=us-east-1,s3Url=${OBJECT_STORE_CLUSTER_HOST},publicUrl=http://${OBJECT_STORE_CLUSTER_IP},s3ForcePathStyle=true"
+    fi
+
+    velero_credentials
+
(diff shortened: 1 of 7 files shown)


DanStough (approved):

addons/velero/template/base/install.sh
```diff
@@ -0,0 +1,117 @@
...
+        return 0
+    fi
+
+    if [ "${K8S_DISTRO}" != "rke2" ]; then
+        return 0
+    fi
+
+    render_yaml_file "$src/restic-daemonset-privileged.yaml" > "$dst/restic-daemonset-privileged.yaml"
+    inserpt_patches_strategic_merge "$dst/kustomization.yaml" restic-daemonset-privileged.yaml
+}
```
DanStough, line 97: Do we need `VELERO_RESTIC_REQUIRES_PRIVILEGED` anymore if this is how we manage it? I like that you don't need to remember to set it.

DanStough: If we decide to take it out (now or at a later point) need to update the RKE2 Testgrid run.

areed: I didn't realize that was part of the kurl spec. I'll restore it.

addons/velero/template/base/install.sh
```diff
@@ -0,0 +1,115 @@
...
+    $src/assets/velero-v${VELERO_VERSION}-linux-amd64/velero install --crds-only | grep -v 'Velero is installed'
+
+    local resticArg="--use-restic"
+    if [ "$VELERO_DISABLE_RESTIC" = "1" ]; then
+        resticArg=""
+    fi
+
+    local bslArgs="--no-default-backup-location"
+    if ! kubernetes_resource_exists "$VELERO_NAMESPACE" backupstoragelocation default; then
+        bslArgs="--provider aws --bucket $VELERO_LOCAL_BUCKET --backup-location-config region=us-east-1,s3Url=${OBJECT_STORE_CLUSTER_HOST},publicUrl=http://${OBJECT_STORE_CLUSTER_IP},s3ForcePathStyle=true"
```
salahalsaleh, line 46: `s3ForcePathStyle=true` most probably needs quotes around `true`

addons/velero/template/base/install.sh
```diff
@@ -0,0 +1,115 @@
...
+    fi
+
+    velero_credentials
+
+    $src/assets/velero-v${VELERO_VERSION}-linux-amd64/velero install \
+        $resticArg \
+        $bslArgs \
+        --plugins velero/velero-plugin-for-aws:v1.1.0,velero/velero-plugin-for-gcp:v1.1.0,velero/velero-plugin-for-microsoft-azure:v1.1.0 \
+        --secret-file velero-credentials \
+        --use-volume-snapshots=false \
```
salahalsaleh, line 56: we use `restic` to backup volume data, so i believe this needs to use the default `true` value

addons/velero/template/base/install.sh
```diff
@@ -0,0 +1,115 @@
...
+    fi
+
+    local bslArgs="--no-default-backup-location"
+    if ! kubernetes_resource_exists "$VELERO_NAMESPACE" backupstoragelocation default; then
+        bslArgs="--provider aws --bucket $VELERO_LOCAL_BUCKET --backup-location-config region=us-east-1,s3Url=${OBJECT_STORE_CLUSTER_HOST},publicUrl=http://${OBJECT_STORE_CLUSTER_IP},s3ForcePathStyle=true"
+    fi
+
+    velero_credentials
+
+    $src/assets/velero-v${VELERO_VERSION}-linux-amd64/velero install \
```
salahalsaleh, line 51: this install command is missing the `--use-restic` flag

addons/velero/template/base/install.sh
areed: Volumes snapshots usually integrate with a cloud provider, e.g. EBS on AWS. We don't have a VolumeSnapshotLocation in earlier versions of velero on kurl.

salahalsaleh: ah, i see.

salahalsaleh (approved):

addons/velero/template/base/install.sh
areed: It seems to end up quoted in kubectl
```
spec:
  config:
    publicUrl: http://10.96.1.229
    region: us-east-1
    s3ForcePathStyle: "true"
    s3Url: http://rook-ceph-rgw-rook-ceph-store.rook-ceph
  objectStorage:
    bucket: velero
  provider: aws
```

salahalsaleh: nice, just wanted to confirm, thanks!

PR #1290, opened 2021-03-08
blockDevices built-in preflight check for rook and openebs + cstor

emosbaugh:


pkg/preflight/assets/host-preflights.yaml
```diff
@@ -57,7 +59,7 @@ spec:
     - tcpLoadBalancer:
         checkName: "Kubernetes API Server Load Balancer"
         collectorName: "Kubernetes API Server Load Balancer"
-        exclude: '{{ and .Installer.Spec.Kubernetes.LoadBalancerAddress .IsPrimary (not .IsJoin) (not .IsUpgrade) | not }}'
+        exclude: '{{ and .Installer.Spec.Kubernetes.Version .Installer.Spec.Kubernetes.LoadBalancerAddress .IsPrimary (not .IsJoin) (not .IsUpgrade) | not }}'
```
laverya, line 62: I have no idea what this means
any chance we could get a comment with a prettier description?

pkg/preflight/builtin_test.go
```diff
@@ -211,7 +176,39 @@ spec:
...
+							b, err := yaml.Marshal(v)
+							if assert.NoError(t, err) {
+								assert.Equal(t, yamlNormalize(t, []byte(q.value)), yamlNormalize(t, b))
+							}
+						}
+					}
+				}
+			}
 		})
 	}
```
laverya, line 204: I don't really understand why this test is being replaced
I might be misreading but the new test seems a lot different than the previous one, and doesn't seem to fit the rest of the PR to me

emosbaugh: the existing tests forced me to copy paste the entire yaml each new test case. the file started getting extremely long so i refactored the test so you can test an individual property each case

laverya (approved):

pkg/preflight/assets/host-preflights.yaml
emosbaugh: i added some comments

laverya (approved):

PR #1298, opened 2021-03-10
ci: add airgap flag to tgrun instance spec

DanStough:
Refactor to allow PR test specs to specific an airgap build without affecting the installer spec.


emosbaugh:
Looks like this was already done here https://github.com/replicatedhq/kURL/pull/1796/files#diff-4ca2fc25367cf73895d3d2aa8a495c2604278b59bcddb0e01133b8585777f2d0R43

PR #1311, opened 2021-03-11
Host preflight tcp connect k8s api

emosbaugh:


pkg/preflight/assets/host-preflights.yaml
```diff
@@ -276,3 +280,23 @@ spec:
...
+        collectorName: "Kubernetes API TCP Connection Status"
+        exclude: '{{ and .Installer.Spec.Kubernetes.Version .Installer.Spec.Kubernetes.MasterAddress .IsJoin | not }}'
+        outcomes:
+          - fail:
+              when: "connection-refused"
+              message: Connection to the Kubernetes API at address {{ .Installer.Spec.Kubernetes.MasterAddress }} was refused
+          - fail:
+              when: "connection-timeout"
+              message: Timed out connecting to the Kubernetes API at address {{ .Installer.Spec.Kubernetes.MasterAddress }}
+          - fail:
```
areed, line 294: This collector never attempts to bind

areed (approved):

pkg/preflight/assets/host-preflights.yaml
emosbaugh: I took this preflight from your host preflights google doc. should i be using something else as reference?

areed: That's the main reference

PR #1312, opened 2021-03-12
Update KUBECONFIG unset to echo into .bash_profile instead of .profile

mjneth:
Using .profile to unset the KUBECONFIG env var isn't working:
```
[vagrant@hostname ~]$ cat .profile
unset KUBECONFIG
[vagrant@hostname ~]$ bash -l
[vagrant@hostname ~]$ kubectl get nodes
error: error loading config file "/etc/kubernetes/admin.conf": open /etc/kubernetes/admin.conf: permission denied
```

but adding it to .bash_profile instead does work:
```
[vagrant@hostname ~]$ cat .bash_profile | grep KUBECONFIG
unset KUBECONFIG
[vagrant@hostname ~]$ bash -l
[vagrant@hostname ~]$ kubectl get nodes
NAME                                      STATUS   ROLES    AGE    VERSION
nodename   Ready    master   100m   v1.19.3
```

kURL seems to also try to set the permissions of `/etc/kubernetes/admin.conf` but that's not working:
```
[vagrant@ip-10-0-2-15 ~]$ ll /etc/kubernetes/admin.conf
-r--r-----. 1 root root 5565 Mar 12 15:07 /etc/kubernetes/admin.conf
[vagrant@ip-10-0-2-15 ~]$ id vagrant
uid=1000(vagrant) gid=1000(vagrant) groups=1000(vagrant)
```

Updating this output should help reduce issues with the customer being able to use kubectl after install.

### DIFF
--- a/scripts/common/common.sh
+++ b/scripts/common/common.sh
@@ -431,7 +431,7 @@ function kubeconfig_setup_outro() {
 
             printf "To access the cluster with kubectl, ensure the KUBECONFIG environment variable is unset:\n"
             printf "\n"
-            printf "${GREEN}    echo unset KUBECONFIG >> ~/.profile${NC}\n"
+            printf "${GREEN}    echo unset KUBECONFIG >> ~/.bash_profile${NC}\n"
             printf "${GREEN}    bash -l${NC}\n"
             return
         fi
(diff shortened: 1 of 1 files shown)


laverya (approved):
since this is printed we don't really have to handle the zsh etc scenarios - the rare person using zsh will know to change things :)

mjneth:
> since this is printed we don't really have to handle the zsh etc scenarios - the rare person using zsh will know to change things :)

@laverya I wondered if it should also say like "For bash, xxxxxxxxxx. If you use a different shell then see your shell's documentation for how to unset environment variables when you log in". Are other shells besides bash actually supported though?

laverya:
Shells besides bash aren't officially supported, really. So no need to do that. (I see no reason why they wouldn't work, but we aren't currently taking steps to support them)

PR #1314, opened 2021-03-12
Calico

areed:


web/src/installers/index.ts
```diff
@@ -129,6 +129,9 @@ export const calicoConfigSchema = {
   properties: {
     version: { type: "string" },
     s3Override: { type: "string", flag: "s3-override", description: "Override the download location for addon package distribution (used for CI/CD testing alpha addons)" },
+    podCIDR: { type: "string", description: "The subnet where pods will be found" },
+    podCidrRange: { type: "string", description: "The size of the CIDR where pods can be found" },
+    isEncryptionDisabled: { type: "boolean", description: "Is encryption in the Weave CNI disabled" },
```
DanStough, line 134: Should be `Calico CNI`

scripts/common/common.sh
```diff
@@ -286,6 +286,13 @@ function kubernetes_resource_exists() {
     kubectl -n "$namespace" get "$kind" "$name" &>/dev/null
 }
 
+function kubernetes_cluster_resource_exists() {
+    local kind=$1
+    local name=$2
+
+    kubectl -n "$namespace" get "$kind" "$name" &>/dev/null
+}
+
```
DanStough, line 295: Isn't this the same thing as the above `kubernetes_resource_exists`? Maybe you wanted to remove the NS field?

kurlkinds/pkg/apis/cluster/v1beta1/installer_types.go
```diff
@@ -215,8 +215,9 @@ type Ekco struct {
 }
 
 type Calico struct {
-	S3Override string `json:"s3Override,omitempty" yaml:"s3Override,omitempty"`
-	Version    string `json:"version" yaml:"version"`
+	S3Override           string `json:"s3Override,omitempty" yaml:"s3Override,omitempty"`
+	Version              string `json:"version" yaml:"version"`
+	IsEncryptionDisabled bool   `json:"isEncryptionDisabled,omitempty" yaml:"isEncryptionDisabled,omitempty"`
 }
```
DanStough, line 221: I think we need these too:
- podCidrRange
- podCIDR

addons/calico/3.18.1/ip-in-ip-encapsulation.yaml
```diff
@@ -0,0 +1,14 @@
...
+  namespace: kube-system
+spec:
+  template:
+    spec:
+      containers:
+        - name: calico-node
+          env:
+            # Enable IPIP
+            - name: CALICO_IPV4POOL_IPIP
+              value: "Off"
```
DanStough, line 14: Weird that they enable this with "off" (shrug)

addons/calico/3.18.1/install.sh
```diff
@@ -0,0 +1,99 @@
...
+function calico_cli() {
+    local src="$DIR/addons/calico/$CALICO_VERSION"
+
+    if [ ! -f "$src/assets/calicoctl" ] && [ ! "$AIRGAP" = "1" ]; then
+        mkdir -p "$src/assets"
+        curl -L https://github.com/projectcalico/calicoctl/releases/download/v3.18.0/calicoctl > "$src/assets/calicoctl"
+    fi
+
+    chmod +x "$src/assets/calicoctl"
+    mv "$src/assets/calicoctl" /usr/local/bin/
```
DanStough, line 61: Wouldn't this only apply in the non-airgap case? I didn't see calicoctl in the manifest.

DanStough (changes requested):
The only dealbreaker is I think POD CIDR  and POD CIDR RANGE need to be added to Kurlkinds.

Also just a note, we probably need add some kind of story to add a reset to `task.sh`.

DanStough (approved):
If you run `make generate` on the controller we're good.

PR #1315, opened 2021-03-12
Kotsadm http host preflights

emosbaugh:


addons/kotsadm/1.33.2/host-preflight.yaml
```diff
@@ -0,0 +1,26 @@
...
+kind: HostPreflight
+metadata:
+  name: kurl-builtin
+spec:
+  collectors:
+    - http:
+        collectorName: Can Access Replicated API
+        get:
+          url: https://replicated.app
+        exclude: '.Installer.Spec.Kurl.Airgap }}'
```
areed, line 11:
```suggestion
        exclude: '{{ .Installer.Spec.Kurl.Airgap }}'
```

addons/kotsadm/1.33.2/host-preflight.yaml
```diff
@@ -0,0 +1,26 @@
...
+
+  analyzers:
+    - http:
+        checkName: Can Access Replicated API
+        collectorName: Can Access Replicated API
+        exclude: '.Installer.Spec.Kurl.Airgap }}'
```
areed, line 17:
```suggestion
        exclude: '{{ .Installer.Spec.Kurl.Airgap }}'
```

addons/kotsadm/1.34.0/host-preflight.yaml
```diff
@@ -0,0 +1,26 @@
...
+kind: HostPreflight
+metadata:
+  name: kurl-builtin
+spec:
+  collectors:
+    - http:
+        collectorName: Can Access Replicated API
+        get:
+          url: https://replicated.app
+        exclude: '.Installer.Spec.Kurl.Airgap }}'
```
areed, line 11:
```suggestion
        exclude: '.{{ Installer.Spec.Kurl.Airgap }}'
```

addons/kotsadm/1.34.0/host-preflight.yaml
```diff
@@ -0,0 +1,26 @@
...
+
+  analyzers:
+    - http:
+        checkName: Can Access Replicated API
+        collectorName: Can Access Replicated API
+        exclude: '.Installer.Spec.Kurl.Airgap }}'
```
areed, line 17:
```suggestion
        exclude: '{{ .Installer.Spec.Kurl.Airgap }}'
```

addons/kotsadm/alpha/host-preflight.yaml
```diff
@@ -0,0 +1,26 @@
...
+kind: HostPreflight
+metadata:
+  name: kurl-builtin
+spec:
+  collectors:
+    - http:
+        collectorName: Can Access Replicated API
+        get:
+          url: https://replicated.app
+        exclude: '.Installer.Spec.Kurl.Airgap }}'
```
areed, line 11:
```suggestion
        exclude: '{{ .Installer.Spec.Kurl.Airgap }}'
```

addons/kotsadm/alpha/host-preflight.yaml
```diff
@@ -0,0 +1,26 @@
...
+
+  analyzers:
+    - http:
+        checkName: Can Access Replicated API
+        collectorName: Can Access Replicated API
+        exclude: '.Installer.Spec.Kurl.Airgap }}'
```
areed, line 17:
```suggestion
        exclude: '{{ .Installer.Spec.Kurl.Airgap }}'
```

pkg/cli/host_preflight.go
```diff
@@ -26,32 +28,49 @@ func NewHostPreflightCmd(cli CLI) *cobra.Command {
...
+			preflightSpec, err := decodePreflightSpec(builtin, data)
 			if err != nil {
-				return errors.Wrap(err, "execute installer template")
+				return errors.Wrap(err, "builtin")
 			}
 
-			preflightSpec, err := preflight.Decode(spec)
-			if err != nil {
-				return errors.Wrap(err, "decode spec")
+			for _, filename := range v.GetStringSlice("spec") {
```
areed, line 57: Could you comment on what's going on in the loop? I don't understand why the maybe append functions are comparing titles.

emosbaugh: I was thinking about adding registry http checks and thought there would be a lot of duplication since say lots of addons use docker hub. I am trying to avoid this but there are no ids in collectors and analyzers. I could make a change to troubleshoot to check that the checks are equivalent. What do you prefer here? I will additionally add a comment once we have an agreed upon solution.

areed: I see.  I think a comment here will be fine.

areed (approved):

pkg/cli/host_preflight.go
emosbaugh: The more i think about the approach here the more i dislike it. I may try to think of another solution.

areed (approved):

PR #1324, opened 2021-03-15
fix: update cri-tools 1.20

DanStough:


scripts/upgrade.sh
```diff
@@ -33,7 +33,6 @@ maybe_upgrade() {
     local kubeletMajor="$major"
     local kubeletMinor="$minor"
     local kubeletPatch="$patch"
-    local minorVersionDifference=$(($KUBERNETES_TARGET_VERSION_MINOR - $kubeletMinor))
```
DanStough, line 36: Unused

scripts/upgrade.sh
```diff
@@ -74,8 +73,6 @@ maybe_upgrade() {
         fi
 
         kubernetes_host
-        systemctl daemon-reload
-        systemctl restart kubelet
```
DanStough, line 78: Already included in `kubernetes_host`

DanStough:
Test configurations:
- Two node P-S fresh install of 1.19.7 on Ubuntu 18
- Two-node upgrade from 1.18.10 to 1.19.7 on Centos 7 using base script from kurl.sh

scripts/common/kubernetes.sh
```diff
@@ -61,7 +61,11 @@ function kubernetes_sysctl_config() {
 function kubernetes_install_host_packages() {
     k8sVersion=$1
 
-    logStep "Install kubelet, kubeadm, kubectl and cni host packages"
+    # ALWAYS update crictl: https://listman.redhat.com/archives/rhsa-announce/2019-October/msg00038.html
```
areed, line 64: 1. This won't work before `kubernetes_get_host_packages_online` for online installs
2. We only need to upgrade if they have an outdated crictl.

scripts/common/kubernetes.sh
```diff
@@ -72,30 +76,57 @@ function kubernetes_install_host_packages() {
...
         ;;
     esac
 
+    # Install Kubeadm from binary (see kubernetes.io)
+    cp -f "$DIR/packages/kubernetes/${k8sVersion}/assets/kubeadm" /usr/bin/
+    chmod a+rx /usr/bin/kubeadm
+
+    mkdir -p /etc/systemd/system/kubelet.service.d
+    cp -f "$DIR/tmp-kubeadm.conf" /etc/systemd/system/kubelet.service.d/10-kubeadm.conf
+    chmod 755 /etc/systemd/system/kubelet.service.d/10-kubeadm.conf
```
areed, line 121: I've got 0640 on my test install. I don't think it needs to be executable.

scripts/common/kubernetes.sh
```diff
@@ -61,7 +61,11 @@ function kubernetes_sysctl_config() {
 function kubernetes_install_host_packages() {
     k8sVersion=$1
 
-    logStep "Install kubelet, kubeadm, kubectl and cni host packages"
+    # ALWAYS update crictl: https://listman.redhat.com/archives/rhsa-announce/2019-October/msg00038.html 
+    tar -C /usr/bin -xzf "$DIR/packages/kubernetes/${k8sVersion}/assets/crictl-linux-amd64.tar.gz"
+    chmod a+rx /usr/bin/crictl
```
emosbaugh, line 66: is this necessary? does crictl talk to a daemon? if not can we work on limiting perms?

scripts/common/kubernetes.sh
```diff
@@ -72,30 +76,57 @@ function kubernetes_install_host_packages() {
...
+                    sed "s:__ENV_LOCATION__:sysconfig:g" -i "$DIR/tmp-kubeadm.conf"
                     rpm --upgrade --force --nodeps $DIR/packages/kubernetes/${k8sVersion}/rhel-7/*.rpm
                     ;;
             esac
         ;;
     esac
 
+    # Install Kubeadm from binary (see kubernetes.io)
+    cp -f "$DIR/packages/kubernetes/${k8sVersion}/assets/kubeadm" /usr/bin/
+    chmod a+rx /usr/bin/kubeadm
```
emosbaugh, line 117: what were the permissions prior to this commit?

scripts/common/kubernetes.sh
emosbaugh: looks the same, never mind

```
ethan@ethan-kurl-docker:~$ ls -l /usr/bin/kubeadm
-rwxr-xr-x 1 root root 39059456 Jan 13 19:26 /usr/bin/kubeadm
ethan@ethan-kurl-docker:~$ ls -l /usr/bin/crictl
-rwxr-xr-x 1 ubuntu ubuntu 26061879 Dec 13  2018 /usr/bin/crictl
```

scripts/common/kubernetes.sh
DanStough: I think I was going off: https://github.com/replicatedhq/kURL/blob/ff8fb27ad67a3164a83b4c5b990c61377f3c0469/scripts/common/upgrade.sh#L104

scripts/common/kubernetes.sh
DanStough: As I understand if crictl does talk to the kubelet and also to containerd. You still need elevated privileges to access the containerd socket though.

scripts/common/kubernetes.sh
DanStough: Assuming this is resolved per above

scripts/common/kubernetes.sh
DanStough: I think was an artifact of troubleshooting. I get 644 when I create a test file doing something similar so I'll make sure to change it to 640

scripts/common/kubernetes.sh
DanStough: Moved this after we grab packages. I didn't necessarily check the crictl version since we are always syncing what's in the bundle with the system if any one component is missing. Let me know if this doesn't resolve this comment.

areed (approved):

PR #1329, opened 2021-03-16
fix: improve interactive session detection

DanStough:


scripts/common/common.sh
```diff
@@ -1,6 +1,3 @@
-
-STORAGE_PROVISIONER=rook
-
```
DanStough, line 3: Unused

scripts/common/preflights.sh
```diff
@@ -365,14 +365,10 @@ function host_preflights() {
...
         # interactive terminal
-        if [ -t 0 ] ; then
+        if canPrompt; then      
             if ! "${DIR}"/bin/kurl host preflight "${MERGED_YAML_SPEC}" ${opts} </dev/tty ; then
-                printf "${RED}Host preflights have failures. Do you want to proceed anyway? ${NC} "
-                if ! confirmN "-t 30"; then
-                    report_install_fail "preflight"
-                    bail "Use the \"preflight-ignore\" flag to proceed."
-                fi
-                # TODO: report preflight fail
```
DanStough, line 375: `kurl` already prompts the user here, although the default is backward from what we wanted in the script (Yes is default in the binary)

scripts/common/common.sh
```diff
@@ -709,3 +706,12 @@ MACHINE_ID=
 function get_machine_id() {
     MACHINE_ID="$(${DIR}/bin/kurl host protectedid || true)"
 }
+
+function canPrompt() {
```
emosbaugh, line 710: can you make this snake case?

scripts/common/preflights.sh
```diff
@@ -365,14 +365,10 @@ function host_preflights() {
         # TODO: report preflight fail
     else
         # interactive terminal
-        if [ -t 0 ] ; then
+        if canPrompt; then
```
emosbaugh, line 368: remove trailing whitespace

scripts/common/preflights.sh
emosbaugh: but what if there are failures?

emosbaugh (changes requested):

scripts/common/common.sh
DanStough: Np. There are some other places we'll need to make these changes too like `prompts.sh`

scripts/common/preflights.sh
DanStough: Yes, I failed to consider that case. You'll get a double prompt if you hit No to exit on the warnings. I think the solution could be using different exits codes in `kurl`. I looked at this for a second and might need some help dealing with the errors package if we want to go that route. I reverted it for now.

scripts/common/common.sh
```diff
@@ -1,6 +1,3 @@
-
-STORAGE_PROVISIONER=rook
```
emosbaugh, line 2: Why is this going away?

scripts/common/common.sh
emosbaugh: Ya, I think for all new code we should stick to snake

scripts/common/common.sh
DanStough: `STORAGE_PROVISIONER` is unused anywhere in the code

emosbaugh (approved):

scripts/common/preflights.sh
```diff
@@ -365,16 +365,25 @@ function host_preflights() {
...
     else
         # interactive terminal
-        if [ -t 0 ] ; then
-            if ! "${DIR}"/bin/kurl host preflight "${MERGED_YAML_SPEC}" ${opts} </dev/tty ; then
-                printf "${RED}Host preflights have failures. Do you want to proceed anyway? ${NC} "
-                if ! confirmY "-t 10"; then
+        if can_prompt; then
+            set +e
+            "${DIR}"/bin/kurl host preflight "${MERGED_YAML_SPEC}" </dev/tty
+            local kurlExitCode=$?
```
emosbaugh, line 371: snake case?

emosbaugh (approved):

PR #1360, opened 2021-03-22
Antrea CNI

areed:


addons/antrea/0.13.1/Manifest
```diff
@@ -0,0 +1 @@
+image antrea-ubuntu projects.registry.vmware.com/antrea/antrea-ubuntu:v0.13.1
```
laverya, line 1: this scares me
does it work on non-ubuntu OSes?

laverya (approved):
Would it be possible to get a testgrid spec added using this?

Also a pr testgrid spec like [this](https://github.com/replicatedhq/kURL/blob/9c055140463a077e2009a3c33f46c94beef64fb1/addons/kotsadm/template/testgrid/k8s-docker.yaml) - specifically something that will actually use the network and fail if it doesn't work

addons/antrea/0.13.1/Manifest
areed: Yes, works on CentOS

PR #1368, opened 2021-03-24
Crictl revert revert

areed:


DanStough (approved):
Looks good. Did we need to change this so that it would update Crictl for non-k8s upgrades (just running the same installer again)?

scripts/Manifest
```diff
@@ -2,6 +2,7 @@ DISABLE_REPORTING=1
 KURL_URL=
 DIST_URL=
 INSTALLER_ID=
+CRICTL_VERSION=1.20.0
```
DanStough, line 5: I think this breaks k8s 1.15.X, which wasn't updated in the PR. My read of the [version compatibility chart](https://github.com/kubernetes-sigs/cri-tools#current-status) was that there isn't any crictl version with the CVE fix (>=1.18 IICR) that is compatible with k8s 1.15.x, but critools 1.20 is compatible with all k8s versions >=1.16. `semverCompare` will evaluate to -1 but then will try to unpack non-existent cri-tools. 

On second review I also see a potential problem moving between k8s 1.15 versions or new k8s 1.15 installs because that component won't be there either. Easy solution would be to just check for the tarball before you unpack, but there would be weird behavior with 1.15 that would always re-install the critools. That might be better than doing a version check in the code that goes away soon?

DanStough (changes requested):

scripts/Manifest
areed: Another problem is that kubeadm is missing in 1.15 bundles now.

I suggest we treat Kubernetes 1.15 the same as 1.16+ and install the latest crictl. I don't see any crictl commands failing with v1.20.0 on a Kubernetes 1.15.3 install and it's not a critical component anyway.

DanStough: I thought for all versions of k8s we were previously shipping kubeadm: https://github.com/replicatedhq/kURL/blob/master/bundles/k8s-ubuntu1804/Dockerfile#L21

Yeah, if there's not a compatibility issue that solution makes sense. Maybe 1.18 would be safer. I'm not really sure I understand why crictl is necessary but I thought I recall some posts about weird compatibility issues.

areed: I don't think crictl is necessary. I think most people use `kubectl`, `docker`, and `ctr` when working on the command line.

areed: I've tested crictl 1.20.0 with Kubernetes 1.15.3 install using docker and containerd. All looks good.

DanStough (approved):

PR #1372, opened 2021-03-25
CI scan images

emosbaugh:


.github/workflows/daily-scan-images.yaml
```diff
@@ -0,0 +1,50 @@
+name: daily-scan-images
+
+on:
+  # schedule:
+  # - cron: "0 16 * * *"
```
emosbaugh, line 5: I'm going to switch this to run daily

divolgin (approved):

web/src/installers/versions.js
```diff
@@ -0,0 +1,210 @@
...
+    "1.18.4",
+    "1.17.13",
+    "1.17.7",
+    "1.17.3",
+    "1.16.4",
+    "1.15.3",
+    "1.15.2",
+    "1.15.1",
+    "1.15.0",
+    "1.20.4", "1.20.2", "1.20.1", "1.20.0", // cron-kubernetes-update
```
laverya, line 19: the autoupdate scripts will all need to be updated to point at this filepath and not the previous one

emosbaugh: updated

PR #1379, opened 2021-03-27
More host preflights

areed:


emosbaugh (commented):
I was trying to avoid not writing valid yaml because that is what we expect of all other kots.io documents but I can see how that may be a problem

areed:
> I was trying to avoid not writing valid yaml because that is what we expect of all other kots.io documents but I can see how that may be a problem

I don't follow. Which kots.io documents have to be valid yaml?

emosbaugh:
All. Because otherwise we can't lint them

emosbaugh (approved):
The host preflight tool is starting to feel a little cumbersome with all these additional args. Lets keep going with it as is for now but it may be worth revisiting at some point.

emosbaugh:
Testgrid is still running

PR #1380, opened 2021-03-27
Automated Containerd version update 1.4.4

replicatedbot:
Automated changes by the [cron-containerd-update](https://github.com/replicatedhq/kURL/blob/master/.github/workflows/update-containerd.yaml) GitHub action

### DIFF
--- a/addons/containerd/1.3.7/install.sh
+++ b/addons/containerd/1.3.7/install.sh
@@ -37,7 +37,6 @@ function containerd_configure() {
 
     sed -i '/systemd_cgroup/d' /etc/containerd/config.toml
     sed -i '/containerd.runtimes.runc.options/d' /etc/containerd/config.toml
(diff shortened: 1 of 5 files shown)


laverya:
https://github.com/replicatedhq/kURL/commit/f992b258ebb00e4c195a6cddd39cd60faa3bd92a

I think we need to update the template

emosbaugh:
@laverya #1407 (cc @MikaelSmith for next time)

MikaelSmith:
So it should go at https://github.com/replicatedhq/kURL/blob/master/addons/containerd/template/base/install.sh#L39 instead?

emosbaugh:
We are working towards automating all add-on updates. They do not regenerate older versions so I would suggest adding it there as well as where you did.

emosbaugh:
Closing this. Will regenerate.

PR #1389, opened 2021-03-30
Rook ceph forked images

emosbaugh:


laverya (approved):

MikaelSmith:
Does anything make sure these updates get adopted in existing installs?

emosbaugh:
If the kurl script is re-run the cephcluster resource will be patched with the updated image. I am not sure if rook operator will roll all pods. If not, I would imagine if pods are deleted then they would be recreated with the updated image.

PR #1395, opened 2021-03-31
Use s3cmd image for backup and restore

emosbaugh:


divolgin (approved):

addons/registry/2.7.1/install.sh
```diff
@@ -11,6 +11,11 @@ function registry() {
         registry_object_store_bucket
         render_yaml_file "$DIR/addons/registry/2.7.1/tmpl-deployment-objectstore.yaml" > "$DIR/kustomize/registry/deployment-objectstore.yaml"
         insert_resources "$DIR/kustomize/registry/kustomization.yaml" deployment-objectstore.yaml
+
+        cp "$DIR/addons/registry/2.7.1/patch-deployment-velero.yaml" "$DIR/kustomize/registry/patch-deployment-velero.yaml"
+        insert_patches_strategic_merge "$DIR/kustomize/registry/kustomization.yaml" patch-deployment-velero.yaml
+        render_yaml_file "$DIR/addons/registry/2.7.1/tmpl-configmap-velero.yaml" > "$DIR/kustomize/registry/configmap-velero.yaml"
+        insert_resources "$DIR/kustomize/registry/kustomization.yaml" configmap-velero.yaml
```
laverya, line 18: so if I understand correctly, the object store install will get the s3cmd image, while other installs will get the default registry image?

laverya (approved):

addons/registry/2.7.1/install.sh
emosbaugh: im just trying not to change anything

PR #1401, opened 2021-04-01
automate prometheus updates

laverya:


addons/prometheus/template/values.yaml
```diff
@@ -0,0 +1,2421 @@
+# Default values for kube-prometheus-stack.
```
emosbaugh, line 1: are these all defaults? wanna add comments where there are changes?

addons/prometheus/template/generate.sh
```diff
@@ -0,0 +1,71 @@
...
+    # update names - 'prometheus-prometheus' to 'k8s' (as this has a PV attached)
+    sed -i "s/prometheus-prometheus/k8s/g" "../$VERSION/operator/default.yaml"
+
+    # get images in files
+    grep 'image: '  "../$VERSION/operator/default.yaml" | sed 's/ *image: "*/image name /' | sed 's/"//' | sort | uniq > "../$VERSION/Manifest"
+    grep 'image: '  "../$VERSION/operator/adapter.yaml" | sed 's/ *image: "*/image name /' | sed 's/"//' | sort | uniq >> "../$VERSION/Manifest"
+}
+
+function add_as_latest() {
+    sed -i "/cron-prometheus-update/a\      \"${VERSION}\"\," ../../../web/src/installers/versions.js
```
emosbaugh, line 54: 2 less spaces

addons/prometheus/template/generate.sh
```diff
@@ -0,0 +1,71 @@
...
+    cp -r ./base/* "../${VERSION}"
+    mkdir -p "../${VERSION}/crds"
+    mkdir -p "../${VERSION}/operator"
+
+    # run the helm commands
+    helm repo add prometheus-community https://prometheus-community.github.io/helm-charts
+    helm repo update
+
+    # get a copy of the stack with CRDs
+    helm template replaceme prometheus-community/kube-prometheus-stack --values ./values.yaml -n monitoring --include-crds > "../$VERSION/crds/crds-all.yaml"
```
emosbaugh, line 24: should you add --version arg here?

addons/prometheus/template/generate.sh
```diff
@@ -0,0 +1,71 @@
+#!/bin/bash
+
+set -euo pipefail
+
+VERSION=""
+function get_latest_release_version() {
+    VERSION=$(curl -I https://github.com/prometheus-operator/prometheus-operator/releases/latest | \
```
emosbaugh, line 7: does this version always track the helm version?

addons/prometheus/template/base/install.sh
```diff
@@ -0,0 +1,128 @@
...
+    if ! kubectl get prometheuses --all-namespaces &>/dev/null; then
+        return 1
+    fi
+    return 0
+}
+
+function prometheus_rook_ceph() {
+    local dst="$1"
+
+    if kubectl get ns | grep -q rook-ceph; then
```
emosbaugh, line 92: seems like this conditional would be better to wrap the function call. or at least reverse the conditional and return for less indentation

addons/prometheus/template/base/install.sh
```diff
@@ -0,0 +1,128 @@
...
+        cat <<EOF >> "$dst/operator/prometheus-roleBindingSpecificNamespaces.yaml"
```
emosbaugh, line 93: should you make these files and just cp src dst?

emosbaugh (changes requested):

addons/prometheus/template/values.yaml
laverya: can do - there's a fair few of them

addons/prometheus/template/base/install.sh
laverya: this code was all lifted from the previous version
I can make it cleaner

addons/prometheus/template/generate.sh
laverya: hmm
yeah, there'll be a gap before it gets updated
let me see if there's a better solution there

addons/prometheus/template/generate.sh
laverya: would we ever want a version other than latest?

addons/prometheus/template/values.yaml
laverya: there's a bunch of changes
I'll start adding comments

addons/prometheus/template/generate.sh
laverya: found a better solution

addons/prometheus/template/generate.sh
emosbaugh: well you were lifting the version from github but using the helm chart

addons/prometheus/template/base/install.sh
emosbaugh: thanks

addons/prometheus/template/generate.sh
laverya: ah. well, now they're both using latest :)

addons/prometheus/template/generate.sh
```diff
@@ -0,0 +1,71 @@
...
+
+function generate() {
+    # make the base set of files
+    mkdir -p "../${VERSION}"
+    cp -r ./base/* "../${VERSION}"
+    mkdir -p "../${VERSION}/crds"
+    mkdir -p "../${VERSION}/operator"
+
+    # get a copy of the stack with CRDs
+    helm template replaceme prometheus-community/kube-prometheus-stack --values ./values.yaml -n monitoring --include-crds > "../$VERSION/crds/crds-all.yaml"
```
emosbaugh, line 20: i still think you should add version to these

addons/prometheus/template/generate.sh
```diff
@@ -0,0 +1,71 @@
...
+
+function main() {
+    # run the helm commands
+    helm repo add prometheus-community https://prometheus-community.github.io/helm-charts
+    helm repo update
+
+    get_latest_release_version
+
+    if [ -d "../${VERSION}" ]; then
+        rm -rf "../${VERSION}"
```
emosbaugh, line 61: whats our philosophy here? do we want to regen these? i see addons doing it both ways.

addons/prometheus/template/values.yaml
```diff
@@ -0,0 +1,2422 @@
+# originally from https://github.com/prometheus-community/helm-charts/blob/main/charts/kube-prometheus-stack/values.yaml
```
emosbaugh, line 1: can you use the sha or tag in this link?

addons/prometheus/template/base/operator/prometheus-service.yaml
```diff
@@ -0,0 +1,3 @@
+- op: replace
+  path: /metadata/name
```
emosbaugh, line 2: nice

emosbaugh (approved):

addons/prometheus/template/generate.sh
laverya: I want to regen in case files were removed
but it doesn't make a huge difference

addons/prometheus/template/generate.sh
laverya: The problem is that the version of the chart is something like "14.4.0" while the app version is "0.46.0"

laverya: but done

emosbaugh (approved):

PR #1409, opened 2021-04-02
update prometheus manifest generation code to use unique image names

laverya:
I think having them all be named 'name' resulted in only one image showing up


addons/prometheus/0.46.0/Manifest
```diff
@@ -1,8 +1,8 @@
-image name grafana/grafana:7.4.5
-image name k8s.gcr.io/kube-state-metrics/kube-state-metrics:v1.9.8
-image name quay.io/kiwigrid/k8s-sidecar:1.10.7
-image name quay.io/prometheus-operator/prometheus-operator:v0.46.0
-image name quay.io/prometheus/alertmanager:v0.21.0
-image name quay.io/prometheus/node-exporter:v1.1.2
-image name quay.io/prometheus/prometheus:v2.24.0
-image name directxman12/k8s-prometheus-adapter-amd64:v0.8.3
+image prometheusimage1 grafana/grafana:7.4.5
```
emosbaugh, line 1: any way to use the name here? `quay.io/prometheus/node-exporter:v1.1.2 > node-exporter`

emosbaugh (changes requested):

addons/prometheus/0.46.0/Manifest
laverya: I'll see

emosbaugh (approved):

PR #1417, opened 2021-04-02
Secure openebs installation

MikaelSmith:
`chmod 700 /var/openebs` to ensure that non-root users are unable to access OpenEBS-related mountpoints. This shouldn't impede operation as the container runtime runs as root. Ensures the directory exists before modifying as it's normally created by Kubelet when mounting hostPaths.

On the primary install, iscsi is only setup when using cstor. Do the same for joining new nodes.


emosbaugh:
So far this looks good from my testing. Merging

PR #1424, opened 2021-04-05
Automate rook updates

emosbaugh:


addons/rook/template/base/cluster/patches/filesystem-Json6902.yaml
```diff
@@ -0,0 +1,10 @@
+---
+- op: replace
+  path: /metadata/name
+  value: rook-shared-fs
+- op: replace
```
areed, line 5: I assume you need a JSON 6902 patch to change the name but why are these other values changed here instead of [here](https://github.com/replicatedhq/kURL/pull/1424/files#diff-f173aefeba134a1320dc3e515f5987891bec74af92a0dc0294ea726a8a0148e2R12)?

areed (approved):

addons/rook/template/base/cluster/patches/filesystem-Json6902.yaml
emosbaugh: Because strategicmergepatch cant merge crd arrays without transformer configurations https://github.com/kubernetes-sigs/kustomize/blob/master/examples/transformerconfigs/crd/README.md

PR #1432, opened 2021-04-08
Host preflights for port availability

areed:
Add host preflights for ports 2381, 10248, 10257 and 10259 since cluster
components have healthz endpoints on those ports.

Add host preflight for 6781 abd 6782 since Weave crashes when
unavailable.

Add host preflight for 9100 since node exporter in Prometheus crashes
when unavailable.

No host preflight added for 10249 because kube-proxy continues to
function when unavailable.

### DIFF
--- a/pkg/preflight/assets/host-preflights.yaml
+++ b/pkg/preflight/assets/host-preflights.yaml
@@ -34,9 +34,29 @@ spec:
         port: 2380
         exclude: '{{ and .IsPrimary (not .IsUpgrade) | not }}'
     - tcpPortStatus:
+        collectorName: "ETCD Health Server TCP Port Status"
+        port: 2381
+        exclude: '{{ and .IsPrimary (not .IsUpgrade) | not }}'
+        interface: lo
+    - tcpPortStatus:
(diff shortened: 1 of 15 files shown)


pkg/preflight/assets/host-preflights.yaml
```diff
@@ -184,6 +204,50 @@ spec:
...
+              message: Another process was already listening on port 2381.
+          - fail:
+              when: "connection-timeout"
+              message: Timed out connecting to port 2381. Check your firewall.
+          - fail:
+              when: "error"
+              message: Unexpected port status
+          - pass:
+              when: "connected"
+              message: Port 2381 is available
```
emosbaugh, line 226: what was the decision on this port? do we need to change it or document it?

emosbaugh (approved):

pkg/preflight/assets/host-preflights.yaml
areed: I meant to discuss that at standup. I left it as is because kubeadm configures the etcd pod to use this port as a health check.

PR #1438, opened 2021-04-09
Sonobuoy addon

emosbaugh:


areed (approved):
This makes Sonobuoy assets available through an add-on in kurl, but does not run conformance tests. Automated PRs will be created for new releases.

PR #1443, opened 2021-04-09
fail run on sonobuoy failure

emosbaugh:
Currently sonobuoy cli failures do not fail the run. This pr does:

1. If the cli fails the run will be reported a failure
2. shorten the sonobuoy wait to 10 minutes, if it takes any longer then that its likely never going to succeed
3. skip sonobuoy preflights, right now we pre-create the namespace to add the image pull secrets and this will fail preflight


areed (approved):
Notify testgrid when sonobuoys fails. This does not use the assets from the sonobuoy add-on.

PR #1444, opened 2021-04-09
Fix sonobuoy

emosbaugh:


areed (approved):
Correctly converts sonobuoy spec to bash and installs sonobuoy like other addons

PR #1457, opened 2021-04-13
Include sonobuoy quick mode images in conformance package

emosbaugh:
This will add conformance images for running sonobuoy e2e tests in quick mode. This will not include all images as this adds about 4GB to the conformance package. All kubernetes conformance packages 1.17+ have been regenerated.

### DIFF
--- a/bin/save-manifest-assets.sh
+++ b/bin/save-manifest-assets.sh
@@ -33,6 +33,8 @@ while read -r line; do
     fi
     kind=$(echo $line | awk '{ print $1 }')
 
+    echo "LINE $line"
+
     case "$kind" in
         image)
             filename=$(echo $line | awk '{ print $2 }')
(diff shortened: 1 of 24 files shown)


bin/save-manifest-assets.sh
```diff
@@ -33,6 +33,8 @@ while read -r line; do
     fi
     kind=$(echo $line | awk '{ print $1 }')
 
+    echo "LINE $line"
```
areed, line 36: Debug?

areed (approved):
This PR adds library/nginx:1.14-alpine to conformance packages and a feature to pass in versions to the k8s generate tasks instead of looking up all available versions.

bin/save-manifest-assets.sh
emosbaugh: No its helpful to know where it fails. I'd rather leave it in there.

PR #1465, opened 2021-04-14
Fix kurlkinds build and add to ci

emosbaugh:
Fix kurlkinds tests. Add to CI


areed (approved):
Fix kurlkinds to be compatible with updated controller-runtime version.

PR #1468, opened 2021-04-14
Sonobuoy support for rke2 and k3s

emosbaugh:
1. Build kubernetes conformance packages when building k3s and rke2 packages
2. Add kubernetes conformance packages to k3s and rke2 bundles when sonobuoy is included.
3. Install sonobuoy add-on when included in spec for both k3s and rke2 installers


areed (approved):
This PR makes the sonobuoy add-on work with k3s and rke the same way it already does with kubernetes

PR #1477, opened 2021-04-15
Validate Rook version with Kubernetes

areed:
The API returns a validation error if attempting to create a spec with Rook 1.0.4 and Kubernetes 1.20+. The install script bails if it detects an attempt to install Rook 1.0.4 with Kubernetes 1.20+, which should only happen if you edit the spec manually or if you had previously installed Rook 1.0.4 then upgraded to Rook 1.4 in the spec.

### DIFF
--- a/addons/rook/1.4.3/install.sh
+++ b/addons/rook/1.4.3/install.sh
@@ -5,8 +5,12 @@ function rook_pre_init() {
     if [ -n "$version" ] && [ "$version" != "1.4.3" ]; then
         printf "Rook $version is already installed, will not upgrade to 1.4.3\n"
         export SKIP_ROOK_INSTALL='true'
+        if [ "$version" = "1.0.4" ] && [ "$KUBERNETES_TARGET_VERSION_MINOR" -ge 20 ]; then
+            bail "Rook 1.0.4 is not compatible with Kubernetes 1.20+"
+        fi
     elif [ "$ROOK_BLOCK_STORAGE_ENABLED" != "1" ]; then
         bail "Rook 1.4.3 requires enabling block storage"
+
     fi
 }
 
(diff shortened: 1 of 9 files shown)


web/src/controllers/InstallerAPI.ts
```diff
@@ -101,7 +101,7 @@ export class Installers {
     }
     i.id = i.hash();
 
-    const err = await i.validate();
+    const err = await i.resolve().validate();
```
areed, line 104: Convert `latest` to actual version so the Rook/Kubernetes version validator succeeds.

areed:
Testgrid failures on https://testgrid.kurl.sh/run/pr-1477-d3aaedf-rook-1.0.4-k8s-docker are attempting to downgrade from 1.4.3 to 1.0.4, which is not supported.

addons/rook/1.4.3/install.sh
```diff
@@ -5,6 +5,10 @@ function rook_pre_init() {
     if [ -n "$version" ] && [ "$version" != "1.4.3" ]; then
         printf "Rook $version is already installed, will not upgrade to 1.4.3\n"
         export SKIP_ROOK_INSTALL='true'
+
+        if [ "$version" = "1.0.4" ] && [ "$KUBERNETES_TARGET_VERSION_MINOR" -ge 20 ]; then
+            bail "Rook 1.0.4 is not compatible with Kubernetes 1.20+"
```
emosbaugh, line 10: why bail here? what if the user wants to upgrade other addons?

addons/rook/1.4.9/install.sh
```diff
@@ -14,20 +14,23 @@ function rook_pre_init() {
     next_version_minor="${minor}"
     next_version_patch="${patch}"
 
-    if [ -n "${version}" ]; then
+    if [ -n "${current_version}" ]; then
```
emosbaugh, line 17: thanks

emosbaugh (approved):

addons/rook/1.4.3/install.sh
areed: There's risk that the upgraded version of the add-on requires the upgraded version of Kubernetes. The ticket is ambiguous though:
> when it's in spec: new install - bail and error; existing install - skip update of those components and error

AmberAlston: The bail suggestion was in regards to a fresh install vs an existing installation upgrade. On a fresh install we didn't want to proceed only to leave them in a non-functional or partially installed state.

emosbaugh (approved):

PR #1478, opened 2021-04-15
Only rebuild staging packages weekly

areed:
Prod packages will continue to be updated from staging packages at
release time.


laverya (approved):
This disables automatic rebuilding of packages on production - instead packages will be rebuilt and tested on staging before deployment to production

PR #1479, opened 2021-04-15
prometheus 0.46 fixes

laverya:


addons/prometheus/template/testgrid/k8s-docker.yaml
```diff
@@ -4,7 +4,7 @@
     weave:
       version: "latest"
     rook:
-      version: "1.4.3"
```
emosbaugh, line 7: wanna revert this if it doesnt help?

emosbaugh (approved):

addons/prometheus/template/testgrid/k8s-docker.yaml
laverya: yeah, will do

PR #1481, opened 2021-04-16
Rook 1.5.9

emosbaugh:


emosbaugh:
Test grid is failing. Will fix

emosbaugh:
I added a log function to the common.sh but this isn't available in the testgrid pr. This brings up the bigger question of whether or not an add-on should be fully self-contained but that is a question for another day.

addons/rook/1.4.9/install.sh
```diff
@@ -169,7 +186,12 @@ function rook_ready_spinner() {
     echo "Awaiting rook-ceph operator"
 
     if ! spinner_until 600 rook_version_deployed ; then
-        logWarn "Detected multiple Rook versions"
+        local rook_versions=
+        rook_versions="$(kubectl -n rook-ceph get deployment -l rook_cluster=rook-ceph -o jsonpath='{range .items[*]}{"rook-version="}{.metadata.labels.rook-version}{"\n"}{end}' | sort | uniq)"
+        if [ -n "${rook_versions}" ] && [ "$(echo "${rook_versions}" | wc -l)" -gt "1" ]; then
```
areed, line 191: `"$(echo "${rook_versions}" | wc -l)"`

Is this doing string concatenation then evaluating it?

areed (approved):
Add Rook 1.5.9 with ability to upgrade from 1.4 by entering `y` at prompt. Upgrades work by just installing the new version, there aren't any special migration scripts that run.

addons/rook/1.4.9/install.sh
emosbaugh: This gets all version labels and counts them. Then checks if its greater than 1 to see if there are more than one version running in the cluster. I use that variable in the log line below which is why i dont just store the count.

PR #1486, opened 2021-04-16
Testgrid report last started at

emosbaugh:
This will now show last started at in the UI so its more apparent what is actually running.


laverya (approved):
Show the last time that a run was started on testgrid

PR #1487, opened 2021-04-16
Testgrid - send logs more often

emosbaugh:
This will send the logs more often and allow the user to view them while the test run instance is still running.


laverya (approved):
show logs earlier - not progressive logs, but more incremental

PR #1490, opened 2021-04-17
Fix rook install

emosbaugh:
Rook new installs on 1.4.9 and 1.5.9 are broken as it polls forever for the cluster before deploying the cluster.
Fix race condition on upgrades that causes the cluster to be upgraded before it is ready.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -526,6 +526,7 @@ build/packages/k-3-s/%/rhel-8:
 build/templates: build/templates/install.tmpl build/templates/join.tmpl build/templates/upgrade.tmpl build/templates/tasks.tmpl
 
 build/bin: build/bin/kurl
+	rm -rf kurl_util/bin
 	${MAKE} -C kurl_util build
 	cp -r kurl_util/bin build
 
--- a/addons/rook/1.4.9/install.sh
+++ b/addons/rook/1.4.9/install.sh
@@ -211,25 +211,37 @@ function rook_ceph_healthy() {
 
 # rook_version_deployed check that there is only one rook-version reported across the cluster
 function rook_version_deployed() {
+    # wait for our version to start reporting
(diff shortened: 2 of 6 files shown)


Makefile
```diff
@@ -526,6 +526,7 @@ build/packages/k-3-s/%/rhel-8:
 build/templates: build/templates/install.tmpl build/templates/join.tmpl build/templates/upgrade.tmpl build/templates/tasks.tmpl
 
 build/bin: build/bin/kurl
+	rm -rf kurl_util/bin
```
emosbaugh, line 529: Rebuild kurl_utils on make watchrsync

addons/rook/1.4.9/install.sh
```diff
@@ -153,6 +153,17 @@ function rook_cluster_deploy_upgrade() {
         return 0
     fi
 
+    echo "Awaiting rook-ceph operator"
+
+    if ! spinner_until 600 rook_version_deployed ; then
```
emosbaugh, line 158: This code only needs to run on upgrades. It is currently blocking new rook installations.

addons/rook/1.4.9/install.sh
```diff
@@ -211,25 +211,37 @@ function rook_ceph_healthy() {
 
 # rook_version_deployed check that there is only one rook-version reported across the cluster
 function rook_version_deployed() {
+    # wait for our version to start reporting
+    if ! kubectl -n rook-ceph get deployment -l rook_cluster=rook-ceph -o jsonpath='{range .items[*]}{"rook-version="}{.metadata.labels.rook-version}{"\n"}{end}' | grep -q "${ROOK_VERSION}" ; then
```
emosbaugh, line 215: make sure that the version is reporting before checking if multiple versions are reporting to prevent race condition.

areed (approved):

PR #1496, opened 2021-04-20
list-all-packages.sh includes kubernetes-conformance packages

emosbaugh:
Adds kubernetes-conformance packages to the list of packages to be copied from staging to prod on deploy.

### DIFF
--- a/bin/list-all-packages.sh
+++ b/bin/list-all-packages.sh
@@ -19,6 +19,12 @@ function pkgs() {
             continue
         fi
         echo "${name}-${version}.tar.gz"
+        if [ "${name}" = "kubernetes" ] || [ "${name}" = "k-3-s" ] || [ "${name}" = "rke-2" ]; then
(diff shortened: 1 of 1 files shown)


bin/list-all-packages.sh
```diff
@@ -19,6 +19,12 @@ function pkgs() {
             continue
         fi
         echo "${name}-${version}.tar.gz"
+        if [ "${name}" = "kubernetes" ] || [ "${name}" = "k-3-s" ] || [ "${name}" = "rke-2" ]; then
```
areed, line 22:
```suggestion
        if [ "${name}" = "kubernetes" ]; then
```

areed (changes requested):
This lists conformance packages three times

bin/list-all-packages.sh
emosbaugh: yes, but we dont have 1.18.13 for kubernetes and we do for rke2. I will deduplicate.

areed (approved):

PR #1499, opened 2021-04-20
Cron rebuild packages matrix

emosbaugh:
Currently packages are rebuilt by triggering a separate workflow for each. There is no visibility into builds that failed without navigating to the actions tab and scrolling through all the actions.

This attempts to address this problem by building a matrix and running each package build in parallel as separate jobs within the same workflow.


laverya (approved):
run rebuild packages jobs in parallel, not serially

emosbaugh:
not quite. look at the description

emosbaugh:
> run rebuild packages jobs in parallel, not serially

not quite. look at the description

PR #1511, opened 2021-04-22
optional cephfs

laverya:


emosbaugh (changes requested):
I think you need to make kurlkinds

addons/rook/template/base/install.sh
```diff
@@ -146,14 +146,30 @@ function rook_cluster_deploy() {
     render_yaml_file_2 "$dst/tmpl-rbd-storageclass.yaml" > "$dst/rbd-storageclass.yaml"
     insert_resources "$dst/kustomization.yaml" rbd-storageclass.yaml
 
+    # conditional cephfs
+    if [ "${ROOK_SHARED_FILESYSTEM_DISABLED}" != "1" ]; then
+        insert_resources "$dst/kustomization.yaml" cephfs-storageclass.yaml
+        insert_resources "$dst/kustomization.yaml" filesystem.yaml
+        insert_patches_strategic_merge "$dst/kustomization.yaml" patches/cephfs-storageclass.yaml
+        insert_patches_strategic_merge "$dst/kustomization.yaml" patches/filesystem.yaml
+        cat << EOF >> "$dst/kustomization.yaml"
```
emosbaugh, line 155: this seems brittle? is it possible to use function insert_patches_(Json6902)?

emosbaugh (changes requested):

addons/rook/template/base/install.sh
laverya: we do not yet have such a function, though I can add one

laverya:
to clarify, I will remove the duplicate `insert_patches_json_6902` function from the addon directory before merging - that's just needed for testgrid

emosbaugh (approved):

PR #1512, opened 2021-04-22
Copy packages with no changes from previous release

emosbaugh:
This will compare the diff from the previous release from staging for each individual package and if there are no changes copy the package rather than rebuilding.

This only addresses the issue for staging for now. Will follow up with a pr for releasing to prod.

### DIFF
--- a/bin/upload-dist-prod.sh
+++ b/bin/upload-dist-prod.sh
@@ -2,6 +2,9 @@
 
 set -eo pipefail
 
+# shellcheck source=list-all-packages.sh
+source ./bin/list-all-packages.sh
+
 function require() {
     if [ -z "$2" ]; then
         echo "validation failed: $1 unset"
@@ -29,7 +32,7 @@ function upload_staging() {
 }
 
 # build and upload missing staging packages
-for package in $(bin/list-all-packages.sh)
+for package in $(list_all | awk '{print $1}')
 do
     if ! aws s3api head-object --bucket="${S3_BUCKET}" --key="staging/${GITSHA}/${package}" &>/dev/null; then
         upload_staging "${package}"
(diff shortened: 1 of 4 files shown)


bin/upload-dist-prod.sh
```diff
@@ -29,7 +32,7 @@ function upload_staging() {
 }
 
 # build and upload missing staging packages
-for package in $(bin/list-all-packages.sh)
+for package in $(list_all | awk '{print $1}')
```
laverya, line 35: maybe make this a function in list-all-packages? Since a lot of the callers seem to need it

bin/upload-dist-staging.sh
```diff
@@ -30,18 +54,71 @@ function upload() {
...
+            local md5=
+            md5="$(aws s3api head-object --bucket "${S3_BUCKET}" --key "staging/${package}" | grep '"md5":' | sed 's/[",:]//g' | awk '{print $2}')"
+            aws s3 cp "s3://${S3_BUCKET}/staging/${package}" "s3://${S3_BUCKET}/staging/${GITSHA}/$PKG" \
+                --metadata md5="${md5}",gitsha="${GITSHA}"
+        fi
+    else
+        echo "s3://${S3_BUCKET}/staging/${GITSHA}/${package} already exists"
+        local md5=
+        md5="$(aws s3api head-object --bucket "${S3_BUCKET}" --key "staging/${GITSHA}/${package}" | grep '"md5":' | sed 's/[",:]//g' | awk '{print $2}')"
+        aws s3 cp "s3://${S3_BUCKET}/staging/${GITSHA}/${package}" "s3://${S3_BUCKET}/staging/$PKG" \
```
laverya, line 78: this seems like the only place you use `PKG`

laverya (approved):

PR #1513, opened 2021-04-22
Validate cluster network after CNI is installed

areed:
Run a client and server pod. The client continuously attempts to connect
to the server with the kurlnet service name. A successful connection
means cluster DNS, service networking and pod networking are not
fundamentally broken and it's safe to continue installing other add-ons.

To validate the feature, temporarily block service or pod cidr traffic:
```
sudo iptables -A INPUT -d 10.96.0.0/22 -j DROP
```

Wait for the check network spinner and the client connection error logs. Then unblock traffic and see the script proceed.

```
sudo iptables -L INPUT --line-numbers
sudo iptables -D INPUT 6 # line number will vary
```

### DIFF
--- a/addons/antrea/0.13.1/install.sh
+++ b/addons/antrea/0.13.1/install.sh
@@ -45,6 +45,8 @@ function antrea() {
     kubectl apply -k $dst
 
     antrea_cli
+
+    check_network
 }
 
 function antrea_join() {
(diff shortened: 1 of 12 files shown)


areed:
Testgrid failing because `check_network` is not available yet in the add-ons

laverya (approved):
Wait for a client can connect to a server over a service before continuing after setting up a CNI

PR #1524, opened 2021-04-23
Don't add cephfs resources to cluster kustomization.yaml

areed:
They will be conditionally added at install time if the shared
filesystem is enabled.


laverya (approved):
Sorry for missing this in the optional cephfs pr

PR #1525, opened 2021-04-23
Automated Rook version update 1.5.10

replicatedbot:
Automated changes by the [cron-rook-update](https://github.com/replicatedhq/kURL/blob/master/.github/workflows/update-rook.yaml) GitHub action


areed:
Still has filesytem under resources

PR #1534, opened 2021-04-26
fix autogenerated prometheus hardcoded images

laverya:
First commit is what the automation was producing - hardcoded images for some deployments due to saving the default values file when I setup automation.

Second commit is the fix to the values file and what changes that made to the generated spec.

This does not fix the previous version, 0.47.0-15.2.0, which uses operator v0.46.0 and prometheus v2.24.0 instead of prometheus v2.26.0.


web/src/installers/versions.js
```diff
@@ -60,6 +60,7 @@ module.exports.InstallerVersions = {
   ],
   prometheus: [
     // cron-prometheus-update
+    "0.47.0-15.2.1",
     "0.47.0-15.2.0",
```
emosbaugh, line 64: Do you need to regenerate any versions?

laverya: just regenerated `0.47.0-15.2.0`

emosbaugh (approved):

PR #1539, opened 2021-04-27
add new kurl_util function to get an addon spec as json

laverya:
example output,  from the default 'latestl yaml spec:
```
./kurl_util/bin/yamlutil -j -fp build/default.yaml -jf "spec.contour"
{"version":"latest"}
./kurl_util/bin/yamlutil -j -fp build/default.yaml -jf "spec"
{"contour":{"version":"latest"},"docker":{"version":"latest"},"ekco":{"version":"latest"},"kubernetes":{"version":"latest"},"prometheus":{"version":"latest"},"registry":{"version":"latest"},"rook":{"version":"latest"},"weave":{"version":"latest"}}
```

### DIFF
--- a/kurl_util/cmd/yamlutil/main.go
+++ b/kurl_util/cmd/yamlutil/main.go
@@ -10,6 +10,7 @@ import (
 	"os"
 	"strings"
 
+	"github.com/pkg/errors"
 	"gopkg.in/yaml.v2"
 )
 
@@ -31,7 +32,7 @@ func readFile(path string) []byte {
 	return configuration
 }
 
-func removeField(filePath, yamlField string) {
+func removeField(readFile func(string) []byte, filePath, yamlField string) {
 	var buffer []byte
 
 	configuration := readFile(filePath)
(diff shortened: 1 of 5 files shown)


kurl_util/cmd/yamlutil/main.go
```diff
@@ -99,20 +101,71 @@ func retrieveField(filePath, yamlPath string) {
...
+	if err != nil {
+		log.Fatalf("error: %v", err)
+	}
+
+	fields := strings.Split(jsonPath, ".")
+
+	// get the specified field
+	for _, field := range fields {
+		concrete := parsed.(map[interface{}]interface{})
+		parsed = concrete[field]
```
emosbaugh, line 120: should you do some safety checking here that the field exists with a nice error message?

emosbaugh (changes requested):
Wanna start the good practice of adding tests for these? Seems like an obvious one

kurl_util/cmd/yamlutil/main.go
laverya: sure, makes sense

emosbaugh (approved):
are these tests run in ci

PR #1540, opened 2021-04-27
cache addon specs between runs, and skip applying ones that have not changed

laverya:


areed (approved):
This stores a hash of the last applied config for an addon in a configmap. On the next install if the hash is the same skip re-installing it. Any add-on can define an X_already_applied function that gets called when everything else is skipped.

laverya:
Not the hash - instead the actual addon spec

PR #1549, opened 2021-04-30
Use runc 1.0.0-rc92 in containerd 1.3+

areed:
https://github.com/containerd/containerd/issues/5438


laverya (approved):
forcibly overwrite installed runc version with v1.0.0-rc92 when installing containerd

PR #1552, opened 2021-04-30
Rebuild rook images from UI

areed:
The UI allows you to choose the branch to build from, so we can fix images and get their tag in a single PR.


emosbaugh (approved):

emosbaugh:
Just a note. We only maintain images for 1.0.4

PR #1553, opened 2021-04-30
Fix more CVES in rook 1.0.4

areed:


addons/rook/1.0.4/build-images/ceph/Makefile
```diff
@@ -1,7 +1,7 @@
 SHELL := /bin/bash
 
 IMAGE := kurlsh/ceph:v14.2.0
-WHITELIST := "(GHSA\-|perl\-Pod\-Escapes)"
+WHITELIST := "(GHSA\-|perl\-Pod\-Escapes|libldb,1\.5\.4\-2\.el7)"
```
areed, line 4: Updated image is `1.5.4-2.el7` but scan is looking for `1.5.4-2.el7_9`. Based on [this](https://centos.pkgs.org/7/centos-updates-x86_64/libldb-1.5.4-2.el7.x86_64.rpm.html) I don't think the `el7_9` is required to get the fix.

emosbaugh (approved):

PR #1554, opened 2021-04-30
Begin wait for network check after pod starts

areed:
Otherwise the kurlnet-client pod may still be creating when attempting to get the logs and the install script will fail.

### DIFF
--- a/scripts/common/kubernetes.sh
+++ b/scripts/common/kubernetes.sh
@@ -646,6 +646,14 @@ function kubernetes_pod_completed() {
     return 1
 }
 
+function kubernetes_pod_succeeded() {
+    local name="$1"
+    local namespace="$2"
+
+    local phase=$(kubectl -n $namespace get pod $name -ojsonpath='{ .status.phase }')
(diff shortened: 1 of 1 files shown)


scripts/common/kubernetes.sh
```diff
@@ -733,12 +733,13 @@ metadata:
 spec:
   restartPolicy: OnFailure
   containers:
-  containers:
   - name: pod
     image: $KURL_UTIL_IMAGE
     command: [/usr/local/bin/network, --client, --address, http://kurlnet.default.svc.cluster.local:8080]
 EOF
 
+    spinner_until 120 kubernetes_pod_started kurlnet-client default
```
laverya, line 741: I am personally of the opinion that every spinner needs a log line before it saying what we're waiting for

laverya (approved):
wait (up to 120s) for the network-check pods to start running before beginning to wait (for 60s) for the network-check pods to complete

laverya (approved):

PR #1555, opened 2021-05-01
Automated Antrea version update

replicatedbot:
Automated changes by the [cron-antrea-update](https://github.com/replicatedhq/kURL/blob/master/.github/workflows/update-antrea.yaml) GitHub action


emosbaugh:
This is not right, closing

PR #1562, opened 2021-05-04
handle kebab-case -> camelCase conversion when looking up addons in yaml

laverya:


scripts/common/utilbinaries.sh
```diff
@@ -275,7 +275,12 @@ function is_ha() {
...
 
 function get_addon_config() {
-    addon_name=$1
+    local addon_name=$1
+    addon_name=$(kebab_to_camel "$addon_name")
 
     $BIN_YAMLUTIL -j -fp $MERGED_YAML_SPEC -jf "spec.$addon_name"
 }
+
+function kebab_to_camel() {
```
emosbaugh, line 284: id love to see a unit test for this function

emosbaugh (approved):

scripts/common/utilbinaries.sh
laverya: let me see if we have any shell unit tests, otherwise I'll work on setting some up

emosbaugh (approved):

PR #1569, opened 2021-05-05
Version kurl installer

emosbaugh:


salahalsaleh (approved):

areed (commented):
I don't see how the versioned install script gets versioned packages. I would expect that either the server modifies DIST_URL before rendering or the bash script modifies DIST_URL at runtime.

emosbaugh:
> I don't see how the versioned install script gets versioned packages. I would expect that either the server modifies DIST_URL before rendering or the bash script modifies DIST_URL at runtime.

Good point. That is an important piece I missed. Updating

areed (approved):
This PR allows a version of kurl to be included in URLs for the four main scripts. If included, the kurl server will fetch the script from the versioned directory in the kurl-sh s3 bucket and render that. Versioned scripts have KURL_VERSION set and use that along with DIST_URL to fetch versioned addons and packages.

PR #1584, opened 2021-05-07
Don't set a default stdin read timeout so prompts don't get automatically accepted as true

mjneth:
- Don't set a default stdin read timeout so prompts don't get automatically accepted as true

I'm unsure how to properly test this thought 😬 

https://app.clubhouse.io/replicated/story/33311

https://github.com/replicated-collab/swimlane-replicated/issues/144


emosbaugh:
Hi Matt. Thanks for the contribution!

We have functions `confirmY` and `confirmN` which auto-accept true and false respectively.

Unfortunately a change like this would have to be opt-in via a parameter in the kurl/v1beta1.Installer CRD as well as a flag to the script so as not to break existing customer automation.

emosbaugh:
Hi @mjneth . Thanks again for the contribution. We are attempting to address this issue in #1596 . Please take a look and let me know if it will resolve your issue. Thanks!

mjneth:
> Hi @mjneth . Thanks again for the contribution. We are attempting to address this issue in #1596 . Please take a look and let me know if it will resolve your issue. Thanks!

Seems like that's what we're looking for, no auto-accept of prompts for interactive installs. Thank you!

PR #1599, opened 2021-05-13
Dedicated primary

areed:
Add `taint_primaries` task that adds the default kubeadm taint back to primary nodes, which kurl leaves off because the first node installed must always be a primary that can run workloads.

Add tolerations to rook and ekco by default and in the task to support older installs. Weave and Antrea already have tolerations for the taint.

### DIFF
--- a/scripts/tasks.sh
+++ b/scripts/tasks.sh
@@ -386,16 +386,33 @@ function taint_primaries() {
 
     # Rook tolerations
     if kubectl get namespace rook-ceph &>/dev/null; then
-        # Apply tolerations to cluster config so operator doesn't remove them on next orchestration
         kubectl -n rook-ceph patch cephclusters rook-ceph --type=merge -p '{"spec":{"placement":{"all":{"tolerations":[{"key":"node-role.kubernetes.io/master","operator":"Exists"}]}}}}'
-
-        while read -r deployment; do
-            kubectl -n rook-ceph patch "$deployment" --type=merge -p '{"spec":{"template":{"spec":{"tolerations":[{"key":"node-role.kubernetes.io/master","operator":"Exists"}]}}}}'
-        done < <(kubectl -n rook-ceph get deployments -oname)
-
-        while read -r daemonset; do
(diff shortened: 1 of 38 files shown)


scripts/tasks.sh
```diff
@@ -378,4 +381,65 @@ function set_kubeconfig_server() {
     fi
 }
 
+function taint_primaries() {
+    export KUBECONFIG=/etc/kubernetes/admin.conf
+
+    # Rook tolerations
+    if kubectl get namespace rook-ceph &>/dev/null; then
+        kubectl -n rook-ceph patch cephclusters rook-ceph --type=merge -p '{"spec":{"placement":{"all":{"tolerations":[{"key":"node-role.kubernetes.io/master","operator":"Exists"}]}}}}'
+        cat <<EOF | kubectl -n rook-ceph patch deployment rook-ceph-operator -p "$(cat)"
```
emosbaugh, line 390: why not use the patch file?

scripts/tasks.sh
```diff
@@ -378,4 +381,65 @@ function set_kubeconfig_server() {
     fi
 }
 
+function taint_primaries() {
```
emosbaugh, line 384: does this work with other csi providers? if not should we add some note to docs

scripts/tasks.sh
```diff
@@ -378,4 +381,65 @@ function set_kubeconfig_server() {
...
+    export KUBECONFIG=/etc/kubernetes/admin.conf
+
+    # Rook tolerations
+    if kubectl get namespace rook-ceph &>/dev/null; then
+        kubectl -n rook-ceph patch cephclusters rook-ceph --type=merge -p '{"spec":{"placement":{"all":{"tolerations":[{"key":"node-role.kubernetes.io/master","operator":"Exists"}]}}}}'
```
emosbaugh, line 389: it would be great if there was some way to add an addon hook for this but not gonna make it a blocker.

scripts/tasks.sh
areed: No, OpenEBS and Longhorn do not tolerate the taint

emosbaugh (approved):

Issue #1595, opened 2021-05-11
Grafana Dashboard unable to connect to Prometheus data store


nichelia:
Affected version: `v0.47.0-15.2.0`
Reproducible: Yes (3/3 times)
Available Workaround/Fix: Yes

When deploying with Kots and Metrics & Monitoring Prometheus is selected, the Grafana dashboard is unable to show any data on the dashboards. This is due to the fact that Grafana cannot talk to Prometheus. The usual error appears on the UI as a pop-up message on the top right corner with the message:
`Templating [cluster] Error updating options: Bad Gateway`

The problem is identified as a misconfiguration of the Prometheus service name (url) under the `configmap/prometheus-grafana-datasource` in the `monitoring` namespace.

**Temporary Fix**
In the meantime, to resolve the issue follow the below steps (courtesy of Dmitriy):
1. Open vim editor for the `prometheus-grafana-datasource` configuration under `monitoring` namespace:
   `kubectl edit configmap prometheus-grafana-datasource -n monitoring`
2. Replace `http://k8s:9090/` with `http://prometheus-k8s:9090/`.
3. Restart Grafana service.
    `kubectl scale --replicas=0 deployment/grafana -n monitoring`
    `kubectl scale --replicas=1 deployment/grafana -n monitoring`

**Long term fix**
Reconfigure the Prometheus URL under Grafana's datasources to use the right service name for Prometheus, by default.

PR #1615, opened 2021-05-20
- Update prometheus-url from k8s:9090 to prometheus-k8s:9090

soriaj:
PR to address GH issue [1595](https://github.com/replicatedhq/kURL/issues/1595)


addons/prometheus/0.46.0-14.9.0/operator/default.yaml
```diff
@@ -466,7 +466,7 @@ data:
     datasources:
     - name: Prometheus
       type: prometheus
-      url: http://k8s:9090/
+      http://prometheus-k8s:9090/
```
laverya, line 469: Wouldn't this be `url: http://prometheus-k8s:9090`?

PR #1630, opened 2021-05-24
Upgrade containerd and runc

areed:
1. Upgrade or reinstall containerd to match the spec. Previously containerd was never upgraded or reinstalled.
2. Use the file `containerd-last-applied` to store the hash of the last applied config for that host. The `kurl-current-config` configmap does not work for containerd because Kubernetes cluster may not be up when containerd is installed and because containerd needs to be installed on every node in the cluster.
3. Use runc version 1.0.0-rc95 in containerd and docker. Docker is never upgraded or reinstalled so only new installs will have this version of runc. Containerd installs will get the new version of runc on upgrade or by passing the `force-reapply-addons` flag.
4. Fixed a bug where the install failed when the `force-reapply-addons` flag was passed.

There is no change to the behavior of changing from docker to containerd. Docker will remain installed and containerd install will be skipped.

### DIFF
--- a/scripts/common/addon.sh
+++ b/scripts/common/addon.sh
@@ -29,13 +29,21 @@ function addon_install() {
         export REPORTING_CONTEXT_INFO="addon $name $version"
         # shellcheck disable=SC1090
         . $DIR/addons/$name/$version/install.sh
-        $name
+        # containerd is a special case because there is also a binary named containerd on the host
+        if [ "$name" = "containerd" ]; then
+            containerd_install
+        else
+            $name
+        fi
         export REPORTING_CONTEXT_INFO=""
     fi
 
     addon_set_has_been_applied $name
 
     if commandExists ${name}_join; then
+        ADDONS_HAVE_HOST_COMPONENTS=1
+    fi
+    if [ "$name" = "containerd" ]; then
         ADDONS_HAVE_HOST_COMPONENTS=1
     fi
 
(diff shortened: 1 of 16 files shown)


scripts/common/addon.sh
```diff
@@ -168,7 +177,15 @@ function addon_cleanup() {
 
 function addon_has_been_applied() {
     local name=$1
-    last_applied=$(kubectl get configmap -n kurl kurl-last-config -o jsonpath="{.data.addons-$name}")
+
+    if [ "$name" = "containerd" ]; then
+        if [ -f $DIR/containerd-last-applied ]; then
```
emosbaugh, line 182: is there any reason why you chose to do this rather than say add to the config map at some point after k8s is installed, say in the outro?

emosbaugh (approved):

scripts/common/addon.sh
emosbaugh: when you say `Kubernetes cluster may not be up`, do you mean that on initial install or may it not be up on upgrade as well?

areed: I meant that it won't be up on initial install.

Using the config map doesn't work for containerd because remote nodes may be workers and not have kubectl. And we have to keep track of the version of containerd installed on that node. It's not like other addons where we only care about the last version applied to the cluster.

PR #1631, opened 2021-05-24
feat: add proxy configuration to velero >1.5.1

DanStough:


addons/velero/1.5.1/tmpl-restic-daemonset-proxy.yaml
```diff
@@ -0,0 +1,14 @@
...
+spec:
+  template:
+    spec:
+      containers:
+        - name: restic
+          env:
+          - name: HTTP_PROXY
+            value: ${PROXY_ADDRESS}
+          - name: NO_PROXY
+            value: ${NO_PROXY_ADDRESSES},${OBJECT_STORE_CLUSTER_IP},${KUBERNETES_CLUSTER_IP}
```
areed, line 14: `NO_PROXY_ADDRESSES` includes the service cidr and both restic and velero are written in Go so it should be redundant to add the two cluster IPs.

areed (approved):

PR #1632, opened 2021-05-24
Go dependency updates

emosbaugh:


pkg/preflight/builtin_test.go
```diff
@@ -10,7 +10,7 @@ import (
 	"github.com/replicatedhq/kurl/pkg/installer"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
-	yaml "gopkg.in/yaml.v3"
+	yaml "gopkg.in/yaml.v2"
```
laverya, line 13: why the downgrade?

laverya: ooh because it was the only v3 usage, ok

laverya (approved):
Looks good, but I want to see this in testgrid before we release just in case the yaml library version change breaks things

emosbaugh:
Damn, builds are failing. I ran it locally and it worked. Ill have to see what is the difference.

pkg/preflight/builtin_test.go
emosbaugh: had to add it back anyway. unmarshal failed

kurl_util/cmd/yamltobash/main.go
```diff
@@ -16,7 +16,7 @@ import (
 	kurlscheme "github.com/replicatedhq/kurl/kurlkinds/client/kurlclientset/scheme"
 	kurlv1beta1 "github.com/replicatedhq/kurl/kurlkinds/pkg/apis/cluster/v1beta1"
 	kurlversion "github.com/replicatedhq/kurl/pkg/version"
-	"gopkg.in/yaml.v1"
```
emosbaugh, line 19: @areed is there any reason this is using v1?

areed: Not that I'm aware of

laverya (approved):

PR #1651, opened 2021-05-27
Fix versioned kurl-util image and kurl-bin-utils file

emosbaugh:
running the following you get a partial bundle because it tries to include the latest kurl-bin-utils package which is not part of the older versioned bundle.

```
curl -LO https://kurl.sh/bundle/version/v2021.05.21-1/latest.tar.gz
```


emosbaugh:
[ch34471]

areed (approved):

PR #1653, opened 2021-05-27
Update to latest patch release of Ceph for CVE fixes

MikaelSmith:


MikaelSmith:
I've tested this out locally using the ceph image directly and rebuilding the rook-ceph image. I updated the ceph image in the CephCluster for an existing 1.0.4 install, and updated the rook-ceph-operator deployment with the new rook-ceph image.

Rebuilding rook-ceph was necessary to update the client so it could talk to the server after also running `ceph config set mon auth_allow_insecure_global_id_reclaim false`.

I'd need help with the rest of the steps in https://github.com/replicatedhq/kURL/blob/master/addons/rook/1.0.4/build-images/README.md.

MikaelSmith:
It's not yet clear to me whether new installs need to set `mon_warn_on_insecure_global_id_reclaim_allowed = false`.

addons/rook/1.0.4/build-images/ceph/Dockerfile
```diff
@@ -1,6 +1,6 @@
 # https://github.com/ceph/ceph-container
 
-FROM ceph/ceph:v14.2.0
+FROM ceph/ceph:v14.2.21
```
MikaelSmith, line 3: It probably makes sense to add this as a new addon version like `1.0.4-14.2.21`. Image tag would also need to change.

emosbaugh:
Resolved by  #1726 and #1742

PR #1657, opened 2021-05-28
add a preflight check for longhorn disk usage

laverya:


pkg/preflight/assets/host-preflights.yaml
```diff
@@ -360,3 +363,19 @@ spec:
...
+            when: "total < 50Gi"
+            message: The disk containing directory /var/lib/longhorn has less than 50Gi of total space
+        - fail:
+            when: "used/total > 80%"
+            message: The disk containing directory /var/lib/longhorn is more than 80% full
+        - warn:
+            when: "used/total > 60%"
+            message: The disk containing directory /var/lib/longhorn is more than 60% full
+        - pass:
+            message: The disk containing directory /var/lib/longhorn has at least 30Gi disk space available
```
areed, line 381:
```suggestion
            message: The disk containing directory /var/lib/longhorn has at least 20Gi disk space available
```

areed (approved):
There's a [mechanism](https://github.com/replicatedhq/kURL/blob/master/addons/weave/2.8.1/install.sh#L88) to add these to the add-on. The benefit of that is we could add more preflights specific to newer versions of longhorn as they're released. Not a blocker though.

pkg/preflight/assets/host-preflights.yaml
laverya: "The disk containing directory /var/lib/longhorn has at least 20Gi disk space available and is at least 50Gi" would be even better, will do

addons/longhorn/1.1.1/host-preflight.yaml
```diff
@@ -0,0 +1,27 @@
...
+  collectors:
+  - diskUsage:
+      collectorName: "Longhorn Disk Usage"
+      path: /var/lib/longhorn
+
+  analyzers:
+  - diskUsage:
+      checkName: "Longhorn Disk Usage"
+      collectorName: "Longhorn Disk Usage"
+      exclude: '{{kurl and .Installer.Spec.Longhorn.Version (not .IsUpgrade) | not }}' # only run if there is a longhorn version set and this is not an upgrade
```
areed: I think the check for Longhorn version is redundant in here

areed (approved):

addons/longhorn/1.1.1/host-preflight.yaml
laverya: fixed it before the review, funnily enough
Race conditions!

PR #1661, opened 2021-05-28
testgrid versioned installer

emosbaugh:


testgrid/tgrun/pkg/scheduler/scheduler.go
```diff
@@ -264,12 +284,30 @@ func stringInSlice(s string, slice []string) bool {
...
 	if len(matches) != 3 {
 		return "", fmt.Errorf("unable to parse url %q", url)
 	}
 
-	return fmt.Sprintf("%s/bundle/%s.tar.gz", matches[1], matches[2]), nil
+	if kurlVersion == "" {
+		return url, nil
+	}
+
+	return fmt.Sprintf("%s/version/%s/%s", matches[1], kurlVersion, matches[2]), nil
```
laverya, line 298: won't this result in including `version//` in URLs that didn't supply a kurlVersion?

Is that what we want to do?

emosbaugh: No cause line 294

laverya (approved):

PR #1675, opened 2021-06-01
Ekco 0.10.2

areed:
The ekco shutdown script now deletes all pods on the node with PVCs
mounted, not just pods with PVCs provisioned by Rook.


addons/ekco/0.10.2/reboot/shutdown.sh
```diff
@@ -0,0 +1,45 @@
+#!/bin/bash
```
areed, line 1: This file changed lines 7-14 and 22-26. There were no other changes since 0.10.1. If this PR is approved I'll push a new 0.10.2 tag for ekco identical to 0.10.1 before merging.

emosbaugh (approved):

PR #1676, opened 2021-06-01
Always regenerate containerd config

areed:
Previously we did not regenerate containerd's config file so that users
could make modifications and not have them wiped out the next time kurl
runs. That leads to deprecated options being used with containerd 1.4+.
Examples include runtime_type = "io.containerd.runc.v1" and
sandbox_image = "k8s.gcr.io/pause:3.1".

Add a new field `containerd.preserveConfig` to always preserve the existing config file and a new field `containerd.tomlConfig` to specify config that will be merged into the containerd config.

### DIFF
--- a/kurl_util/Makefile
+++ b/kurl_util/Makefile
@@ -53,7 +53,7 @@ test: lint vet
 	go test ./cmd/...
 
 .PHONY: build
-build: bin/yamlutil bin/subnet bin/docker-config bin/config bin/installermerge bin/yamltobash bin/bashmerge bin/bcrypt bin/htpasswd bin/network
+build: bin/yamlutil bin/subnet bin/docker-config bin/config bin/installermerge bin/yamltobash bin/bashmerge bin/bcrypt bin/htpasswd bin/network bin/toml
 
 bin/yamlutil: cmd/yamlutil/main.go
 	go build ${LDFLAGS} -o bin/yamlutil cmd/yamlutil/main.go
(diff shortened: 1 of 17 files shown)


laverya (approved):
Should we add a flag to replace this behavior? Or a way for a patch file to include a containerd spec like we do with Docker?

kurl_util/Makefile
```diff
@@ -53,7 +53,7 @@ test: lint vet
 	go test ./cmd/...
 
 .PHONY: build
-build: bin/yamlutil bin/subnet bin/docker-config bin/config bin/installermerge bin/yamltobash bin/bashmerge bin/bcrypt bin/htpasswd bin/network
+build: bin/yamlutil bin/subnet bin/docker-config bin/config bin/installermerge bin/yamltobash bin/bashmerge bin/bcrypt bin/htpasswd bin/network bin/toml
```
laverya, line 56: at some point we should really look at making these multiple entrypoints to one binary (`kurl-util toml` vs `toml`) to reduce size

now's not the time though

kurl_util/cmd/toml/main_test.go
```diff
@@ -0,0 +1,79 @@
...
+	for _, leaf := range expect {
+		assert.Contains(t, actual, leaf)
+	}
+}
+
+func TestMerge(t *testing.T) {
+	base := `
+version = 2
+
+[debug]
```
laverya, line 39: What happens if `debug` doesn't exist? Can we test that (patching in a new value instead of changing a value) too?

kurl_util/cmd/toml/main_test.go
```diff
@@ -0,0 +1,79 @@
...
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	patchTree, err := toml.Load(patch)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	merge(baseTree, patchTree)
```
laverya, line 74: Should we compare to the return value here (the actual string we'll be writing) instead of to the modified baseTree?

laverya (approved):

PR #1678, opened 2021-06-02
check mount propagation is bidirectional in Longhorn

laverya:
this duplicates the check within https://raw.githubusercontent.com/longhorn/longhorn/v1.1.1/scripts/environment_check.sh (but without the dependency on jq, and using the kurl util image instead of busybox)

### DIFF
--- a/addons/longhorn/1.1.1/install.sh
+++ b/addons/longhorn/1.1.1/install.sh
@@ -127,6 +127,7 @@ function check_mount_propagation() {
     local src=$1
 
     kubectl get ns longhorn-system >/dev/null 2>&1 || kubectl create ns longhorn-system >/dev/null 2>&1
+    kubectl delete -n longhorn-system ds longhorn-environment-check || true
 
     render_yaml_file "$src/tmpl-mount-propagation.yaml" > "$src/mount-propagation.yaml"
     kubectl apply -f "$src/mount-propagation.yaml"
@@ -144,11 +145,15 @@ function validate_longhorn_ds() {
     local allpods=$(kubectl get daemonsets -n longhorn-system longhorn-environment-check --no-headers | tr -s ' ' | cut -d ' ' -f4)
     local bidirectional=$(kubectl get pods -l app=longhorn-environment-check -o=jsonpath='{.items[*].spec.containers[0].volumeMounts[*]}' | grep -o 'Bidirectional' | wc -l)
 
-    if [ "$bidirectional" -lt "$allpods" ]; then
-        logWarn "Only $bidirectional of $allpods nodes support Longhorn storage"
-    fi
+    if [ "$allpods" == "" ]; then
+        logWarn "unable to determine health and status of longhorn-environment-check daemonset"
+    else
+        if [ "$bidirectional" -lt "$allpods" ]; then
+            logWarn "Only $bidirectional of $allpods nodes support Longhorn storage"
+        fi
 
-    if [ "$bidirectional" -eq "0" ]; then
-        logWarn "no Longhorn mount propagation pods detected"
+        if [ "$bidirectional" -eq "0" ]; then
+            logWarn "No Longhorn mount propagation pods detected"
+        fi
     fi
 }
--- a/addons/longhorn/1.1.1/install.sh
+++ b/addons/longhorn/1.1.1/install.sh
@@ -126,9 +126,11 @@ function longhorn_preflight() {
 function check_mount_propagation() {
     local src=$1
 
+    kubectl get ns longhorn-system 2>/dev/null 1>/dev/null || kubectl create ns longhorn-system 1>/dev/null
+
     render_yaml_file "$src/tmpl-mount-propagation.yaml" > "$src/mount-propagation.yaml"
     kubectl create -f "$src/mount-propagation.yaml"
-    echo "Waiting for Longhorn Mount Propagation Check Daemonset to be ready"
+    echo "Waiting for the Longhorn Mount Propagation Check Daemonset to be ready"
     spinner_until 120 longhorn_daemonset_is_ready longhorn-manager
 
     validate_longhorn_ds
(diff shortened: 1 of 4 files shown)


addons/longhorn/template/base/install.sh
```diff
@@ -119,3 +122,30 @@ function longhorn_preflight() {
...
+    echo "Waiting for the Longhorn Mount Propagation Check Daemonset to be ready"
+    spinner_until 120 longhorn_daemonset_is_ready longhorn-manager
+
+    validate_longhorn_ds
+
+    kubectl delete -f "$src/mount-propagation.yaml"
+}
+
+
+validate_longhorn_ds() {
```
emosbaugh, line 142:
```
function validate_longhorn_ds
```

addons/longhorn/template/base/install.sh
```diff
@@ -119,3 +122,30 @@ function longhorn_preflight() {
     local src="${DIR}/addons/longhorn/${LONGHORN_VERSION}"
     echo "${src}/host-preflight.yaml"
 }
+
+function check_mount_propagation() {
+    local src=$1
+
+    kubectl get ns longhorn-system 2>/dev/null 1>/dev/null || kubectl create ns longhorn-system 1>/dev/null
```
emosbaugh, line 129: our standard way to redirect stdout and stderr to dev null is `>/dev/null 2>&1`

addons/longhorn/template/base/install.sh
```diff
@@ -119,3 +122,30 @@ function longhorn_preflight() {
...
+
+    render_yaml_file "$src/tmpl-mount-propagation.yaml" > "$src/mount-propagation.yaml"
+    kubectl create -f "$src/mount-propagation.yaml"
```
emosbaugh, line 132: what if delete fails and you re-run? should we delete before we create here?

addons/longhorn/template/base/install.sh
```diff
@@ -119,3 +122,30 @@ function longhorn_preflight() {
...
+    validate_longhorn_ds
+
+    kubectl delete -f "$src/mount-propagation.yaml"
+}
+
+
+validate_longhorn_ds() {
+    local allSupported=true
+
+    local allpods=$(kubectl get daemonsets -n longhorn-system longhorn-environment-check --no-headers | tr -s ' ' | cut -d ' ' -f4)
```
emosbaugh, line 145: if this fails allpods will be empty. can you account for that in your conditional on line 148?

emosbaugh (changes requested):

addons/longhorn/template/base/install.sh
laverya: or maybe just have it be `apply -f`

emosbaugh:
Sure but do you need to re-run it in case changes were made?

addons/longhorn/template/base/install.sh
```diff
@@ -119,3 +122,33 @@ function longhorn_preflight() {
...
+    kubectl delete -f "$src/mount-propagation.yaml"
+}
+
+function validate_longhorn_ds() {
+    local allSupported=true
+
+    local allpods=$(kubectl get daemonsets -n longhorn-system longhorn-environment-check --no-headers | tr -s ' ' | cut -d ' ' -f4)
+    local bidirectional=$(kubectl get pods -l app=longhorn-environment-check -o=jsonpath='{.items[*].spec.containers[0].volumeMounts[*]}' | grep -o 'Bidirectional' | wc -l)
+
+    if [ "$bidirectional" -lt "$allpods" ]; then
```
emosbaugh, line 147: this will fail with an error if allpods is empty

emosbaugh (changes requested):

addons/longhorn/template/base/install.sh
```diff
@@ -119,3 +122,33 @@ function longhorn_preflight() {
...
     echo "${src}/host-preflight.yaml"
 }
+
+function check_mount_propagation() {
+    local src=$1
+
+    kubectl get ns longhorn-system >/dev/null 2>&1 || kubectl create ns longhorn-system >/dev/null 2>&1
+
+    render_yaml_file "$src/tmpl-mount-propagation.yaml" > "$src/mount-propagation.yaml"
+    kubectl apply -f "$src/mount-propagation.yaml"
```
emosbaugh, line 132: what if this daemonset existed prior to whatever changes were applied?

laverya: I'll add a delete command first then

addons/longhorn/template/base/install.sh
laverya: That would show up in longhorn_daemonset_is_ready first, though

addons/longhorn/1.1.1/install.sh
```diff
@@ -119,3 +122,38 @@ function longhorn_preflight() {
...
+    local allSupported=true
+
+    local allpods=$(kubectl get daemonsets -n longhorn-system longhorn-environment-check --no-headers | tr -s ' ' | cut -d ' ' -f4)
+    local bidirectional=$(kubectl get pods -l app=longhorn-environment-check -o=jsonpath='{.items[*].spec.containers[0].volumeMounts[*]}' | grep -o 'Bidirectional' | wc -l)
+
+    if [ "$allpods" == "" ]; then
+        logWarn "unable to determine health and status of longhorn-environment-check daemonset"
+    else
+        if [ "$bidirectional" -lt "$allpods" ]; then
+            logWarn "Only $bidirectional of $allpods nodes support Longhorn storage"
```
emosbaugh, line 152: I'm unclear the point of this check. In the linked preflight from the upstream this would block the installation.

laverya: I'll change it to exit with an error if no nodes work

laverya: I've added a 'bail if none are detected' condition

But our other preflights continue if there's any chance of a successful install

emosbaugh: i still just dont think i understand the goal here. are we trying to match the logic in the longhorn preflight? if so this doesnt seem to match

```
  if [ "$allSupported" != "true" ]; then
   ...
   exit 1
```

emosbaugh (approved):

emosbaugh (approved):

PR #1704, opened 2021-06-10
Github workflow max jobs is 256

emosbaugh:
Actions is getting error cron-rebuild-packages-staging : .github#L1 Strategy produced more than 256

This action no longer runs on a schedule. Its now on workflow dispatch.

This PR runs max matrix jobs 255 and additionally adds an input variable index to run the matrix starting at a given index.

This is quick and dirty to get it working again. In the future it may be best to batch these.

### DIFF
--- a/bin/list-all-packages-actions-matrix.sh
+++ b/bin/list-all-packages-actions-matrix.sh
@@ -15,10 +15,22 @@ function require() {
 require KURL_UTIL_IMAGE "${KURL_UTIL_IMAGE}" # required for common package
 require KURL_BIN_UTILS_FILE "${KURL_BIN_UTILS_FILE}"
 
+index="${1:-0}"
+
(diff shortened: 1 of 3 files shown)


bin/list-all-packages-actions-matrix.sh
```diff
@@ -15,10 +15,22 @@ function require() {
...
 for package in $(list_all | awk '{print $1}')
 do
+    if [ "${i}" -lt "${index}" ]; then
+        i=$((i+1))
+        continue
+    fi
+    if [ "$(("${i}"-"${index}"))" = "255" ]; then
+        break
+    fi
+    i=$((i+1))
```
laverya, line 33: will this skip the first package? (and if it should run before the printf call, why not combine with line 27?)

PR #1709, opened 2021-06-11
Use docker rhel 8 packages

emosbaugh:
This will now use the official repos for docker on rhel 8 based distros.

Docker 18.09.8, 19.03.4 and 19.03.10 will no longer work on rhel 8 based distros.

I will include a follow up pr to add back the brute force method for installing old docker on rhel 8.

### DIFF
--- a/bin/list-all-packages.sh
+++ b/bin/list-all-packages.sh
@@ -55,6 +55,7 @@ function list_other() {
     echo "docker-18.09.8.tar.gz bundles/"
     echo "docker-19.03.4.tar.gz bundles/"
     echo "docker-19.03.10.tar.gz bundles/"
+    echo "docker-19.03.15.tar.gz bundles/"
     echo "docker-20.10.5.tar.gz bundles/"
 }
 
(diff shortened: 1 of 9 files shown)


bin/list-all-packages.sh
```diff
@@ -55,6 +55,7 @@ function list_other() {
     echo "docker-18.09.8.tar.gz bundles/"
     echo "docker-19.03.4.tar.gz bundles/"
     echo "docker-19.03.10.tar.gz bundles/"
+    echo "docker-19.03.15.tar.gz bundles/"
```
laverya, line 58: new docker version?

laverya (approved):

bin/list-all-packages.sh
emosbaugh: since docker 19.03.10 is unsupported on rhel 8 but the latest validated version for most kubernetes version is 19.03 i included a supported version

https://github.com/kubernetes/kubernetes/blob/v1.20.7/build/dependencies.yaml#L68-L70

PR #1716, opened 2021-06-11
Collect host preflight results in /var/lib/kurl/host-preflights directory

emosbaugh:
This pull request will only output a file with kurl version, preflight flags and results for tracking purposes. It will not yet collect in a support bundle.


scripts/common/preflights.sh
```diff
@@ -391,12 +391,19 @@ function preflights_require_no_kubernetes_or_current_node() {
...
+HOST_PREFLIGHTS_RESULTS_OUTPUT_DIR="host-preflights"
 function host_preflights() {
     local is_primary="$1"
     local is_join="$2"
     local is_upgrade="$3"
 
     local opts=
+
+    local out_file=
+    out_file="${HOST_PREFLIGHTS_RESULTS_OUTPUT_DIR}/results-$(date +%s).txt"
```
laverya, line 403: won't this be relative to the path that this is being run in? Should we put this with the rest of the kurl cache?

scripts/common/preflights.sh
```diff
@@ -423,18 +430,26 @@ function host_preflights() {
...
+        cat "${out_file}"
+        host_preflights_mkresults "${out_file}" "${opts}"
+
         # TODO: report preflight fail
     else
         # interactive terminal
         if prompts_can_prompt; then
             set +e
-            "${DIR}"/bin/kurl host preflight "${MERGED_YAML_SPEC}" ${opts} </dev/tty
+            "${DIR}"/bin/kurl host preflight "${MERGED_YAML_SPEC}" ${opts} > "${out_file}" </dev/tty
```
laverya, line 443: should we use tee here instead?

emosbaugh: that would change the exit code. i can only tee on line 433 but then it would be inconsistent

scripts/common/preflights.sh
emosbaugh: fixed

scripts/common/preflights.sh
laverya: I'm fine with this since (discussed elsewhere) progress is on stderr and thus won't be interrupted, only results are on stdout

laverya (approved):

laverya (approved):

PR #1729, opened 2021-06-17
Kwsorensen/rook ceph secure global

kwsorensen:
Pull request to fix [HEALTH_WARN mon is allowing insecure global_id reclaim
](https://docs.ceph.com/en/latest/security/CVE-2021-20288/)


MikaelSmith:
https://github.com/rook/rook/releases/tag/v1.6.5 has their own work to address this. But maybe the kURL environment is simple enough that you don't need to worry about outdated clients during the upgrade process.

laverya:
I would recommend (instead of this) 
1. make this change: addons/rook/1.5.11/operator/configmap-rook-config-override.yaml
2. make the same change in templates (addons/rook/templates/base/operator/configmap-rook-config-override.yaml)
3. let automation make 1.5.12

kwsorensen:
Ways Tested:
Fresh install with fixed version 5.11 - good
Fresh install with fixed version 5.12 - good
Fresh install with un-fixed version 5.11 into a fixed version 5.12 - Ceph status still shows errors.

areed (commented):
https://github.com/replicatedhq/kURL/pull/1729/files#diff-5041c635371838ae5901774a8bcf202a1543309d1462226eb36278b4a07da8f9R182

Calling `rook_patch_insecure_clients` here would fix re-install of 1.5.11 with the `force-reapply-addons` flag.

areed (approved):
This disables insecure global_id reclaim after all clients have been patched to authenticate correctly.

addons/rook/1.5.11/install.sh
```diff
@@ -370,3 +373,20 @@ function rook_lvm2() {
 
     install_host_archives "$src"
 }
+
+function rook_clients_secure {
+    if [[ $(kubectl -n rook-ceph exec deploy/rook-ceph-tools -- ceph status | grep "mon is allowing insecure global_id reclaim") ]];
+       then return 1
```
emosbaugh, line 379: you dont need `fi`? for readability can you change to this format?

```
if ... ; then
  return 1
fi

emosbaugh (changes requested):

emosbaugh (approved):

emosbaugh (approved):

PR #1732, opened 2021-06-18
kotsadm as a statefulset - migrate object store data

sgalsaleh:


DanStough (changes requested):
I think you can also go ahead and remove minio from the manifest: https://github.com/replicatedhq/kURL/blob/b4d984288b5be7fa1247ea9b59be6d1674b22b71/addons/kotsadm/alpha/Manifest#L3

sgalsaleh:
@DanStough that image is used by the minio instance we deploy to configure NFS/Hostpath storage destination for snapshots, so we can't remove it just yet.

DanStough (approved):

divolgin (approved):

PR #1743, opened 2021-06-23
retain a maximum of 9GB of prometheus metrics data

laverya:


laverya:
this uses an experimental prometheus flag, but that's still better than nothing: https://prometheus.io/docs/prometheus/latest/storage/#operational-aspects + https://github.com/prometheus-operator/prometheus-operator/blob/1c35faa561f2557fa156438c13c5763d3b6ac6c4/pkg/prometheus/statefulset.go#L346

emosbaugh (approved):

PR #1744, opened 2021-06-23
fix(registry): fix kurl-registry-ip flag for containerd

DanStough:


addons/registry/2.7.1/install.sh
```diff
@@ -1,8 +1,6 @@
-
+# shellcheck disable=SC2148
```
laverya, line 1: what does this disable?

DanStough: Not having the shebang at the beginning of the file.

laverya (approved):

PR #1745, opened 2021-06-23
Add Bloomberg's goldpinger

laverya:
https://github.com/bloomberg/goldpinger

### DIFF
--- a/addons/goldpinger/template/values.yaml
+++ b/addons/goldpinger/template/values.yaml
@@ -24,7 +24,13 @@ serviceAccount:
 goldpinger:
   port: 80
 
-extraEnv: []
+extraEnv:
+  # added block from https://github.com/bloomberg/goldpinger/blob/4f8d872/extras/example-serviceaccounts.yml#L44-L48 (replicated)
+  # podIP is used to select a randomized subset of nodes to ping.
+  - name: POD_IP
+    valueFrom:
+      fieldRef:
+        fieldPath: status.podIP
 
 service:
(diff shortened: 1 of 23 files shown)


addons/goldpinger/template/values.yaml
```diff
@@ -0,0 +1,102 @@
...
+affinity: {}
+
+## Enable this if pod security policy enabled in your cluster
+## It will bind ServiceAccount with unrestricted podSecurityPolicy
+## Ref: https://kubernetes.io/docs/concepts/policy/pod-security-policy/
+podSecurityPolicy:
+  enabled: false
+  policyName: unrestricted-psp
+
+serviceMonitor:
```
DanStough, line 96: Might agree with @emosbaugh to turn this on by default, but wouldn't let it hold up the PR.

DanStough (changes requested):
I think you may need to add the addon to the distro file(s), like this one for kubeadm https://github.com/replicatedhq/kURL/blob/bb9d9c3d72e9052d85b8ae8313cb6d10266a106d/scripts/distro/kubeadm/distro.sh#L35-L58

addons/goldpinger/template/values.yaml
emosbaugh: I would turn this on IF the prom addon is enabled

laverya: Will do

laverya: enabled (but not through the helm chart)

DanStough (approved):

PR #1749, opened 2021-06-23
Speed up build

emosbaugh:


areed (approved):
Makes every worker build 5 packages instead of a single package.

PR #1768, opened 2021-06-24
Update PVC storage class when restoring to longhorn

areed:


addons/velero/1.5.1/tmpl-restic-daemonset-proxy.yaml
```diff
@@ -2,6 +2,7 @@ apiVersion: apps/v1
 kind: DaemonSet
 metadata:
   name: restic
+  namespace: ${VELERO_NAMESPACE}
```
areed, line 5: Patch namesapce must match base in K8s 1.21

laverya (approved):

PR #1783, opened 2021-06-25
Automated Prometheus version update 0.48.1-16.12.1

replicatedbot:
Automated changes by the [cron-prometheus-update](https://github.com/replicatedhq/kURL/blob/master/.github/workflows/update-prometheus.yaml) GitHub action


laverya:
You know what would be great with this? A diff between the previous addon version and the new one
(but with helm chart labels filtered out)

emosbaugh:
Good point. This could be a no-op

PR #1787, opened 2021-06-25
Include kotsadm-tls cert in snapshot

areed:


laverya (approved):
I think you should include this in the current kotsadm release too

PR #1795, opened 2021-06-28
prometheus service types

laverya:


areed (commented):
This appears to only support `ClusterIP` and `NodePort`.  Could we add validation so the server returns 400 for any other values?

web/src/installers/index.ts
```diff
@@ -1199,6 +1199,17 @@ export class Installer {
         return {error: {message: "Rook 1.0.4 is not compatible with Kubernetes 1.20+"}};
       }
     }
+
+    if (this.spec.prometheus && this.spec.prometheus.version && this.spec.prometheus.serviceType) {
+      if (this.spec.prometheus.serviceType != "" && this.spec.prometheus.serviceType != "ClusterIP" && this.spec.prometheus.serviceType != "NodePort") {
+        return {error: {message: `Supported Prometheus service types are "NodePort" and "ClusterIP", not "${this.spec.prometheus.serviceType}"`}};
```
areed, line 1205: I think `this.spec.prometheus.serviceType` needs to be escaped unless that's handled at a higher level

areed: Actually I think this is fine

areed (approved):

web/src/installers/index.ts
laverya: I think that's handled, but I can try it in a test

laverya: tried, worked fine

addons/prometheus/0.48.1-16.10.0/install.sh
```diff
@@ -88,7 +94,11 @@ function grafana_admin_secret() {
 function prometheus_outro() {
     printf "\n"
     printf "\n"
-    printf "The UIs of Prometheus, Grafana and Alertmanager have been exposed on NodePorts ${GREEN}30900${NC}, ${GREEN}30902${NC} and ${GREEN}30903${NC} respectively.\n"
+    if [ -z "$PROMETHEUS_SERVICE_TYPE" ] || [ "$PROMETHEUS_SERVICE_TYPE" = "NodePort" ] ; then
+        printf "The UIs of Prometheus, Grafana and Alertmanager have been exposed on NodePorts ${GREEN}30900${NC}, ${GREEN}30902${NC} and ${GREEN}30903${NC} respectively.\n"
+    else
+        printf "The UIs of Prometheus, Grafana and Alertmanager have been exposed on internal ClusterIP services.\n"
```
MikaelSmith, line 100: Is there a reason Grafana is included in this? It seems like the obvious frontend to these services, and requires authentication. Although it would also make sense to configure TLS on it.

PR #1796, opened 2021-06-28
Run testgrid after deploy. Require instances and os specs from the cli

emosbaugh:


kurlkinds/pkg/apis/cluster/v1beta1/installer_types.go
```diff
@@ -106,7 +106,7 @@ type Kubernetes struct {
 }
 
 type Kurl struct {
-	AdditionalNoProxyAddresses   []string `json:"additionalNoProxyAddresses,omitempty" yaml:"proxyAddress,omitempty"`
```
laverya, line 109: was this intended?

testgrid/specs/latest.yaml
```diff
@@ -0,0 +1,20 @@
+- name: ""
```
laverya, line 1: can we also include the spec for longhorn? https://kurl.sh/longhorn

kurlkinds/pkg/apis/cluster/v1beta1/installer_types.go
emosbaugh: Isn't that a bug. It's the wrong yaml tag. Or am I misunderstanding your comment?

testgrid/specs/latest.yaml
emosbaugh: I'll add

testgrid/tgrun/pkg/scheduler/addons.go
```diff
@@ -34,6 +34,9 @@ func generateAllAddOns() ([]kurlv1beta1.InstallerSpec, error) {
 	return installerSpecs, nil
 }
 
+// hack
+var previouslyGeneratedNames = []string{}
```
laverya, line 38: what's this used for? (I realize it's not new)

laverya (approved):

testgrid/tgrun/pkg/scheduler/addons.go
emosbaugh: i have no idea. i moved it closer to where its referenced

kurlkinds/pkg/apis/cluster/v1beta1/installer_types.go
laverya: I misread the +- 🤦

laverya (approved):

PR #1797, opened 2021-06-28
Velero kotsadm restore item action

areed:
This add a [velero plugin](https://velero.io/docs/v1.6/overview-plugins/) to modify the kotsadm deployment at restore time. The plugin has been added to the kurl_util image, which is required to copy the binary into a specific directory in its default command in order to register the plugin when used as an initContainer in the velero deployment. The plugin looks up the modifications from a ConfigMap in the velero namespace that is generated when the kurl installer runs. The values that are modified on the restored kotsadm deployment are proxy env vars and the host CA path.


kurl_util/Makefile
```diff
@@ -88,6 +88,9 @@ bin/network:
 bin/toml:
 	go build ${LDFLAGS} -o bin/toml cmd/toml/main.go
 
+bin/veleroplugin:
```
laverya, line 91:
```suggestion
bin/veleroplugin: cmd/veleroplugin/main.go
```

kurl_util/cmd/veleroplugin/main.go
```diff
@@ -0,0 +1,156 @@
...
+}
+
+func (p *restoreKotsadmPlugin) AppliesTo() (velero.ResourceSelector, error) {
+	return velero.ResourceSelector{
+		IncludedNamespaces: []string{"default"},
+		IncludedResources:  []string{"deployments"},
+	}, nil
+}
+
+func (p *restoreKotsadmPlugin) Execute(input *velero.RestoreItemActionExecuteInput) (*velero.RestoreItemActionExecuteOutput, error) {
```
laverya, line 47: is there any way to make unit tests for functions like this?

laverya (approved):

kurl_util/Makefile
areed: added

kurl_util/cmd/veleroplugin/main.go
areed: added

PR #1842, opened 2021-07-09
add migrate-pvcs command to tasks

laverya:
This runs `pvmigrate` to move data from rook-ceph to longhorn.

### DIFF
--- a/scripts/tasks.sh
+++ b/scripts/tasks.sh
@@ -478,7 +478,7 @@ function migrate_pvcs() {
     export KUBECONFIG=/etc/kubernetes/admin.conf
 
     # get params - specifically need airgap as that impacts binary downloads and skip-rook-health-checks to skip rook health checks
-    shift # the first param is join_token/join-token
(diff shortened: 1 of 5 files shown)


scripts/tasks.sh
```diff
@@ -469,4 +474,98 @@ EOF
...
+                exit 1
+                ;;
+        esac
+        shift
+    done
+
+    download_util_binaries
+
+    # check that rook-ceph is healthy
+    CEPH_HEALTH_DETAIL=$(kubectl exec -n rook-ceph deployment/rook-ceph-operator -- ceph health detail)
```
areed, line 506: There are some customers on Rook 1.4.3+ that may be interested in migrating to Longhorn. For those versions you'd have to exec into `rook-ceph-tools`.

areed (approved):

scripts/tasks.sh
laverya: I'll make that exec destination conditional

areed (approved):

PR #1855, opened 2021-07-12
Bundle will fail on missing addon package

emosbaugh:


laverya (approved):
AIUI, this will make the bundle creation process fail early in the situation where one of the layers is unavailable, instead of returning a partial bundle

PR #1858, opened 2021-07-12
feat(kotsadm): auto PR generatation

DanStough:
To run for alpha:
```
curl -H "Authorization: token $GH_PAT" \
    -H 'Accept: application/json' \
    -d "{\"event_type\": \"auto-kotsadm-update\", \"client_payload\": {\"version\": \"1.48.0-beta.0\" }}" \
    "https://api.github.com/repos/replicatedhq/kurl/dispatches"
```

To run for release:
```
curl -H "Authorization: token $GH_PAT" \
    -H 'Accept: application/json' \
    -d "{\"event_type\": \"auto-kotsadm-update\", \"client_payload\": {\"version\": \"1.48.0" }}" \
    "https://api.github.com/repos/replicatedhq/kurl/dispatches"
```

Will add docs to the SO.


web/src/installers/versions.js
```diff
@@ -87,6 +87,7 @@ module.exports.InstallerVersions = {
     "1.7.4",
   ],
   kotsadm: [
+    // auto-kotsadm-update
```
divolgin, line 90: not `"__KOTSADM_BINARY_VERSION__",`?

divolgin (commented):
I'm not sure if having the `templates` folder in the add on directory will break anything, but I like the general approach.

web/src/installers/versions.js
DanStough: Following convention here that it's the name of the job, e.g. :https://github.com/replicatedhq/kURL/blob/6abb99ac12abc5e07423d0e7a0c7811282a74260/web/src/installers/versions.js#L72-L73

DanStough:
Actually the `templates` directory was existing and it holds the KURL spec for testgrid that runs when a new PR is created for kotsadm. The `base` and `generate.sh` convention is borrowed from the automated updates in other addons (e.g. antrea, longhorn, goldpinger, etc).

divolgin:
The addons directory is special.  This is the first directory there that does not correspond to an addon version.

divolgin (approved):

areed (approved):

PR #1862, opened 2021-07-14
Add task to migrate RGW to Minio

areed:
Add a new sync-object-store subcommand to the kurl CLI and run it in a
Pod.
This does not update kotsadm, velero, or the registry to use Minio.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -14,6 +14,7 @@ require (
 	github.com/evanphx/json-patch v4.11.0+incompatible // indirect
 	github.com/fatih/color v1.12.0
 	github.com/foomo/htpasswd v0.0.0-20200116085101-e3a90e78da9c
+	github.com/go-ini/ini v1.62.0 // indirect
 	github.com/go-logr/zapr v0.4.0 // indirect
 	github.com/golang/groupcache v0.0.0-20210331224755-41bb18bfe9da // indirect
 	github.com/golang/mock v1.6.0
@@ -25,6 +26,7 @@ require (
 	github.com/imdario/mergo v0.3.12 // indirect
 	github.com/itchyny/gojq v0.12.4
 	github.com/mattn/go-isatty v0.0.13
+	github.com/minio/minio-go v6.0.14+incompatible // indirect
 	github.com/mitchellh/copystructure v1.2.0 // indirect
 	github.com/onsi/gomega v1.13.0
 	github.com/pelletier/go-toml v1.9.3
--- a/pkg/cli/sync_object_store.go
+++ b/pkg/cli/sync_object_store.go
@@ -1,0 +1,115 @@
+package cli
+
+import (
+	"context"
+	"fmt"
+	"log"
+
+	"github.com/minio/minio-go"
+	"github.com/spf13/cobra"
+)
+
+func NewSyncObjectStoreCmd(cli CLI) *cobra.Command {
+	var srcHost string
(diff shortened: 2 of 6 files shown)


go.mod
```diff
@@ -25,6 +26,7 @@ require (
 	github.com/imdario/mergo v0.3.12 // indirect
 	github.com/itchyny/gojq v0.12.4
 	github.com/mattn/go-isatty v0.0.13
+	github.com/minio/minio-go v6.0.14+incompatible // indirect
```
laverya, line 29: this is thankfully licensed apache 2.0

pkg/cli/sync_object_store.go
```diff
@@ -0,0 +1,115 @@
...
+	"github.com/spf13/cobra"
+)
+
+var srcHost string
+var srcAccessKeyID string
+var srcAccessKeySecret string
+
+var dstHost string
+var dstAccessKeyID string
+var dstAccessKeySecret string
```
laverya, line 18: can we move these inside the NewSyncObjectStoreCmd function?

areed: yes

pkg/cli/sync_object_store.go
```diff
@@ -0,0 +1,115 @@
...
+
+	srcObjectInfoChan := src.ListObjects(bucket, "", true, ctx.Done())
+
+	for srcObjectInfo := range srcObjectInfoChan {
+		srcObject, err := src.GetObject(bucket, srcObjectInfo.Key, minio.GetObjectOptions{})
+		if err != nil {
+			return count, fmt.Errorf("Get %s from source: %v", srcObjectInfo.Key, err)
+		}
+
+		_, err = dst.PutObject(bucket, srcObjectInfo.Key, srcObject, srcObjectInfo.Size, minio.PutObjectOptions{
```
laverya, line 102: what happens if the object already exists in the destination? Should we delete it and recreate it?

Or compare hashes to speed up things if you cancelled and reran the process?

pkg/cli/sync_object_store.go
```diff
@@ -0,0 +1,115 @@
...
+	if err != nil {
+		return count, fmt.Errorf("Failed to check if bucket %q exists in destination: %v", bucket, err)
+	}
+	if !exists {
+		if err := dst.MakeBucket(bucket, ""); err != nil {
+			return count, fmt.Errorf("Failed to make bucket %q in destination: %v", bucket, err)
+		}
+	}
+
+	srcObjectInfoChan := src.ListObjects(bucket, "", true, ctx.Done())
```
laverya, line 94: should we also do something to prune objects in dst but not src?

pkg/cli/sync_object_store.go
areed: This will overwrite whatever object is there. Yes, I can optimize with etag.

pkg/cli/sync_object_store.go
areed: I think that should be opt-in with a flag since it's destructive, but we can wait and see if anybody asks for it before we build it.

laverya (approved):

pkg/cli/sync_object_store.go
areed: Actually RGW and Minio ETags don't match for the same object so I don't think this can be optimized.

PR #1865, opened 2021-07-16
weave doesn't fail on connection-timeout

kwsorensen:


kwsorensen:
Testing in Troubleshoot /w following host configuration. 

`iptables -A INPUT -p TCP --dport 6781 -j REJECT`


```
apiVersion: troubleshoot.sh/v1beta2
kind: HostPreflight
metadata:
  name: port
spec:
  collectors:
    - tcpPortStatus:
        collectorName: k8s
        port: 6781
  analyzers:
    - tcpPortStatus:
        collectorName: k8s
        outcomes:
          - fail:
              when: "connection-refused"
              message: Connection to port 7443 was refused.
          - fail:
              when: "address-in-use"
              message: Another process was already listening on port 7443.
          - fail:
              when: "error"
              message: Unexpected port status
          - pass:
              message: Port 7443 is open

```

addons/weave/2.6.4/host-preflight.yaml
```diff
@@ -34,20 +34,14 @@ spec:
...
+          - fail:
               when: "address-in-use"
               message: Another process was already listening on port 6781.
-          - fail:
-              when: "connection-timeout"
-              message: Timed out connecting to port 6781. Check your firewall.
           - fail:
               when: "error"
               message: Unexpected port status
           - pass:
```
areed, line 43: This no longer short-circuits. Everything needs to have a when.

addons/weave/template/base/host-preflight.yaml
```diff
@@ -56,20 +50,14 @@ spec:
           - fail:
               when: "connection-refused"
               message: Connection to port 6782 was refused. This is likely to be a routing problem since this preflight configures a test server to listen on this port.
-          - warn:
+          - fails:
```
areed, line 53: fail

addons/weave/2.6.4/host-preflight.yaml
kwsorensen: I just built and tested it with the latest preflight. It appears to be working as intended without a when.

addons/weave/template/base/host-preflight.yaml
kwsorensen: good catch

kwsorensen: fixed with https://github.com/replicatedhq/kURL/pull/1865/commits/22f5a6bafa7b7d33af3d16c6da839f09590e6f67

areed (approved):

PR #1866, opened 2021-07-16
Update kotsadm, registry, and velero to use minio

areed:


scripts/tasks.sh
```diff
@@ -640,12 +641,75 @@ EOF
...
+
+            resticrepo=$(kubectl -n velero get resticrepositories --selector=velero.io/storage-location=default --no-headers | awk '{ print $1 }')
+            if [ -n "$resticrepo" ]; then
+                kubectl -n velero patch resticrepositories "$resticrepo" --type=merge -p "{\"spec\":{\"resticIdentifier\":\"s3:http://${MINIO_HOST}/velero/restic/default\"}}"
+            fi
+        else
+            echo "default backupstoragelocation was not rgw, skipping"
+        fi
+    fi
+    if kubernetes_resource_exists velero secret cloud-credentials; then
```
laverya, line 693: should we only do this if the storage location was set to rook-ceph?

laverya (approved):

scripts/tasks.sh
areed: It checks on line 694 that the creds are for RGW before overwriting them. I think it's a bit more resilient to only rely on that check so that if there was some failure after the storage location was updated to minio it will still update the credentials.

PR #1867, opened 2021-07-16
Minio spec field to auto-migrate from RGW

areed:


scripts/common/object_store.sh
```diff
@@ -54,3 +54,120 @@ function object_store_bucket_exists() {
         -H "Authorization: AWS $OBJECT_STORE_ACCESS_KEY:$sig" \
         "http://$OBJECT_STORE_CLUSTER_IP/$bucket" >/dev/null 2>&1
 }
+
+function migrate_rgw_to_minio() {
```
areed, line 58: copied from tasks.sh unchanged

areed:
The failing check is because of the spec change

kurlkinds/config/crds/v1beta1/cluster.kurl.sh_installers.yaml
```diff
@@ -330,13 +330,16 @@ spec:
...
+                  migrateFromRGW:
+                    type: boolean
                   namespace:
                     type: string
                   s3Override:
                     type: string
                   version:
                     type: string
                 required:
+                - migrateFromRGW
```
laverya, line 342: should this really be required? I think you left out omitEmpty here: https://github.com/replicatedhq/kURL/pull/1867/files#diff-28c4c4c36269f130733810745e60144350618987ea7bdfed3fe68699749ad485R130

web/src/installers/index.ts
```diff
@@ -233,6 +234,7 @@ export const minioConfigSchema = {
     s3Override: { type: "string", flag: "s3-override", description: "Override the download location for addon package distribution (used for CI/CD testing alpha addons)" },
     namespace: { type: "string", flag: "minio-namespace", description: "The namespace Minio is installed to" },
     hostPath: { type: "string", flag: "minio-hostpath", description: "Configure the minio deployment to use a local hostPath for storing data." },
+    migrateFromRGW: { type: "boolean" },
```
laverya, line 237: Should we add a flag and description for this?

kurlkinds/config/crds/v1beta1/cluster.kurl.sh_installers.yaml
areed: right

web/src/installers/index.ts
areed: I thought they weren't used anywhere

areed: but yeah, I'll add

laverya: I thought it was used to generate the tables on kurl.sh? I might be wrong though

laverya (approved):

PR #1877, opened 2021-07-19
Triggered Kotsadm version update 1.47.2

replicatedbot:
Automated changes by the [auto-kotsadm-update](https://github.com/replicatedhq/kURL/blob/master/.github/workflows/update-kotsadm.yaml) GitHub action

### DIFF
--- a/web/src/installers/versions.js
+++ b/web/src/installers/versions.js
@@ -88,6 +88,7 @@ module.exports.InstallerVersions = {
   ],
   kotsadm: [
     // auto-kotsadm-update
+    "v1.47.2",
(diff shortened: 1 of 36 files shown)


web/src/installers/versions.js
```diff
@@ -88,6 +88,7 @@ module.exports.InstallerVersions = {
   ],
   kotsadm: [
     // auto-kotsadm-update
+    "v1.47.2",
```
sgalsaleh, line 91:
```suggestion
    "1.47.2",
```

sgalsaleh (approved):

PR #1885, opened 2021-07-22
adding a docker folder space check

kwsorensen:


pkg/preflight/assets/host-preflights.yaml
```diff
@@ -115,6 +118,21 @@ spec:
...
+              when: "total < 30Gi"
+              message: The disk containing directory /var/lib/docker has less than 30Gi of total space
+          - fail:
+              when: "used/total > 80%"
+              message: The disk containing directory /var/lib/docker is more than 80% full
+          - warn:
+              when: "used/total > 60%"
+              message: The disk containing directory /var/lib/docker is more than 60% full
+          - pass:
+              message: The disk containing directory /var/lib/docker has at least 30Gi disk space available
```
laverya, line 135: we should only run this analyzer if the docker CRI is selected - it isn't needed for containerd

kwsorensen: done by adding an exclude to collector and analyzer

laverya (approved):

PR #1890, opened 2021-07-23
fix kotsadm has alpha version which is not in semver format

emosbaugh:


web/src/test/controllers/installers.ts
```diff
@@ -1016,7 +1016,7 @@ spec:
       const bad = function (): string {
         return Installer.resolveLatestPatchVersion("1.1.x", InstallerVersions["minio"]);
       };
-      expect(bad).to.throw("Invalid Version");
+      expect(bad).to.throw("latest minor version not found for 1.1.x");
```
kwsorensen, line 1019: should this be "latest patch version not found for 1.1.x"

emosbaugh: Thanks!

kwsorensen (approved):

PR #1907, opened 2021-07-28
remove rook ceph task

laverya:


scripts/tasks.sh
```diff
@@ -600,4 +603,76 @@ function migrate_rgw_to_minio_task() {
...
+    # scale ekco to 0 replicas if it exists
+    if kubernetes_resource_exists kurl deployment ekc-operator; then
+        kubectl -n kurl scale deploy ekc-operator --replicas=0
+    fi
+
+    # remove all rook-ceph CR objects
+    printf "Removing rook-ceph custom resource objects - this may take some time:\n"
+    kubectl delete cephcluster -n rook-ceph rook-ceph # deleting this first frees up resources
+    kubectl get crd | grep 'ceph.rook.io' | awk '{ print $1 }' | xargs -I'{}' kubectl -n rook-ceph delete '{}' --all
+    kubectl delete kubectl delete volumes.rook.io --all
```
areed, line 650:
```suggestion
    kubectl delete volumes.rook.io --all
```

areed (commented):
This should unset `rook_priority_class` and `maintain_rook_storage_nodes` in the ekc-operator ConfigMap scaling ekco up again so it doesn't try to perform these two tasks:

https://github.com/replicatedhq/ekco/blob/master/pkg/ekcoops/operator.go#L57.
https://github.com/replicatedhq/ekco/blob/master/pkg/ekcoops/operator.go#L72.

scripts/tasks.sh
laverya: 🤦

areed (approved):

PR #1919, opened 2021-07-29
Upgrade tsed

emosbaugh:


web/src/commands/serve.ts
```diff
@@ -25,7 +28,17 @@ export async function main(argv: any): Promise<void> {
 
   metrics.bootstrapFromEnv();
 
-  await new Server(
-    argv.bugsnagKey,
```
sgalsaleh, line 29: so we don't use bugsnag for this anymore?

sgalsaleh (approved):

web/src/commands/serve.ts
emosbaugh: i dont think this variable was used

emosbaugh:
[ch36256]

PR #1948, opened 2021-08-02
Npm upgrades and fixes

emosbaugh:


web/src/client/minio_test.ts
```diff
@@ -23,9 +23,9 @@ describe("script with minio config", () => {
 	it("200", async () => {
 		const uri = await client.postInstaller(minio);
 
-		expect(uri).to.match(/b737e21/);
+		expect(uri).to.match(/5e5ebc3/);
```
emosbaugh, line 26: Looks like this change also changed the hash https://github.com/replicatedhq/kURL/pull/1926/files

sgalsaleh (approved):

PR #1962, opened 2021-08-03
failure skipped

kwsorensen:


addons/rook/1.4.3/install.sh
```diff
@@ -13,7 +13,7 @@ function rook_pre_init() {
             SKIP_KUBERNETES_HOST=1
         fi
     elif [ "$ROOK_BLOCK_STORAGE_ENABLED" != "1" ]; then
-        bail "Rook 1.4.3 requires enabling block storage"
+        logWarn "Rook 1.4.3 requires enabling block storage"
```
laverya, line 16: What's the point to warning here?

Why not just remove all references to `ROOK_BLOCK_STORAGE_ENABLED` in 1.4.3+?

kwsorensen: Well the goal was to inform users that the configuration was going to be ignored. Really I just wanted something temporary so that I could launch in test grid and it wouldn't fail. Obviously it's failing to create TG runs while that api validation is throwing errors for Rook version > 1.4.* without an explicit isBlockStorageEnabled.

kwsorensen: On further thought this warning will probably only confuse people.

web/src/client/index_test.ts
```diff
@@ -877,7 +877,7 @@ describe("POST /installer/validate", () => {
       } catch (error) {
         err = error;
       }
-      expect(err).to.have.property("message", "Rook versions >= 1.4.0 require blockStorageEnabled to be set to true");
+      expect(err).to.be.undefined;
```
laverya, line 880: honestly a lot of these tests can probably be removed entirely now that we're getting rid of the check

laverya (approved):

PR #2006, opened 2021-08-10
Migrate from docker to containerd

areed:
When docker is detected and the kurl spec has containerd follow these steps to migrate prior to installing containerd:

1. Run the ekco shutdown script to delete pods that depend on other pods on the node being running, i.e. pods that mount PVCs.
1. Delete all remaining pods from the node. This has to use the kubelet's kubeconfig since that's the only kubeconfig available on both primaries and secondaries. It does not have permission to do a drain, so manually delete every pod by UID.
1. Stop kubelet.
1. Force delete all docker containers.
1. Reconfigure kubelet to use containerd instead of the dockershim. This is done by adding `--container-runtime` and `--container-runtime` endpoint flags to /var/lib/kubelet/kubeadm-flags.env. Also patch the kubeadm.conf file to make these changes persistent.

When the script is complete, perform these steps to finish the migration.
1. Remove docker and its data
1. Start kubelet (remotes only - if kubeadm init was run kubelet will already be started).
1. Uncordon the node (remotes only - if kubeadm init was run the node will already be uncordoned).

### DIFF
--- a/addons/containerd/1.3.7/install.sh
+++ b/addons/containerd/1.3.7/install.sh
@@ -7,6 +7,8 @@ function containerd_install() {
         bail "The filesystem mounted at /var/lib/containerd does not have ftype enabled"
     fi
 
+    containerd_migrate_from_docker
+
     install_host_packages "$src" containerd.io
 
     case "$LSB_DIST" in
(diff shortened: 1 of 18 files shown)


murphybytes (changes requested):
would it make sense to share the containerd_migrate_from_docker function?

murphybytes (approved):

PR #2011, opened 2021-08-11
Added support for NO_MASQ_LOCAL in weave addon

murphybytes:
Allows end user to set NO_MASQ_LOCAL for the weave add on. See [Clubhouse Issue](https://app.clubhouse.io/replicated/story/35183/allow-opt-in-to-preserve-client-source-ip-addresses-when-using-weave)


kurlkinds/config/crds/v1beta1/cluster.kurl.sh_installers.yaml
```diff
@@ -1,10 +1,10 @@
 
 ---
-apiVersion: apiextensions.k8s.io/v1
+apiVersion: apiextensions.k8s.io/v1beta1
 kind: CustomResourceDefinition
 metadata:
   annotations:
-    controller-gen.kubebuilder.io/version: v0.6.1
+    controller-gen.kubebuilder.io/version: v0.2.5
```
areed, line 7: This won't work on k8s 1.22. Could you regen with latest kubebuilder?

addons/weave/2.5.2/tmpl-patch-no-masq-local.yaml
```diff
@@ -0,0 +1,14 @@
...
+  name: weave-net
+  namespace: kube-system
+spec:
+  template:
+    spec:
+      containers:
+        - name: weave
+          env:
+          - name: NO_MASQ_LOCAL
+            value: $NO_MASQ_LOCAL
```
areed, line 14: This works without quotes?

kurlkinds/pkg/apis/cluster/v1beta1/installer_types.go
```diff
@@ -187,6 +187,10 @@ type Weave struct {
 	PodCidrRange         string `json:"podCidrRange,omitempty" yaml:"podCidrRange,omitempty"`
 	S3Override           string `json:"s3Override,omitempty" yaml:"s3Override,omitempty"`
 	Version              string `json:"version" yaml:"version"`
+	// NoMasqLocal if not present defaults to 1, which will expose the original client IP address in the
+	// X-Forwarded-For header.  If present and assigned to 0 the IP address of the Weave interface is returned
+	// in the header.
+	NoMasqLocal *int `json:"noMasqLocal,omitempty" yaml:"noMasqLocal,omitempty"`
```
areed, line 193: It would be nicer to expose this to the user as a boolean.

areed (commented):
It will need to be added to https://github.com/replicatedhq/kURL/blob/master/web/src/installers/index.ts#L105 and then a test in https://github.com/replicatedhq/kURL/tree/master/web/src/client would be great.

areed (approved):

PR #2014, opened 2021-08-12
feat: add support for kotsadm snapshot migration

DanStough:
Add migration to kotsadm for minio-based snapshots.

Add lvp plugin and sidecar to the default Velero install in case anyone leverages this feature.

### DIFF
--- a/addons/kotsadm/alpha/install.sh
+++ b/addons/kotsadm/alpha/install.sh
@@ -1,3 +1,4 @@
+#!/usr/bin/env bash
 
 function kotsadm() {
     local src="$DIR/addons/kotsadm/alpha"
@@ -14,6 +15,7 @@ function kotsadm() {
         KOTSADM_DISABLE_S3="1"
     fi
 
+    # Migrate kotsadm deployment to statefulset
     if [ "$KOTSADM_DISABLE_S3" == "1" ]; then
         cp "$DIR/addons/kotsadm/alpha/statefulset/kotsadm-statefulset.yaml" "$DIR/kustomize/kotsadm/kotsadm-statefulset.yaml"
         insert_resources "$dst/kustomization.yaml" kotsadm-statefulset.yaml
@@ -112,6 +114,11 @@ function kotsadm() {
     kubectl label pvc kotsadm-postgres-kotsadm-postgres-0 velero.io/exclude-from-backup- kots.io/backup=velero --overwrite
 
     kotsadm_cli $src
+
+    # Migrate existing hostpath and nfs snapshot minio to velero lvp plugin
+    if [ "$KOTSADM_DISABLE_S3" == "1" ] && [ -n "$VELERO_VERSION" ] ; then
+        kubectl kots velero migrate-minio-filesystems -n default
+    fi
 }
 
 function kotsadm_join() {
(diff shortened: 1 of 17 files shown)


addons/kotsadm/alpha/install.sh
```diff
@@ -107,6 +109,11 @@ function kotsadm() {
     kubectl label pvc kotsadm-postgres-kotsadm-postgres-0 velero.io/exclude-from-backup- kots.io/backup=velero --overwrite
 
     kotsadm_cli $src
+
+    # Migrate existing hostpath and nfs snapshot minio to
```
sgalsaleh, line 113: to what? :)

addons/kotsadm/template/base/install.sh
```diff
@@ -15,6 +16,7 @@ function kotsadm() {
         KOTSADM_DISABLE_S3="1"
     fi
 
+    # Migrate kotsadm deployments to statefulset
```
sgalsaleh, line 19:
```suggestion
    # Migrate kotsadm deployment to statefulset
```

addons/kotsadm/alpha/install.sh
```diff
@@ -15,6 +16,7 @@ function kotsadm() {
         KOTSADM_DISABLE_S3="1"
     fi
 
+    # Migrate kotsadm deployments to statefulset
```
sgalsaleh, line 19:
```suggestion
    # Migrate kotsadm deployment to statefulset
```

addons/kotsadm/alpha/install.sh
```diff
@@ -107,6 +109,11 @@ function kotsadm() {
     kubectl label pvc kotsadm-postgres-kotsadm-postgres-0 velero.io/exclude-from-backup- kots.io/backup=velero --overwrite
 
     kotsadm_cli $src
+
+    # Migrate existing hostpath and nfs snapshot minio to 
+    if [ "$KOTSADM_DISABLE_S3" == "1" ]; then
+        kubectl kots velero migrate-minio-filesystems -n default
```
sgalsaleh, line 115: have we thought about all the scenarios that this command might affect? what if the velero addon is not included for example? i see that the command will fail if it didn't detect velero https://github.com/replicatedhq/kots/blob/b61a9a23f7d746dc8eaeabdd90fbeff2170e23dc/pkg/kotsadm/minio.go#L142

sgalsaleh: what if an object store isn't used?

sgalsaleh: i know that customers need to explicitly pass this flag now, but it will be opt-out in the near future

addons/velero/1.6.0/Manifest
```diff
@@ -3,5 +3,7 @@ image restic-restore velero/velero-restic-restore-helper:v1.6.0
 image velero-aws velero/velero-plugin-for-aws:v1.2.0
 image velero-gcp velero/velero-plugin-for-gcp:v1.2.0
 image velero-azure velero/velero-plugin-for-microsoft-azure:v1.2.0
+image local-volume-provider replicated/local-volume-provider:v0.1.0
```
sgalsaleh, line 6: will this plugin not work with velero 1.5.x? i don't see it included there

addons/kotsadm/alpha/install.sh
DanStough: 🤦

addons/velero/1.6.0/Manifest
DanStough: I did some compatibility testing and was able to successfully do a backup and restore on 1.5.x with LVP, so I will add this.

sgalsaleh (approved):

kwsorensen (approved):

PR #2023, opened 2021-08-13
License prompt update

kwsorensen:


laverya (approved):

scripts/common/prompts.sh
```diff
@@ -156,7 +156,7 @@ function common_prompts() {
 
 function prompt_license() {
     if [ -n "$LICENSE_URL" ]; then
-        if ["$AIRGAP" = "1"]; then
+        if [ "$AIRGAP" = "1"  ]; then
```
laverya, line 159:
```suggestion
        if [ "$AIRGAP" = "1" ]; then
```

laverya: might as well make it symmetrical 😆

kwsorensen: spacing is hard

PR #2044, opened 2021-08-17
Revert "Revert "Migrate from docker to containerd""

laverya:
Reverts replicatedhq/kURL#2025

### DIFF
--- a/scripts/common/common.sh
+++ b/scripts/common/common.sh
@@ -334,7 +334,7 @@ function kubernetes_resource_exists() {
 
 function install_cri() {
     # In the event someone changes the installer spec from docker to containerd, maintain backward capability with old installs
-    if [ -n "$DOCKER_VERSION" ] || [ "$SKIP_DOCKER_INSTALL" = "1" ] ; then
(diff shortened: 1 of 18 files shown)


laverya:
one weird thing to me - this https://github.com/replicatedhq/kURL/blob/0dd309e901b07db2ee19d26a8e349155e8a8d291/scripts/common/docker.sh#L79 is run as part of the post-init functions, which happens after all the addons are applied. How is that supposed to work if kubelet is stopped during the containerd install/migration?

Should it be run in the containerd code instead?

scripts/common/common.sh
```diff
@@ -385,6 +385,26 @@ function try_1m() {
     done
 }
 
+# try a command every 2 seconds until it succeeds, up to 150 tries max; useful for kubectl commands
+# where the Kubernetes API could be restarting
+function try_5m() {
```
laverya, line 390: this doesn't actually try for 5m - it tries 150 times
if each try takes 6s, then with the 2s wait it would be 20m

PR #2047, opened 2021-08-20
add minio.claimSize param to configure the size of the Minio PVC

laverya:


addons/minio/template/testgrid/k8s-docker.yaml
```diff
@@ -36,3 +36,29 @@
...
+      version: "latest"
+    weave:
+      version: "latest"
+    longhorn:
+      version: "latest"
+    containerd:
+      version: "latest"
+    minio:
+      version: "__testver__"
+#      claimSize: "20Gi"
```
murphybytes, line 51: Is this a comment?  If it is, please add some detail, otherwise get rid of it.

laverya: it's a "I want to have a test with this param set, but that requires merging first because that param doesn't exist otherwise"

I can add a comment signifying that

murphybytes (changes requested):
This needs a test, fortunately, there is already one that partially verifies the [change here](https://github.com/replicatedhq/kURL/blob/7ddaafc47ac94d2ca366593745aef986cf329a0e/kurl_util/cmd/yamltobash/main_test.go#L360).  Which as it turns out is busted now because you've added the claimSize param.

laverya:
That test code is FAR too brittle IMO @murphybytes - can we change it to only look for the desired param and not the entire set of environment variables/ordering/etc?

murphybytes:
I agree that it's too brittle, my bad because I wrote it in the first place, I agree I should've used a regex or something to just look for the change under test.

murphybytes (approved):

PR #2061, opened 2021-08-24
Fix internal lb remote

areed:


emosbaugh (approved):
If you answer yes to the prompt does the installer spec cr get updated? how do we handle other prompts?

PR #2076, opened 2021-09-01
update preflight checks to remove default fall-through

kwsorensen:


pkg/preflight/assets/host-preflights.yaml
```diff
@@ -159,8 +163,6 @@ spec:
           - pass:
               when: "connected"
               message: Successfully connected to {{kurl .Installer.Spec.Kubernetes.LoadBalancerAddress }} via load balancer
-          - warn:
-              message: Unexpected port status
```
laverya, line 163: I believe these were _intended_ as fallthrough cases - instead of `when != invalid-address && != connection-refused...`

murphybytes: The code needs to be modified such that if there are no outcomes that the when resolves to true, the default case is returned, in this case 'Unexpected port status'.  In general pass, fail, and default are mutually exclusive.  In some analyzers multiple warning outcomes can be returned if it makes sense to do so.    So what I'd like to see here is code that enforces that the default case is returned IFF no other outcomes are true, along with unit tests.

murphybytes (changes requested):
Please add some context such as links to related Clubhouse issues in the PR description.

PR #2096, opened 2021-09-09
move block device checks to addon code

laverya:
this way it always runs for rook 1.4+

also add disk space check for /opt/replicated/rook when appropriate

finally check available, not just total and percentage, space in /var/lib/docker and /var/lib/kubelet

### DIFF
--- a/addons/openebs/1.6.0/host-preflight.yaml
+++ b/addons/openebs/1.6.0/host-preflight.yaml
@@ -1,0 +1,25 @@
+apiVersion: troubleshoot.sh/v1beta2
+kind: HostPreflight
+metadata:
+  name: kurl-builtin
+spec:
+  collectors:
+  - blockDevices:
+      # cstor is enabled and not upgrade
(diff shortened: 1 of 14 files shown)


addons/openebs/1.6.0/host-preflight.yaml
```diff
@@ -0,0 +1,25 @@
+apiVersion: troubleshoot.sh/v1beta2
+kind: HostPreflight
+metadata:
+  name: kurl-builtin
+spec:
+  collectors:
+  - blockDevices:
+      # cstor is enabled and not upgrade
+      exclude: '{{kurl and (and .Installer.Spec.OpenEBS.Version .Installer.Spec.OpenEBS.IsCstorEnabled) (not .IsUpgrade) | not }}'
```
kwsorensen, line 9: Is there a simpler way to write this exclude?

kwsorensen (approved):

addons/openebs/1.6.0/host-preflight.yaml
laverya: maybe, but this is copied (and cropped) from the original

PR #2112, opened 2021-09-15
nightly addon for kotsadm

divolgin:


addons/kotsadm/nightly/deployment/kotsadm-deployment.yaml
```diff
@@ -0,0 +1,259 @@
...
+        configMap:
+          defaultMode: 511 # hex 777
+          name: kotsadm-web-scripts
+      - name: backup
+        emptyDir: {}
+      - name: migrations
+        emptyDir:
+          medium: Memory
+      containers:
+      - image: kotsadm/kotsadm:v0.0.0-nightly
```
laverya, line 46: Should we add `imagePullPolicy: Never` here? to prevent this from getting updated unexpectedly?

That being said, if you're using nightly in production you're going to get some instability anyways...

divolgin: I think `IfNotPresent`, which should be the default for this tag, is ok here.

laverya (approved):

addons/kotsadm/nightly/deployment/kotsadm-deployment.yaml
laverya: unless you're running multiple nodes, and kotsadm gets rescheduled to a node without the image

PR #2174, opened 2021-09-27
Enable IP forwarding in /etc/sysctl.conf

areed:
This file takes precedence over files in /etc/sysctl.d and recently there
have been more nodes that are setting net.ipv4.ip_forward=0 in /etc/sysctl.conf.

Enable ip forward on all operating systems, not just EL.

Add a post-flight check that ip forwarding is enabled after reading the
new settings in /etc/sysctl.conf.

### DIFF
--- a/scripts/common/kubernetes.sh
+++ b/scripts/common/kubernetes.sh
@@ -52,13 +52,13 @@ function kubernetes_load_ipvs_modules() {
 }
 
 function kubernetes_sysctl_config() {
-    if ! grep -q "net.bridge.bridge-nf-call-ip6tables = 1" /etc/sysctl.conf; then
+    if ! cat /etc/sysctl.conf | grep "^net.bridge.bridge-nf-call-ip6tables" | tail -n1 | grep -Eq "^net.bridge.bridge-nf-call-ip6tables\s*=\s*1"; then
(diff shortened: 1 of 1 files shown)


scripts/common/kubernetes.sh
```diff
@@ -52,16 +52,21 @@ function kubernetes_load_ipvs_modules() {
 }
 
 function kubernetes_sysctl_config() {
-    case "$LSB_DIST" in
-        # TODO I've only seen these disabled on centos/rhel but should be safe for ubuntu
-        centos|rhel|amzn|ol)
-            echo "net.bridge.bridge-nf-call-ip6tables = 1" > /etc/sysctl.d/k8s.conf
-            echo "net.bridge.bridge-nf-call-iptables = 1" >> /etc/sysctl.d/k8s.conf
-            echo "net.ipv4.conf.all.forwarding = 1" >> /etc/sysctl.d/k8s.conf
+    if ! grep -q "net.bridge.bridge-nf-call-ip6tables = 1" /etc/sysctl.conf; then
```
divolgin, line 55: This grep will fail if there isn't exactly one space character before and after the equal sign.  Could be better to split on `=` and trim spaces and check if value is not `1`

PR #2176, opened 2021-09-28
fix: snapshot migration when velero already applied

DanStough:
Is you already have the same version of velero (or minio) installed when you try to migrate from Rook to Longhorn, the migration code will not be picked up because the `_already_applied` lifecycle hook is run in the addon instead of `install`.

### DIFF
--- a/addons/velero/1.6.2/install.sh
+++ b/addons/velero/1.6.2/install.sh
@@ -66,6 +66,18 @@ function velero_install() {
     rm velero-credentials
 }
 
+function velero_already_applied() {
+    local src="$DIR/addons/velero/$VELERO_VERSION"
+    local dst="$DIR/kustomize/velero"
+
+    velero_change_storageclass "$src" "$dst" true
+
+    # This should only be applying the configmap if required
+    if [ -d "$dst" ]; then
+        kubectl apply -f "$dst"
+    fi
+}
+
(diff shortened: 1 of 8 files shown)


addons/velero/1.6.2/install.sh
```diff
@@ -150,10 +162,15 @@ function velero_patch_http_proxy() {
...
     local src="$1"
     local dst="$2"
+    local disable_kustomization="$3"
 
     if kubectl get sc longhorn &> /dev/null && \
     [ "$(kubectl get sc longhorn -o jsonpath='{.metadata.annotations.storageclass\.kubernetes\.io/is-default-class}')" = "true" ]; then
         render_yaml_file "$src/tmpl-change-storageclass.yaml" > "$dst/change-storageclass.yaml"
-        insert_resources "$dst/kustomization.yaml" change-storageclass.yaml
+        if [ -z "$disable_kustomization" ]; then
+            insert_resources "$dst/kustomization.yaml" change-storageclass.yaml
```
jala-dx, line 171: does this have to be $dst/change-storageclass.yaml

DanStough: Per our discussion, we don't need this since all the kustomize resources are in the same directory (and not nested).

jala-dx (approved):

PR #2177, opened 2021-09-29
ability to configure log rotation for kubelet

sgalsaleh:


scripts/install.sh
```diff
@@ -119,17 +119,24 @@ apiVersion: kubelet.config.k8s.io/v1beta1
...
 cgroupDriver: systemd
----
 EOF
     fi
 
+    # conditional kubelet configuration fields
+    if [ -n "$CONTAINER_LOG_MAX_SIZE" ]; then
+        echo "containerLogMaxSize: $CONTAINER_LOG_MAX_SIZE" >> $KUBEADM_CONF_FILE
+    fi
+    if [ $CONTAINER_LOG_MAX_FILES -gt 0 ]; then
```
areed, line 135: Is it possible this is unset? If so bash will print an error.

areed (approved):

scripts/install.sh
sgalsaleh: i tried it both providing a value and not, seemed to work fine. i guess it just defaults to zero since that's the behavior in Go. i will validate it once more to make sure.

sgalsaleh: nevermind, you are right

areed (approved):

PR #2179, opened 2021-09-29
feat(registry): multiple replicas when rwx pvcs available

DanStough:


addons/registry/2.7.1/install.sh
```diff
@@ -20,6 +20,16 @@ function registry() {
         render_yaml_file "$DIR/addons/registry/2.7.1/tmpl-persistentvolumeclaim.yaml" > "$DIR/kustomize/registry/persistentvolumeclaim.yaml"
         insert_resources "$DIR/kustomize/registry/kustomization.yaml" deployment-pvc.yaml
         insert_resources "$DIR/kustomize/registry/kustomization.yaml" persistentvolumeclaim.yaml
+
+        # check for readwritemany volume providers. Longhorn >=1.1.0 or Rook >=1.6.0
+        # rook case should never run, but keeping for future capability
+        if [ -n "$ROOK_VERSION" ]; then
+            semverCompare "$ROOK_VERSION" "1.6.0"
```
jala-dx, line 27: an error check would be good

```
    if [ "$SEMVER_COMPARE_RESULT" = "-1" ]; then
        printf "cannot create additional replica for Rook version $ROOK_VERSION\n"
     
    fi

```

addons/registry/2.7.1/install.sh
```diff
@@ -20,6 +20,16 @@ function registry() {
...
+            maybe_rwx_patch_registry_pvc "$SEMVER_COMPARE_RESULT" 
+        elif [ -n "$LONGHORN_VERSION" ]; then
+            semverCompare "$LONGHORN_VERSION" "1.1.0"
+            maybe_rwx_patch_registry_pvc "$SEMVER_COMPARE_RESULT"
```
jala-dx, line 31: same as above

jala-dx (approved):

addons/registry/2.7.1/install.sh
DanStough: I don't think it's an error though, since it's possible that the provider doesn't support RWX volumes. We would just use the default. I would worry about alarming the use for something that is expected behavior.

How about I add an `echo` in the `maybe_rwx_patch_registry_pvc` like "detected RWX capabilities, scaling our registry deployment"?

jala-dx: echo sounds good so we know what is going on if there is an unsupported version.

PR #2180, opened 2021-09-29
checking hashed tokens

jeffreygolden:
For web access (promoting installers, etc), tokens are required for authentication.  New User and Service Account tokens are sha256 hashed and base64 encoded.  kURL must check for an unhashed team token first, and if no results are found, must hash the token and check again for a user or service account token.

### DIFF
--- a/web/src/util/jwt/index.ts
+++ b/web/src/util/jwt/index.ts
@@ -1,6 +1,6 @@
 import * as jwt from "jsonwebtoken";
 import * as _ from "lodash";
-import CryptoJS from 'crypto-js';
+import * as crypto from 'crypto';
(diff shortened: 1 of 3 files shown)


web/src/util/jwt/index.ts
```diff
@@ -1,5 +1,6 @@
 import * as jwt from "jsonwebtoken";
 import * as _ from "lodash";
+import CryptoJS from 'crypto-js';
```
sgalsaleh, line 3: i don't see this in package.json

jeffreygolden: added

sgalsaleh (approved):

divolgin (approved):

PR #2183, opened 2021-09-30
defaults for docker log rotation

sgalsaleh:


areed (commented):
I think it would be better not add another binary to kurl and not to have to maintain the insertOrReplaceJsonParam code.  How about something like this instead?

```
function init_daemon_json() {
    if [ -f /etc/docker/daemon.json ]; then
        return
    fi

    mkdir -p /etc/docker

    if [ ! -f /var/lib/kubelet/kubeadm-flags.env ]; then

        cat > /etc/docker/daemon.json <<EOF
{
    "log-driver": "json-file",
    "log-opts": {"max-size": "10m"},
    "exec-opts": ["native.cgroupdriver=systemd"]
}
EOF
        return 0
    fi

    cat > /etc/docker/daemon.json <<EOF
{
    "log-driver": "json-file",
    "log-opts": {"max-size": "10m"},
}
EOF

    mkdir -p /etc/systemd/system/docker.service.d
}
```

sgalsaleh:
yeah, i was initially going with that approach, but i felt it's gonna go out of control the more we add conditional options to it. i guess it should be fine for now. i agree, will update.

scripts/common/docker.sh
```diff
@@ -1,20 +1,33 @@
...
+    "log-driver": "json-file",
+    "log-opts": {"max-size": "10m"},
     "exec-opts": ["native.cgroupdriver=systemd"]
 }
 EOF
+    else
+        cat > /etc/docker/daemon.json <<EOF
+{
+    "log-driver": "json-file",
+    "log-opts": {"max-size": "10m"},
```
areed, line 27: Trailing comma

areed (approved):

PR #2187, opened 2021-10-01
fix checking service token auths

sgalsaleh:


web/package.json
```diff
@@ -14,8 +14,8 @@
     "@tsed/platform-express": "^6.62.2",
     "ajv": "^8.6.2",
     "aws-sdk": "^2.964.0",
+    "base64url": "^3.0.1",
     "cors": "^2.8.5",
-    "crypto": "^1.0.1",
```
sgalsaleh, line 18: this is a built-in package in nodejs

jeffreygolden (approved):

PR #2207, opened 2021-10-05
Updated bail message when private address can not be resolved

murphybytes:
See [Short Cut Issue](https://app.shortcut.com/replicated/story/38223/when-an-installation-fails-because-a-private-ip-address-is-required-include-remedies-in-error-message)

### DIFF
--- a/scripts/common/prompts.sh
+++ b/scripts/common/prompts.sh
@@ -311,7 +311,7 @@ function prompt_for_private_ip() {
     fi
 
     if ! prompts_can_prompt ; then
-        bail "kurl.privateAddress required"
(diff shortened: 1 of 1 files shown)


scripts/common/prompts.sh
```diff
@@ -311,7 +311,7 @@ function prompt_for_private_ip() {
     fi
 
     if ! prompts_can_prompt ; then
-        bail "kurl.privateAddress required"
+        bail "kurl.privateAddress required. Try passing the private address to this script e.g. 'sudo ./install.sh private-address=1.2.3.4' or supply the privateAddress field to the kurl add-on."
```
DanStough, line 314: I wouldn't know what `privateAddress` is here. Can we say something about multiple NICs and needing to select one?

DanStough: Maybe we can re-use lines 318-321?

DanStough (approved):

PR #2268, opened 2021-10-21
migrate registry from s3 to pvc

jala-dx:


addons/registry/2.7.1/install.sh
```diff
@@ -14,6 +14,19 @@ function registry() {
...
+            insert_patches_strategic_merge "$DIR/kustomize/registry/kustomization.yaml" patch-deployment-migrate-s3.yaml
+            cp "$DIR/addons/registry/2.7.1/patch-pvc-for-migrate-s3.yaml" "$DIR/kustomize/registry/patch-pvc-for-migrate-s3.yaml"
+            insert_patches_strategic_merge "$DIR/kustomize/registry/kustomization.yaml" patch-pvc-for-migrate-s3.yaml
+            determine_registry_pvc_size
+            render_yaml_file "$DIR/addons/registry/2.7.1/tmpl-persistentvolumeclaim.yaml" > "$DIR/kustomize/registry/persistentvolumeclaim.yaml"
+            insert_resources "$DIR/kustomize/registry/kustomization.yaml" persistentvolumeclaim.yaml
+            render_yaml_file "$DIR/addons/registry/2.7.1/tmpl-configmap-migrate-s3.yaml" > "$DIR/kustomize/registry/configmap-migrate-s3.yaml"
+            insert_resources "$DIR/kustomize/registry/kustomization.yaml" configmap-migrate-s3.yaml
+        fi
     else
```
emosbaugh, line 30: this will migrate the data but it will not reconfigure the registry to use the pvc. i think you may want something like this.

```suggestion
    fi

    if ! registry_pvc_exists && object_store_exists; then
```

emosbaugh: Also you may want to delete the registry service or something like that so that nothing can write to the registry during the migration

https://github.com/replicatedhq/kURL/blob/master/addons/registry/2.7.1/service.yaml

addons/registry/2.7.1/install.sh
```diff
@@ -14,6 +14,19 @@ function registry() {
         insert_patches_strategic_merge "$DIR/kustomize/registry/kustomization.yaml" patch-deployment-velero.yaml
         render_yaml_file "$DIR/addons/registry/2.7.1/tmpl-configmap-velero.yaml" > "$DIR/kustomize/registry/configmap-velero.yaml"
         insert_resources "$DIR/kustomize/registry/kustomization.yaml" configmap-velero.yaml
+
+        # Start migration only when usePvcStorage flag is enabled in the installer spec
+        if [ "$REGISTRY_USE_PVC_STORAGE" = "1" ]; then
```
emosbaugh, line 19: I would log a warning here that a migration is taking place

addons/registry/2.7.1/install.sh
```diff
@@ -14,6 +14,19 @@ function registry() {
...
+            cp "$DIR/addons/registry/2.7.1/patch-deployment-migrate-s3.yaml" "$DIR/kustomize/registry/patch-deployment-migrate-s3.yaml"
+            insert_patches_strategic_merge "$DIR/kustomize/registry/kustomization.yaml" patch-deployment-migrate-s3.yaml
+            cp "$DIR/addons/registry/2.7.1/patch-pvc-for-migrate-s3.yaml" "$DIR/kustomize/registry/patch-pvc-for-migrate-s3.yaml"
+            insert_patches_strategic_merge "$DIR/kustomize/registry/kustomization.yaml" patch-pvc-for-migrate-s3.yaml
+            determine_registry_pvc_size
+            render_yaml_file "$DIR/addons/registry/2.7.1/tmpl-persistentvolumeclaim.yaml" > "$DIR/kustomize/registry/persistentvolumeclaim.yaml"
+            insert_resources "$DIR/kustomize/registry/kustomization.yaml" persistentvolumeclaim.yaml
+            render_yaml_file "$DIR/addons/registry/2.7.1/tmpl-configmap-migrate-s3.yaml" > "$DIR/kustomize/registry/configmap-migrate-s3.yaml"
+            insert_resources "$DIR/kustomize/registry/kustomization.yaml" configmap-migrate-s3.yaml
```
emosbaugh, line 28: and log a success message

emosbaugh:
This doesnt handle the case of a new install with REGISTRY_USE_PVC_STORAGE set

jala-dx:
> This doesnt handle the case of a new install with REGISTRY_USE_PVC_STORAGE set

@emosbaugh this flag is only internal and will not be used in production. @ajp-io has a different story to handle all the migrations in one shot

emosbaugh:
> > This doesnt handle the case of a new install with REGISTRY_USE_PVC_STORAGE set
> 
> @emosbaugh this flag is only internal and will not be used in production. @ajp-io has a different story to handle all the migrations in one shot

I dont understand what you mean by internal. youre adding it as part of the spec.

ajp-io:
@emosbaugh I think what Jalaja is getting at is that there is no reason for customers to trigger this migration today because we don't have all the migrations in place yet. Since you wouldn't be able to migrate the local storage of snapshots, we don't yet have a full migration path, so it's moot.

In general, we don't really need a flag for this migration—it's going to be automatic, not triggered by a flag. But Jalaja needed the flag so that she could trigger it for testing. So I think her point is that the flag is there so it can be tested, but ultimately we won't publish the flag (and I think we can remove it once we get all the migrations orchestrated together).

jala-dx:
> > > This doesnt handle the case of a new install with REGISTRY_USE_PVC_STORAGE set
> > 
> > 
> > @emosbaugh this flag is only internal and will not be used in production. @ajp-io has a different story to handle all the migrations in one shot
> 
> I dont understand what you mean by internal. youre adding it as part of the spec.

@emosbaugh using the env var instead of a flag. We do not really need a flag atm

jala-dx:
> @emosbaugh I think what Jalaja is getting at is that there is no reason for customers to trigger this migration today because we don't have all the migrations in place yet. Since you wouldn't be able to migrate the local storage of snapshots, we don't yet have a full migration path, so it's moot.
> 
> In general, we don't really need a flag for this migration—it's going to be automatic, not triggered by a flag. But Jalaja needed the flag so that she could trigger it for testing. So I think her point is that the flag is there so it can be tested, but ultimately we won't publish the flag (and I think we can remove it once we get all the migrations orchestrated together).

I removed the flag and using ENV variable and tested it.

addons/registry/2.7.1/install.sh
```diff
@@ -211,3 +229,8 @@ function determine_registry_pvc_size() {
 
     export REGISTRY_PVC_SIZE=$registry_pvc_size
 }
+
+function registry_pvc_migrated() {
+    registry_pod=$( kubectl get pods -n kurl -l app=registry -o jsonpath='{.items[0].metadata.name}')
+    kubectl -n kurl logs $registry_pod -c migrate-s3  | grep "migration ran successfully" &>/dev/null
```
emosbaugh, line 235: `grep -q` rather than /dev/null will silence output

emosbaugh: can you just make sure that this function doesnt output a bunch of stuff to stderr say if the registry pod doesnt exist

emosbaugh (changes requested):

addons/registry/2.7.1/install.sh
jala-dx: yeah i used try_1m instead of try1_stderr so we dont see errors when the registry pod restarts. But it echo's if the final attempt fails.

addons/registry/2.7.1/install.sh
```diff
@@ -3,9 +3,14 @@ function registry() {
     regsitry_init_service   # need this again because kustomize folder is cleaned before install
     registry_session_secret
 
+    local will_migrate_pvc=
+    if ! registry_pvc_exists && object_store_exists && [ "$REGISTRY_USE_PVC_STORAGE" = "1" ]; then
+       will_migrate_pvc=1
```
DanStough, line 8: Would it make sense to just move this inside the `if` statement on Line 29?

addons/registry/2.7.1/patch-deployment-migrate-s3.yaml
```diff
@@ -0,0 +1,46 @@
...
+            secretKeyRef:
+              key: secret-access-key
+              name: registry-s3-secret
+        - name: OBJECT_STORE_CLUSTER_IP
+          valueFrom:
+            secretKeyRef:
+              key: object-store-cluster-ip
+              name: registry-s3-secret
+      volumes:
+      - name: registry-data
```
DanStough, line 40: I think this is redundant if it's already inserted into `deployment-pvc.yaml`.

DanStough (approved):

DanStough (approved):

DanStough:
@emosbaugh gave verbal approval of the changes

PR #2332, opened 2021-11-03
Download bundle image overrides concurrently

emosbaugh:
Pull and write image overrides to disk while concurrently streaming the rest of the bundle to limit the idle time when downloading a bundle.

### DIFF
--- a/cmd/server/main.go
+++ b/cmd/server/main.go
@@ -23,6 +23,7 @@ import (
 
 	"github.com/bugsnag/bugsnag-go/v2"
 	"github.com/containers/image/v5/copy"
+	"github.com/containers/image/v5/docker/reference"
 	"github.com/containers/image/v5/signature"
 	"github.com/containers/image/v5/transports/alltransports"
 	"github.com/containers/image/v5/types"
@@ -167,6 +168,8 @@ func bundle(w http.ResponseWriter, r *http.Request) {
 	w.Header().Set("Content-Disposition", "attachment")
 	w.Header().Set("Transfer-Encoding", "chunked")
 
+	// use handleError below here since headers have already been written
+
 	if r.Method == "HEAD" {
 		return
 	}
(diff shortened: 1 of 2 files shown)


cmd/server/main.go
```diff
@@ -187,96 +190,82 @@ func bundle(w http.ResponseWriter, r *http.Request) {
...
+	defer close(errCh)
+
+	imageCh := make(chan string, len(bundle.Images))
+
+	// HACK: Work around the fact that github.com/containers/image/v5/copy copy.Image does not support streaming.
+	// Download images concurrently while streaming the rest of the bundle to minimize idle time.
+	go func() {
+		defer close(imageCh)
+
+		for i, image := range bundle.Images {
```
syntastical, line 214: Based on the title of this PR, wouldn't you want this for loop outside the goroutine, to download concurrently?

emosbaugh: It will download them sequentially but concurrently while streaming the rest of the bundle

cmd/server/main.go
```diff
@@ -187,96 +190,82 @@ func bundle(w http.ResponseWriter, r *http.Request) {
...
+			}
+			_, err = copy.Image(r.Context(), policyContext, localRef, srcRef, &copy.Options{
+				RemoveSignatures:      true,
+				SignBy:                "",
+				ForceManifestMIMEType: "",
+				DestinationCtx:        destCtx,
+				SourceCtx:             srcCtx,
+			})
+
+			runtime.GC()
```
jala-dx, line 259: am not 100% sure, since[ runtime.GC() ](https://pkg.go.dev/runtime#GC)is going to block, can it be run outside the for loop only once?

jala-dx (approved):

cmd/server/main.go
emosbaugh: @divolgin can you help me understand why this is necessary? this code was lifted from kots i believe.

divolgin: to keep memory usage down.  because Go will keep requesting more memory from the system for each image pull, eventually the pod will be evicted

emosbaugh: so sounds like this should stay in the loop. thanks

divolgin (approved):

PR #2334, opened 2021-11-03
Produce invalid bundle archive on error

emosbaugh:
Extracting an incomplete bundle will now error like the following:

```
$ tar xzf 09361df.tar.gz 
tar: Unexpected EOF in archive
tar: Unexpected EOF in archive
tar: Error is not recoverable: exiting now
```


divolgin:
How about just a plain text file with the error?  And tar will still fail too.

divolgin (approved):

divolgin (approved):

PR #2337, opened 2021-11-03
update testgrid specs

kwsorensen:


addons/minio/template/testgrid/k8s-docker.yaml
```diff
@@ -22,7 +22,9 @@
       version: "latest"
     kotsadm:
       version: "latest"
-    rook:
```
emosbaugh, line 25: based on the comment i think you should leave rook here

```
# migration from rook-ceph object store to minio
```

emosbaugh (changes requested):

addons/minio/template/testgrid/k8s-docker.yaml
```diff
@@ -26,14 +26,14 @@
...
-      version: "latest"
+      version: "1.19.15"
     weave:
       version: "latest"
     containerd:
       version: "latest"
     kotsadm:
       version: "latest"
-    rook:
+    longhorn:
```
emosbaugh, line 36: i think you need to add back rook here

emosbaugh (changes requested):

addons/minio/template/testgrid/k8s-docker.yaml
emosbaugh: i am wrong, this should be longhorn

emosbaugh (approved):

PR #2342, opened 2021-11-04
Testgrid post install and upgrade scripts

emosbaugh:


jala-dx:
just have an ask:) if the post install/upgrade fails, is it possible to collect the support bundle and upload it somewhere?

jala-dx (approved):

PR #2366, opened 2021-11-10
Migrate Internal Snapshots from Object Store to PVC

DanStough:
If you install with Velero 1.6.2 from this branch, the new default will be to use the PVC-backed snapshots.

To Test Migration:
1. Install Rook Spec with flag enabled: `curl https://kurl.sh/4b91c41 | sudo bash`
2. Take snapshots through the kotsadm console
3. Run the upgrade locally using this manifest and the install command: `cat install.sh | sudo BETA_VELERO_USE_INTERNAL_PVC=true REGISTRY_USE_PVC_STORAGE=1 bash`
```
INSTALLER_YAML="apiVersion: cluster.kurl.sh/v1beta1
kind: Installer
metadata: 
  name: test
spec: 
  kubernetes: 
    version: 1.20.11
  weave: 
    version: 2.6.5
  containerd: 
    version: 1.4.6
  longhorn: 
    version: 1.1.2
  registry:
    version: 2.7.1
  velero:
    version: 1.6.2
  kotsadm:
    version: 1.57.0
    disableS3: true"
```
4. Go the kotsadm console and verify the snapshots are still there
5. Perform a full restore and observe the app and console are restored.

Additional PR coming to backport to other Velero versions and automated template.


addons/velero/1.6.2/install.sh
```diff
@@ -1,34 +1,63 @@
...
+        logSuccess "Velero migration complete"
+    fi
+
+    # Patch snapshots volumes to "Retain" in case of deletion
+    if kubernetes_resource_exists "$VELERO_NAMESPACE" pvc velero-internal-snapshots; then
+
+        local velero_pv_name
+        echo "Patching internal snapshot volume Reclaim Policy to RECLAIM"
+        try_1m velero_pvc_bound
+        velero_pv_name=$(kubectl get pvc velero-internal-snapshots -n ${VELERO_NAMESPACE} -ojsonpath='{.spec.volumeName}')
```
jala-dx, line 58: is the pvc name velero-internal-snapshots, or snapshot-pvc

addons/velero/1.6.2/install.sh
```diff
@@ -38,64 +67,100 @@ function velero_join() {
...
     local resticArg="--use-restic"
     if [ "$VELERO_DISABLE_RESTIC" = "1" ]; then
         resticArg=""
     fi
 
+    # detect if we need to use object store or pvc
     local bslArgs="--no-default-backup-location"
     if ! kubernetes_resource_exists "$VELERO_NAMESPACE" backupstoragelocation default; then
-        bslArgs="--provider aws --bucket $VELERO_LOCAL_BUCKET --backup-location-config region=us-east-1,s3Url=${OBJECT_STORE_CLUSTER_HOST},publicUrl=http://${OBJECT_STORE_CLUSTER_IP},s3ForcePathStyle=true"
+        bslArgs="--provider replicated.com/pvc --bucket velero-internal-snapshots --backup-location-config storageSize=${VELERO_PVC_SIZE},resticRepoPrefix=/var/velero-local-volume-provider/velero-internal-snapshots/restic"
```
jala-dx, line 84: don't we need to add this logic only if disable_s3 flag is enabled ?

addons/velero/1.6.2/install.sh
DanStough: Should be `velero-internal-snapshots`, or at least that the name I picked. I can change it if there was something different you used.

addons/velero/1.6.2/install.sh
```diff
@@ -38,64 +67,100 @@ function velero_join() {
 function velero_install() {
     local src="$1"
     local dst="$2"
+    local has_secret
```
emosbaugh, line 70: this isnt needed

addons/velero/1.6.2/install.sh
```diff
@@ -38,64 +67,100 @@ function velero_join() {
...
     if ! kubernetes_resource_exists "$VELERO_NAMESPACE" backupstoragelocation default; then
-        bslArgs="--provider aws --bucket $VELERO_LOCAL_BUCKET --backup-location-config region=us-east-1,s3Url=${OBJECT_STORE_CLUSTER_HOST},publicUrl=http://${OBJECT_STORE_CLUSTER_IP},s3ForcePathStyle=true"
+        bslArgs="--provider replicated.com/pvc --bucket velero-internal-snapshots --backup-location-config storageSize=${VELERO_PVC_SIZE},resticRepoPrefix=/var/velero-local-volume-provider/velero-internal-snapshots/restic"
     fi
 
-    velero_credentials
+    # we only need a secret file if it's already set for some other provider (including legacy internal storage)
+    local secretArgs="--no-secret"
+    if kubernetes_resource_exists "$VELERO_NAMESPACE" secret cloud-credentials; then
+        has_secret=true
```
emosbaugh, line 90: this isnt needed

addons/velero/1.6.2/install.sh
```diff
@@ -162,13 +227,142 @@ function velero_patch_http_proxy() {
...
+        VELERO_S3_ACCESS_KEY_SECRET=$(kubectl -n rook-ceph get secret rook-ceph-object-user-rook-ceph-store-kurl -o yaml | grep SecretKey | head -1 | awk '{print $2}' | base64 --decode)
+    else 
+        echo "Previous installation of Minio detected."
+        VELERO_S3_HOST="minio.minio"
+        VELERO_S3_ACCESS_KEY_ID=$(kubectl -n minio get secret minio-credentials -ojsonpath='{ .data.MINIO_ACCESS_KEY }' | base64 --decode)
+        VELERO_S3_ACCESS_KEY_SECRET=$(kubectl -n minio get secret minio-credentials -ojsonpath='{ .data.MINIO_SECRET_KEY }' | base64 --decode)
+    fi
+
+    # If this is run through `velero_already_applied`, we need to create base kustomization file
+    if [ ! -f "$dst/kustomization.yaml" ];then
```
emosbaugh, line 290: how about moving this to velero_already_applied?

addons/velero/1.6.2/install.sh
```diff
@@ -162,13 +227,142 @@ function velero_patch_http_proxy() {
...
+
+    # add patch to add init container for migration
+    render_yaml_file "$src/tmpl-s3-migration-deployment-patch.yaml" > "$dst/s3-migration-deployment-patch.yaml"
+    insert_patches_strategic_merge "$dst/kustomization.yaml" s3-migration-deployment-patch.yaml
+
+    # update the BackupstorageLocation
+    render_yaml_file "$src/tmpl-s3-migration-bsl.yaml" > "$dst/s3-migration-bsl.yaml"
+    insert_resources "$dst/kustomization.yaml" s3-migration-bsl.yaml
+
+    export WILL_MIGRATE_VELERO_OBJECT_STORE="1"
```
emosbaugh, line 312: what about refactoring this function into two separate functions with one `velero_should_migrate_from_object_store()`?

addons/velero/1.6.2/install.sh
```diff
@@ -38,64 +67,100 @@ function velero_join() {
...
+    velero_change_storageclass "$src" "$dst"
+
+    # In the case this is a rook re-apply, no changes might be required
+    if [ -f "$dst/kustomization.yaml" ]; then
+        kubectl apply -k "$dst"
     fi
 
-    if [ -n "$OBJECT_STORE_CLUSTER_IP" ]; then
-        try_1m object_store_create_bucket "$VELERO_LOCAL_BUCKET"
+    # Bail if the migratoin fails, preventing the original object store from being deleted
```
emosbaugh, line 139: migration spelling

addons/velero/1.6.2/install.sh
```diff
@@ -162,13 +227,142 @@ function velero_patch_http_proxy() {
...
+    insert_patches_strategic_merge "$dst/kustomization.yaml" internal-snaps-ds-patch.yaml
+
+}
+
+function velero_pvc_bound() {
+    kubectl get pvc velero-internal-snapshots -n ${VELERO_NAMESPACE} -oyaml | grep -q "phase: Bound"
+}
+
+function velero_pvc_exists() {
+    kubectl -n "${VELERO_NAMESPACE}" get pvc velero-internal-snapshots &>/dev/null
```
emosbaugh, line 346: is there some reason not to use kubernetes_resource_exists here?

addons/velero/1.6.2/install.sh
```diff
@@ -162,13 +227,142 @@ function velero_patch_http_proxy() {
...
+}
+
+# if the PVC size has already been set we should not reduce it
+function determine_velero_pvc_size() {
+    local velero_pvc_size="50Gi"
+    if velero_pvc_exists; then
```
emosbaugh, line 352: does this needs its own function if you use kubernetes_resource_exists?

addons/velero/1.6.2/install.sh
```diff
@@ -162,13 +227,142 @@ function velero_patch_http_proxy() {
...
+    insert_patches_strategic_merge "$dst/kustomization.yaml" internal-snaps-deployment-patch.yaml
+
+    # add patch to add the pvc in the correct location for the restic daemonset
+    render_yaml_file "$src/tmpl-internal-snaps-ds-patch.yaml" > "$dst/internal-snaps-ds-patch.yaml"
+    insert_patches_strategic_merge "$dst/kustomization.yaml" internal-snaps-ds-patch.yaml
+
+}
+
+function velero_pvc_bound() {
+    kubectl get pvc velero-internal-snapshots -n ${VELERO_NAMESPACE} -oyaml | grep -q "phase: Bound"
```
emosbaugh, line 342: better to use templates here?

addons/velero/1.6.2/tmpl-internal-snaps-deployment-patch.yaml
```diff
@@ -0,0 +1,26 @@
+---
```
emosbaugh, line 1: is this a template?

addons/velero/1.6.2/tmpl-internal-snaps-ds-patch.yaml
```diff
@@ -0,0 +1,17 @@
+---
```
emosbaugh, line 1: is this a template?

addons/velero/1.6.2/tmpl-s3-migration-configmap.yaml
```diff
@@ -0,0 +1,33 @@
...
+kind: ConfigMap
+metadata:
+  name: velero-migrate-s3-config
+  labels:
+    app: velero
+data:
+  migrate-s3.sh: |-
+    #!/bin/sh
+    set -euo pipefail
+    export S3_HOST=\$OBJECT_STORE_CLUSTER_IP
```
emosbaugh, line 12: arent these in environment variables set in the kubernetes resource env property using the secret? wont running this through render_yaml_file strip all these vars? also expose the secrets in the config map?

emosbaugh (changes requested):

addons/velero/1.6.2/tmpl-s3-migration-configmap.yaml
emosbaugh: this is really the only blocker for me. everything else is more stylistic changes

addons/velero/1.6.2/install.sh
```diff
@@ -1,34 +1,63 @@
...
 function velero_pre_init() {
     if [ -z "$VELERO_NAMESPACE" ]; then
         VELERO_NAMESPACE=velero
     fi
     if [ -z "$VELERO_LOCAL_BUCKET" ]; then
         VELERO_LOCAL_BUCKET=velero
     fi
+    # TODO (dans): make this configurable from the installer spec
+    # if [ -z "$VELERO_REQUESTED_CLAIM_SIZE" ]; then
+    #     VELERO_REQUESTED_CLAIM_SIZE="50Gi"
```
jala-dx, line 11: we need to document that if the existing snapshot size is more than 50Gi, it has to be specified in the kurl spec.

addons/velero/1.6.2/install.sh
```diff
@@ -1,34 +1,63 @@
...
     velero_change_storageclass "$src" "$dst"
 
+    velero_migrate_from_object_store "$src" "$dst"
+
     kubectl apply -k "$dst"
 
     kubectl label -n default --overwrite service/kubernetes velero.io/exclude-from-backup=true
+
+    # Bail if the migratoin fails, preventing the original object store from being deleted
+    if [ "$WILL_MIGRATE_VELERO_OBJECT_STORE" = "1" ]; then
```
jala-dx, line 46:
```suggestion
    if [ "$BETA_MIGRATE_VELERO_OBJECT_STORE" = "1" ]; then
```

addons/velero/1.6.2/install.sh
jala-dx, line 84: this BSL can be changed only when disableS3 flag is specified in the kurl spec?

addons/velero/1.6.2/install.sh
```diff
@@ -38,64 +67,100 @@ function velero_join() {
...
 
-    # This should only be applying the configmap if required
-    if compgen -G "$dst/*.yaml" > /dev/null; then
-        kubectl apply -f "$dst"
+    # This function will copy the kustomization template from the file if we need a migration.
+    velero_migrate_from_object_store "$src" "$dst"
+
+    # If we need to migrate, we're going to need to basically reconstruct the original install 
+    # underneath the migration
+    if [ "$WILL_MIGRATE_VELERO_OBJECT_STORE" = "1" ]; then
```
jala-dx, line 121:
```suggestion
    if [ "$BETA_MIGRATE_VELERO_OBJECT_STORE" = "1" ]; then
```

addons/velero/1.6.2/install.sh
DanStough: I was using `velero-internal-snapshots` but I'm open to changing. I do think "internal" should be part of the name thought to describe where it is used.

addons/velero/1.6.2/install.sh
DanStough: Nope, I think it just slipped my mind.

addons/velero/1.6.2/install.sh
DanStough: nope, doesn't need another function.

addons/velero/1.6.2/install.sh
DanStough: How are you proposing to use templates in this case? The problem in my mind was that I can't patch the PV until it has been created for the PVC, so this is a hack to know whether there is a corresponding PVC/PV before I try to grab the PV name.

addons/velero/1.6.2/tmpl-s3-migration-configmap.yaml
DanStough: I don't think they get stripped becasue the notation for templates is curly braces, but I agree this doesn't need to be a template. I think I copied from `registry` and it's like that there. Will fix.

addons/velero/1.6.2/install.sh
emosbaugh: I meant the flag to the kubectl comment to extract the phase.

addons/velero/1.6.2/install.sh
DanStough: No, I think this condition basically targets only new installs. If there is any existing BSL, like one created through KOTS, then we pass `--no-default-backup-location` and velero won't touch the existing one. IICR from discussions with @ajp-io during planning, we want to move over all new installs of kurl to the new internal storage mechanism immediately.

addons/velero/1.6.2/install.sh
DanStough: Nope, we only need to run this if we are migrating. That variable is checked in `velero_migrate_from_object_store`.

addons/velero/1.6.2/install.sh
DanStough: ah, ok. I can do that.

addons/velero/1.6.2/tmpl-internal-snaps-ds-patch.yaml
DanStough: nope! good catch.

addons/velero/1.6.2/tmpl-internal-snaps-deployment-patch.yaml
DanStough: nope!

addons/velero/1.6.2/install.sh
DanStough: I think the problem I'm running into is that in velero_already_applied I don't really even know if I'll need this or if it already exists. There are three states - no kustmoization necessary, just the configmap to restore from rook to longhorn, or migration + configmap. Open to refactoring but seems like I would have to create a lot more code to check all the conditions first before creating a kustomize file.

addons/velero/1.6.2/install.sh
ajp-io: We've changed some of our decisions since then, so now I wonder what is best. If we make all new installs use PVCs for everything, we will end up with the split architecture that Marc mentioned where customers could have end-customers deployed differently depending on whether they initially deployed before or after this change. He wasn't a fan of that. Makes me wonder if we should just respect the `disableS3` flag (or whichever new flag we might use) instead, even for new installs. WDYT, Dan?

addons/velero/1.6.2/install.sh
DanStough: Sure, pushing some changes now. Let me know if it's what you're thinking.

addons/velero/1.6.2/install.sh
DanStough: Nope, this is already checked above in `velero_migrate_from_object_store`.

I refactored a bit anyway for clarity.

addons/velero/1.6.2/install.sh
DanStough: Yeah, I guess it makes sense not to change the existing behavior unless you add a flag. Do we want to gate this with a feature flag for a while too while we backport to other velero versions and maybe figure out if there is enough storage in the cluster? I can change the environment variable `BETA_VELERO_MIGRATE_FROM_OBJECT_STORE` to something more generic like `BETA_VELERO_USE_INTERNAL_PVC `

addons/velero/1.6.2/install.sh
jala-dx: let us keep the way you wrote it. I can modify my code.

emosbaugh (approved):

emosbaugh (approved):

jala-dx (approved):

PR #2374, opened 2021-11-19
Debug output for testgrid airgap failures

emosbaugh:
There are a lot of testgrid airgap failures downloading the bundle. I'm trying to surface a little more info.

```
+ echo 'downloading install bundle'
downloading install bundle
+ curl -fsSL -o install.tar.gz https://staging.kurl.sh/bundle/867b283.tar.gz
+ '[' -n '' ']'
+ disable_internet
...
+ tar -xzf install.tar.gz

gzip: stdin: unexpected end of file
tar: Unexpected EOF in archive
tar: Unexpected EOF in archive
tar: Error is not recoverable: exiting now
+ local tar_exit_status=2
+ '[' 2 -ne 0 ']'
+ echo 'failed to unpack airgap file with status 2'
failed to unpack airgap file with status 2
```

### DIFF
--- a/testgrid/tgrun/pkg/runner/embed/runcmd.sh
+++ b/testgrid/tgrun/pkg/runner/embed/runcmd.sh
@@ -32,16 +32,16 @@ function run_install() {
 
         echo "downloading install bundle"
 
-        curl -fsSL -o install.tar.gz "$KURL_URL"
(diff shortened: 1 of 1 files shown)


laverya (approved):
I'm relatively sure this is our airgap builder restarting while the test is downloading

PR #2384, opened 2021-11-24
Pin to minor versions instead of latest

marccampbell:
This needs to be verified on vendor web staging after merge

### DIFF
--- a/web/src/installers/index.ts
+++ b/web/src/installers/index.ts
@@ -3,7 +3,7 @@ import * as yaml from "js-yaml";
 import * as _ from "lodash";
 import * as mysql from "promise-mysql";
 import { Service } from "@tsed/common";
-import Ajv, {KeywordCxt} from "ajv";
+import Ajv from "ajv";
 import * as semver from "semver";
 import { MysqlWrapper } from "../util/services/mysql";
 import { instrumented } from "monkit";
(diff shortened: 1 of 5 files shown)


web/src/installers/index.ts
```diff
@@ -762,15 +762,15 @@ export class Installer {
...
-    i.spec.ekco = { version: "latest" };
-    i.spec.contour = { version: "latest" };
-    i.spec.registry = { version: "latest" };
-    i.spec.prometheus = { version: "latest" };
+    i.spec.kubernetes = { version: "1.21.x" };
+    i.spec.containerd = { version: "1.4.x" };
+    i.spec.weave = { version: "2.6.x" };
+    i.spec.longhorn = { version: "1.1.x"};
+    i.spec.minio = { version: "2020-01-25T02-50-51Z"};
+    i.spec.ekco = { version: "0.12.x" };
```
emosbaugh, line 770: should this be latest? ekco version updates usually have important fixes for better operation of the cluster.

DanStough (approved):

web/src/installers/index.ts
marccampbell: @emosbaugh sure. it's weird to be the only one...  we don't support 0.x notation yet?  what options can we use to allow us to introduce a breaking change in the future?

emosbaugh: pinning to major notation eg 0.x would be great. additionally there have been discussions about making ekco built into kurl and no longer an addon.

emosbaugh, line 770:
```suggestion
    i.spec.ekco = { version: "latest" };
```

emosbaugh (changes requested):

emosbaugh:
[sc-38729]

emosbaugh (approved):

web/src/installers/index.ts
```diff
@@ -762,15 +762,15 @@ export class Installer {
...
+    i.spec.kubernetes = { version: "1.21.x" };
+    i.spec.containerd = { version: "1.4.x" };
+    i.spec.weave = { version: "2.6.x" };
+    i.spec.longhorn = { version: "1.1.x"};
+    i.spec.minio = { version: "2020-01-25T02-50-51Z"};
     i.spec.ekco = { version: "latest" };
-    i.spec.contour = { version: "latest" };
-    i.spec.registry = { version: "latest" };
-    i.spec.prometheus = { version: "latest" };
+    i.spec.contour = { version: "1.18.x" };
```
DanStough, line 771: contour is 1.19 now.

web/src/installers/index.ts
```diff
@@ -762,15 +762,15 @@ export class Installer {
...
     i.id = "latest";
-    i.spec.kubernetes = { version: "latest" };
-    i.spec.containerd = { version: "latest" };
-    i.spec.weave = { version: "latest" };
-    i.spec.longhorn = { version: "latest"};
-    i.spec.minio = { version: "latest"};
+    i.spec.kubernetes = { version: "1.21.x" };
+    i.spec.containerd = { version: "1.4.x" };
+    i.spec.weave = { version: "2.6.x" };
+    i.spec.longhorn = { version: "1.1.x"};
```
DanStough, line 768: Longhorn is 1.2

DanStough (approved):

DanStough (dismissed):

PR #2392, opened 2021-11-29
Unify project into one go module

emosbaugh:
Dependabot go prs are failing because there is a reference from testgrid go module back to the root and they do not update both deps causing conflicts. This pr unifies the project into one module.


testgrid/tgapi/deploy/Dockerfile
```diff
@@ -1,6 +1,9 @@
 FROM golang:1.14
```
laverya, line 1: this should probably match the go version in the go.mod

laverya (approved):

DanStough (approved):

PR #2412, opened 2021-12-02
Install prerequisite packages for Longhorn on all nodes

divolgin:
Use tasks.sh script to install host dependencies for Longhorn after the daemon set is created

### DIFF
--- a/addons/longhorn/1.1.2/install.sh
+++ b/addons/longhorn/1.1.2/install.sh
@@ -30,7 +30,7 @@ function longhorn() {
 
     check_mount_propagation $src
 
-    longhorn_host_init
+    longhorn_host_init "$DIR/addons/longhorn/$LONGHORN_VERSION"
 
 
     render_yaml_file "$src/tmpl-ui-service.yaml" > "$dst/ui-service.yaml"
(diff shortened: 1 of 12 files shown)


addons/longhorn/1.1.2/install.sh
```diff
@@ -190,6 +138,48 @@ function validate_longhorn_ds() {
     fi
 }
 
+# if this is a multinode cluster, we need to prepare all hosts to run the daemonset
+function maybe_init_hosts() {
```
laverya, line 142: I would assume that this function needs to be added to the template too

scripts/common/longhorn.sh
```diff
@@ -0,0 +1,54 @@
+function longhorn_host_init() {
```
laverya, line 1: I'm not a huge fan of moving this to common instead of leaving it in the addon - why are we doing that?

laverya (dismissed):

addons/longhorn/1.1.2/install.sh
divolgin: added

scripts/common/longhorn.sh
divolgin: It is now called from tasks.sh

laverya (approved):

laverya (approved):

PR #2416, opened 2021-12-02
fix(velero): order of resources during re-apply

DanStough:


jala-dx (dismissed):

jala-dx:
still in draft, but lgtm.

jala-dx (approved):

PR #2423, opened 2021-12-03
feat: add OL 8.5 support

DanStough:
VM Images from https://yum.oracle.com/oracle-linux-templates.html


laverya (approved):

testgrid/tgrun/hack/os-spec.yaml
```diff
@@ -0,0 +1,71 @@
+- id: amzn-20
```
laverya, line 1: did you mean to comment the two 'hack' files?

DanStough: Yeah, these are referenced [here](https://stackoverflow.com/c/replicated/questions/339) for testing, so I thought it pertinent to refresh them.

DanStough:
Giving up on getting the qcow2 images to work for now.

PR #2425, opened 2021-12-03
IPv6

areed:
- Add ipv6 flag and field to kurl spec
- Detect ipv6 ip and use as private address
- Update curl commands to work with ipv6 ips
- Configure host sysctls for IPv6
- Configure latest antrea and rook addon versions to work with IPv6
- Local registry uses cluster DNS name on host since images tagged with IPv6 ip are an invalid reference
- Make skaffold work without vandoor db


addons/registry/2.7.1/tmpl-deployment-objectstore.yaml
```diff
@@ -36,7 +36,7 @@ data:
         disable: true
       s3:
         region: "us-east-1"
-        regionendpoint: http://$OBJECT_STORE_CLUSTER_IP
+        regionendpoint: http://$OBJECT_STORE_CLUSTER_IP_BRACKETED
```
DanStough, line 39: Won't this break now with Minio or older versions of Rook?

DanStough (changes requested):
My only dealbreaker is the comment in the Registry Addon but I'm probably wrong.

addons/registry/2.7.1/install.sh
```diff
@@ -155,8 +155,13 @@ function registry_cred_secrets() {
     kubectl -n kurl patch secret registry-htpasswd -p '{"metadata":{"labels":{"kots.io/kotsadm":"true", "kots.io/backup":"velero"}}}'
     rm htpasswd
 
+    local server="$DOCKER_REGISTRY_IP"
+    if [ "$IPV6_ONLY" = "1" ]; then
+        server="registry.kurl.svc.cluster.local"
```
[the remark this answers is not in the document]
areed: I didn't want to risk breaking existing installs or docker. With ipv6 only new installs with containerd will be affected.

addons/registry/2.7.1/tmpl-deployment-objectstore.yaml
areed: yeah, good catch

DanStough (approved):
Awesome!

areed:
Thanks @DanStough , could you merge?

emosbaugh:
docs need to be written for kurl.sh

emosbaugh:
[sc-30387]

PR #2445, opened 2021-12-07
Adding option to specify labels

jdewinne:
SC: 22534

Example: `curl https://kurl.sh/latest | sudo bash -s labels=appels=banana,orange=grape`

Above will add the labels `appels` and `orange` to the node with the corresponding values.
This also works when adding nodes.

Once approved, I can also add the corresponding docs.


marccampbell:
@jdewinne Great addition! We'll have to figure out how to make sure these are validated in testgrid before we can move forward or there will be a risk of regressions in the future.  Shouldn't be too hard of a change, but let's find some time to get this tested.

jdewinne:
@marccampbell are there instructions on how to get it into "testgrid"?

DanStough (approved):

trelore:
Merging this now to test on staging, will add and merge the testgrid tests after I've manually verified it

PR #2465, opened 2021-12-09
fix: modprode ol 8.5

DanStough:


laverya (approved):
fixing it once and for all, nice

PR #2472, opened 2021-12-10
fix(tg): broken web routing

DanStough:


testgrid/web/package.json
```diff
@@ -69,7 +68,7 @@
     "react-helmet": "^6.1.0",
     "react-modal": "^3.14.4",
     "react-monaco-editor": "^0.44.0",
-    "react-router-dom": "^6.0.2",
+    "react-router-dom": "^5.2.0",
```
DanStough, line 71: I went back because rolling forward breaks the properties the `Run` component is expecting, and the only fix option provided was using hooks.

testgrid/web/.babelrc
```diff
@@ -11,6 +11,5 @@
     ["@babel/plugin-proposal-decorators", { "legacy": true }],
     "@babel/plugin-proposal-object-rest-spread",
     "@babel/plugin-proposal-class-properties",
-    "react-hot-loader/babel",
```
DanStough, line 14: I removed this because it throws a warning in the browser, and I'm not sure it's working anyway. It's currently deprecated, but I couldn't get the replacement to work in a reasonable amount of time.

emosbaugh (approved):

PR #2498, opened 2021-12-14
IPv6 expanded support

areed:
* Fix IPv6 on CentOS 8, RHEL 8, and Ubuntu 20.04
* Fix antrea serviceCIDRv6 for new 1.4.0 configmap
* Load ip6_tables modules on all nodes with antrea and persist after reboot (also fixed for IPv4)
* Antrea encryption ipv6 with wireguard. Antrea encryption with IPv4 uses IPSec, but [that's not known to work with antrea with IPv6](https://kubernetes.slack.com/archives/CR2J23M0X/p1639614609127500). The installer will bail if wireguard is not found.
* Fix snapshots/restores with IPv6 by using object store cluster hostname from scripts since the s3cmd fails when using an IPv6 IP as the host.

### DIFF
--- a/addons/antrea/1.4.0/install.sh
+++ b/addons/antrea/1.4.0/install.sh
@@ -23,9 +23,11 @@ function antrea() {
 
     if ! lsmod | grep ip_tables; then
         modprobe ip_tables
+        echo 'ip_tables' > /etc/modules-load.d/kurl-antrea.conf
     fi
     if [ "$IPV6_ONLY" = "1" ]; then
         modprobe ip6_tables
+        echo 'ip6_tables' > /etc/modules-load.d/kurl-antrea.conf
     fi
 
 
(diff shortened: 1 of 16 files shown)


scripts/Manifest
```diff
@@ -9,4 +9,28 @@ KURL_UTIL_IMAGE=replicated/kurl-util:alpha
 KURL_BIN_UTILS_FILE=
 # STEP_VERSIONS array is generated by the server and injected at runtime based on supported k8s versions
 STEP_VERSIONS=("" "" "" "" "" "" "" "" "" "" "" "" "" "" "" "1.15.3" "1.16.4" "1.17.13" "1.18.20" "1.19.15", "1.20.11")
-INSTALLER_YAML=
+INSTALLER_YAML="
```
emosbaugh, line 12: Can you remove this?

emosbaugh (commented):
Is this pr ready for review?

areed:
> Is this pr ready for review?

Sorry, I'll make it a draft

emosbaugh (approved):

PR #2530, opened 2021-12-21
chore(actions): k8s cron get only latest supported versions

DanStough:


packages/kubernetes/template/script.sh
```diff
@@ -2,25 +2,31 @@
...
 set -euo pipefail
 
-# Populate VERSIONS array with 1.20+ and 1.19 and 1.18 latest versions available
+# Populate VERSIONS array latest kURL-support versions (1.19, 1.20, 1.21) available
 VERSIONS=()
 function find_available_versions() {
     docker build -t k8s - < Dockerfile
-    VERSIONS=($(docker run k8s apt list -a kubelet 2>/dev/null | grep -Eo '1\.[2-9][0-9]\.[0-9]+' | sort -rV | uniq))
-    echo "Found ${#VERSIONS[*]} versions for Kubernetes 1.20+: ${VERSIONS[*]}"
+    # VERSIONS=($(docker run k8s apt list -a kubelet 2>/dev/null | grep -Eo '1\.[2][0-9]\.[0-9]+' | sort -rV | uniq))
```
jala-dx, line 9: this restricts to only 1.2x  We can keep it as is and pull in the needed versions only.

packages/kubernetes/template/script.sh
```diff
@@ -2,25 +2,31 @@
...
+    # echo "Found ${#VERSIONS[*]} versions for Kubernetes 1.20+: ${VERSIONS[*]}"
+
+    local versions121=($(docker run k8s apt list -a kubelet 2>/dev/null | grep -Eo '1\.21\.[0-9]+' | sort -rV | uniq))
```
jala-dx, line 12: In here we are restricting it to 1.21, it can be **'1\.[2-9][0-1]\.[0-9]+'**

packages/kubernetes/template/script.sh
```diff
@@ -2,25 +2,31 @@
...
+    if [ ${#versions121[@]} -gt 0 ]; then
+        echo "Found latest version for Kubernetes 1.21: ${versions121[0]}"
+        VERSIONS+=("${versions121[0]}")
+    fi
+
+    local versions120=($(docker run k8s apt list -a kubelet 2>/dev/null | grep -Eo '1\.20\.[0-9]+' | sort -rV | uniq))
```
jala-dx, line 18: In here, it is restricted to 1.20. We need to keep this consistent.

jala-dx (approved):

PR #2565, opened 2022-01-05
update TestGrid test cases

pavansokkenagaraj:
SC: 40502

- Update TestGrid test cases from k8s 1.18.x to 1.20.x and 1.21.x

### DIFF
--- a/testgrid/specs/full.yaml
+++ b/testgrid/specs/full.yaml
@@ -142,6 +142,8 @@
       version: 2.6.5
     contour:
       version: 1.0.1
+    longhorn:
+      version: 1.2.2
     registry:
       version: 2.7.1
     prometheus:
@@ -150,8 +152,10 @@
       version: latest
     velero:
       version: 1.2.0
+    minio: 
+      version: latest
     ekco:
-      version: 0.6.0
+      version: latest
 - name: k8s121x_containerd
   installerSpec:
     kubernetes:
@@ -162,6 +166,8 @@
       version: 2.6.5
     contour:
       version: 1.0.1
+    longhorn:
+      version: 1.2.2
     registry:
       version: 2.7.1
     prometheus:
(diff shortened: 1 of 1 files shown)


testgrid/specs/full.yaml
```diff
@@ -132,18 +132,16 @@
...
-      version: 1.18.x
+      version: 1.20.x
     docker:
       version: 19.03.10
     weave:
       version: 2.6.5
     contour:
       version: 1.0.1
-    rook:
-      version: 1.0.4
```
DanStough, line 146: Keep this - `kotsadm` will need a CSI provider to allocate storage.

testgrid/specs/full.yaml
```diff
@@ -154,18 +152,16 @@
       version: 1.2.0
     ekco:
       version: 0.6.0
```
DanStough, line 154: Optional, I would just use `latest` here for EKCO.

testgrid/specs/full.yaml
```diff
@@ -154,18 +152,16 @@
...
-      version: 19.03.10
+      version: 1.21.x
+    containerd:
+      version: 1.4.6
     weave:
       version: 2.6.5
     contour:
       version: 1.0.1
-    rook:
-      version: 1.0.4
```
DanStough, line 168: Correct that Rook 1.0.4 doesn't work with k8s 1.21, but you still need a CSI provisioner. Either upgrade the version of rook to `latest` or switch to `longhorn`. I trust your decision on which one needs more coverage.

DanStough (changes requested):

testgrid/specs/full.yaml
pavansokkenagaraj: `rook` latest version is `1.0.4` . Switching to `longhorn`

testgrid/specs/full.yaml
pavansokkenagaraj: Can `kotsadm` use minio as storage provider?

pavansokkenagaraj: Switching to `minio`

DanStough (approved):

PR #2614, opened 2022-01-21
feat(host preflights): blocking host preflights and wording changes

DanStough:
#### What type of PR is this?
type::feature

Fixes:
* https://app.shortcut.com/replicated/story/38386/refer-to-preflights-as-host-preflights
* https://app.shortcut.com/replicated/story/41681/the-preflight-ignore-warnings-flag-should-default-to-true-and-this-default-should-be-documented
* https://app.shortcut.com/replicated/story/37569/make-host-preflights-blocking-once-again

```release-note
- [Host preflights](https://kurl.sh/docs/create-installer/host-preflights/) are now blocking. Failures will cause the installer to exit with error. Warnings will still result in the installer continuing. `host-preflight-ignore` and `host-preflight-enforce-warnings` are available to change the default behavior.
```

Yes, TBD.


kurl_util/cmd/bashmerge/main.go
```diff
@@ -167,16 +167,16 @@ func parseBashFlags(installer *kurlv1beta1.Installer, bashFlags string) error {
 				installer.Spec.Kurl = &kurlv1beta1.Kurl{}
 			}
 			installer.Spec.Kurl.IgnoreRemoteUpgradePrompt = true
-		case "preflight-ignore":
```
divolgin, line 170: These flags are publicly documented.  We should keep them for backwards compatibility.

divolgin (dismissed):
you could just do

```
case "host-preflight-ignore", "preflight-ignore":
```

divolgin (approved):

PR #2652, opened 2022-01-31
allow restoring older snapshots of the registry addon onto newly created instances

laverya:
With the addition of ipv6 support, we changed from "object store cluster-ip" to "object store hostname" - and changed this name in secrets/configmaps. This made restoring a snapshot of the registry addon that expects "object-store-cluster-ip" impossible, because the deployment manifests expected a secret field to exist that no longer did.

To fix that, this PR reincludes the object-store-cluster-ip field, but with the same data as object-store-hostname. It also includes logic within the velero restore script to accept using either the OBJECT_STORE_HOSTNAME environment variable OR the previous OBJECT_STORE_CLUSTER_IP env var, as the script will be run with the environment variables present at time of snapshot creation.

Fixes #

```release-note
Fix restoring registry on a new cluster when [migrating](https://kurl.sh/docs/install-with-kurl/migrating).
```

NONE

### DIFF
--- a/addons/registry/2.7.1/tmpl-configmap-velero.yaml
+++ b/addons/registry/2.7.1/tmpl-configmap-velero.yaml
@@ -21,6 +21,11 @@ data:
   restore.sh: |-
     #!/bin/sh
     set -euo pipefail
+
+    if [ -z "\$OBJECT_STORE_HOSTNAME" ]; then
+      export OBJECT_STORE_HOSTNAME=\$OBJECT_STORE_CLUSTER_IP # included for backwards compatibility for snapshot restores with pre-ipv6 snapshot restores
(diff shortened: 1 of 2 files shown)


DanStough (approved):

MikaelSmith:
Ran into a version so old it still inlined the IP address directly in the restore script. Happy for changes to make restores a little more flexible though.

laverya:
Oh? Do you know what version of kURL that was from?

MikaelSmith:
Approximately April 2021. It had 58a51ca002c5493177c3ba5803ca21c982ebd797 but not 0d4a7aa593adfc69c1ba5d88ae819a75f71ce9a4

PR #2654, opened 2022-02-01
kubernetes upgrade fails when enabling EKCO with internal load balancer.

pavanreddymaley:
#### What type of PR is this?
When we try to upgrade both kubernetes and EKCO versions in the same spec along with enabling internal load balancer feature the upgrade fails as the install.sh script will try check the api server endpoint of the internal load balancer (localhost:6444) which has not been deployed yet. 

With this change we fetch the api server endpoint from the cluster and performing status checks on it during the upgrade.  So the upgrade of kubernetes doesn't fail.

Fixes # https://github.com/replicated-collab/forwardnetworks-kots/issues/57

```None```

```NONE```

### DIFF
--- a/scripts/common/kubernetes.sh
+++ b/scripts/common/kubernetes.sh
@@ -252,18 +252,30 @@ function kubernetes_has_remotes() {
     return 1
 }
 
+# Fetch the load balancer endpoint from the cluster.
+function existing_kubernetes_api_address() {
+    kubectl get cm -n kube-system kurl-config -o jsonpath='{ .data.kubernetes_api_address }'
+}
+
+# During the upgrade user might change the load balancer endpoint or want to use EKCO internal load balancer. So, we
+# to be checking the api endpoint status on the existing api server endpoint as the new endpoint is only available after
+# finishing the upgrade.
 function kubernetes_api_address() {
-    local addr="$LOAD_BALANCER_ADDRESS"
-    local port="$LOAD_BALANCER_PORT"
+    if [ -n "$upgrading_kubernetes" ]; then
+        existing_kubernetes_api_address
+    else
+        local addr="$LOAD_BALANCER_ADDRESS"
+        local port="$LOAD_BALANCER_PORT"
 
-    if [ -z "$addr" ]; then
-        addr="$PRIVATE_ADDRESS"
-        port="6443"
+        if [ -z "$addr" ]; then
+            addr="$PRIVATE_ADDRESS"
+            port="6443"
+        fi
+
+        addr=$(${DIR}/bin/kurl format-address ${addr})
+
+        echo "${addr}:${port}"
     fi
-
-    addr=$(${DIR}/bin/kurl format-address ${addr})
-
-    echo "${addr}:${port}"
 }
 
 function kubernetes_api_is_healthy() {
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by jala-dx** [comment]
@pavanreddymaley thank you for the PR. Would you also write down the steps to reproduce this. I can make sure it gets tested and merged.

File: scripts/common/kubernetes.sh [hunk 1]
```diff
@@ -252,18 +252,30 @@ function kubernetes_has_remotes() {
...
+}
+
+# During the upgrade user might change the load balancer endpoint or want to use EKCO internal load balancer. So, we
+# to be checking the api endpoint status on the existing api server endpoint as the new endpoint is only available after
+# finishing the upgrade.
 function kubernetes_api_address() {
-    local addr="$LOAD_BALANCER_ADDRESS"
-    local port="$LOAD_BALANCER_PORT"
+    if [ -n "$upgrading_kubernetes" ]; then
+        existing_kubernetes_api_address
```
jala-dx, line 265: what is the purpose of calling existing_kubernetes_api_address function here ?

pavanreddymaley:
The main use case is related to upgrading the cluster from non-ha to HA using EKCO load balancer.

Steps to re-produce: 

1. Deploy a 3 node non-HA cluster with older version of k8s (ex: 1.19,1.20).
2. Upgrade the k8s cluster along with EKCO enabling internal load balancer.
Details of spec used for testing: https://docs.google.com/document/d/1O4jafOiQXOB0qH4UmXdY8ApkI6YfdE8XLT2enTIf7fU/edit


Regarding : existing_kubernetes_api_address

During the kubernetes upgrade we will verify if the kubernetes api server is healthy and in the above scenario we are enabling the EKCO internal load balancer and so the function kubernetes_api_address returns localhost:6444 and the health check will fail as we haven't configured that yet. The internal load balancer will be configured after the kubernetes upgrade in the init() function.

scripts/common/kubernetes.sh
```diff
@@ -252,18 +252,30 @@ function kubernetes_has_remotes() {
...
+    kubectl get cm -n kube-system kurl-config -o jsonpath='{ .data.kubernetes_api_address }'
+}
+
+# During the upgrade user might change the load balancer endpoint or want to use EKCO internal load balancer. So, we
+# to be checking the api endpoint status on the existing api server endpoint as the new endpoint is only available after
+# finishing the upgrade.
 function kubernetes_api_address() {
-    local addr="$LOAD_BALANCER_ADDRESS"
-    local port="$LOAD_BALANCER_PORT"
+    if [ -n "$upgrading_kubernetes" ]; then
```
emosbaugh, line 264: Where is this variable defined? It is scoped as local to the functions upgrade_kubernetes_local_master_patch and upgrade_kubernetes_local_master_minor

emosbaugh: so i tested this and it works! blows my mind. weird how scoping variables works in bash.

scripts/common/kubernetes.sh
pavanreddymaley: @jala-dx  It will fetch the api server address from cluster instead of variables defined in install.sh script. So, That during the upgrade with ekco we don't fail the build by validating the api server health from the variables ( localhost:6444 ) which might be not present yet.

scripts/common/kubernetes.sh
pavanreddymaley: @emosbaugh  Sorry didn't get your question exactly. I am defining the variable only in the  upgrade functions and during the normal installation the if condition will result false as the variable is not defined so the value is null. So, continues with the build.

emosbaugh:
[sc-42741]

emosbaugh:
Merging this and will kick of tests since we are unable to for external contributors at the moment.

PR #2686, opened 2022-02-11
Save private IP on initial run and don't prompt on future runs

divolgin:
#### What type of PR is this?

type::bug

On initial install.sh run, we automatically detect host's private IP.  On subsequent runs, the IP address cannot be detected automatically, so we read it from distro's config files.   k3s does not have a file where this IP address is easily accessible.

This PR creates it's own file in kURL's install directory (`/var/lib/kurl/kurl.env` by default) and loads it on subsequent runs.

Fixes #

```release-note
* Fixes a bug that caused installer to prompt the user for host's private IP when it should be detected automatically.
```

NONE

### DIFF
--- a/scripts/common/k3s.sh
+++ b/scripts/common/k3s.sh
@@ -256,6 +256,7 @@ EOF
 }
 
 function k3s_post_init() {
+    k3s_save_known_vars
     kurl_config
 }
 
@@ -596,4 +597,15 @@ EOF
     fi
 
     source /etc/profile
+}
+
+KNOWN_VARS_FILE="$KURL_INSTALL_DIRECTORY/kurl.env"
+function k3s_load_known_vars() {
+    if [ -f "$KNOWN_VARS_FILE" ]; then
+        source $KNOWN_VARS_FILE
+    fi
+}
+
+function k3s_save_known_vars() {
+    echo "PRIVATE_ADDRESS=$PRIVATE_ADDRESS" > $KNOWN_VARS_FILE
 }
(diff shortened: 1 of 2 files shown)


scripts/distro/k3s/distro.sh
```diff
@@ -1,6 +1,7 @@
 # TODO (dan): consolidate this with the rke2 distro
 function k3s_discover_private_ip() {
-    echo "$(cat /var/lib/rancher/k3s/agent/pod-manifests/etcd.yaml 2>/dev/null | grep initial-cluster | grep -o "${HOSTNAME}-[a-z0-9]*=https*://[^\",]*" | sed -n -e 's/.*https*:\/\/\(.*\):.*/\1/p')"
```
divolgin, line 3: this file does not exist

laverya (approved):

scripts/common/k3s.sh
```diff
@@ -597,3 +598,14 @@ EOF
 
     source /etc/profile
 }
+
+KNOWN_VARS_FILE="$KURL_INSTALL_DIRECTORY/kurl.env"
```
laverya, line 602: only worry is if we start storing multiple things here IMO (in case they stomp each other)
but it's fine for now

laverya (approved):

scripts/common/k3s.sh
divolgin: yes, totally agree.  it will need to be atomic. load once in the beginning and save once at the end

PR #2695, opened 2022-02-14
if known, ekco should use the local IP address as the LB

laverya:
instead of just using 'localhost'

#### What type of PR is this?
type::bug

This PR changes the behavior of EKCO to use the local IP of the node as the LB address when providing an internal load balancer, because the LB address is used as the api server address later in the codebase, and that should not be the literal string `localhost`. 

Specifically, it breaks join scripts:

```
To add worker nodes to this installation, copy and unpack this bundle on your other nodes, and run the following:

    cat ./join.sh | sudo bash -s airgap kubernetes-master-address=localhost:6444
```

Fixes #

```release-note
Fix an error where using EKCO with `EnableInternalLoadBalancer: true` would cause join scrips to be generated incorrectly.
```

### DIFF
--- a/addons/ekco/0.11.0/install.sh
+++ b/addons/ekco/0.11.0/install.sh
@@ -33,8 +33,13 @@ function ekco_pre_init() {
     if [ "$EKCO_ENABLE_INTERNAL_LOAD_BALANCER" = "1" ]; then
         EKCO_ENABLE_INTERNAL_LOAD_BALANCER_BOOL=true
         HA_CLUSTER=1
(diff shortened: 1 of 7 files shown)


laverya:
Thiiiis may break HA, testing now

laverya:
yeah, this "fix" breaks HA by making all the nodes rely on the LB of the initial node

needs more thought

laverya:
This was not actually the problem. The join preflights failing was the problem.

I'll make a different PR for that.

PR #2698, opened 2022-02-15
Move kURL API to okteto

sgalsaleh:
#### What type of PR is this?

type::chore

moves kURL API to okteto

* removed mysql because the api will just use the instance from vandoor

```release-note
NONE
```

NONE


web/package.json
```diff
@@ -26,7 +26,7 @@
     "node-fetch": "^2.6.7",
     "object-hash": "^2.2.0",
     "pino": "^6.13.0",
-    "pino-pretty": "^7.3.0",
+    "pino-pretty": "5.1.3",
```
sgalsaleh, line 29: this was bumped by dependabot very recently and it caused things to break in dev mode, so i reverted it for now

marccampbell (dismissed):

marccampbell (approved):

divolgin (approved):

PR #2710, opened 2022-02-22
wait for prometheus pods to scale to 0 before running pvmigrate

laverya:
this avoids possible race conditions when pvmigrate attempts to scale down prometheus itself

Fixes #

```release-note

```

### DIFF
--- a/scripts/common/rook.sh
+++ b/scripts/common/rook.sh
@@ -80,6 +80,13 @@ function remove_rook_ceph() {
     printf "Data within /var/lib/rook, /opt/replicated/rook and any bound disks has not been freed.\n"
 }
 
+function prometheus_pods_gone() {
+    if kubectl -n monitoring get pods -l app=prometheus 2>&1 | grep 'prometheus' &>/dev/null ; then
+        return 1
+    fi
+    return 0
+}
+
 # scale down prometheus, move all 'rook-ceph' PVCs to 'longhorn', scale up prometheus
 function rook_ceph_to_longhorn() {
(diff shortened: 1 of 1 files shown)


jala-dx (approved):

jala-dx (approved):

laverya:
If ekco and prometheus are not scaled down before pvmigrate runs, you get something that looks like this:
```
Scaling down StatefulSets and Deployments with matching PVCs
scaling StatefulSet prometheus-k8s from 0 to 0 in monitoring
scaling StatefulSet kotsadm-postgres from 1 to 0 in default

Waiting for pods with mounted PVCs to be cleaned up
Found pod prometheus-k8s-0 in monitoring mounting to-be-migrated PVC prometheus-k8s-db-prometheus-k8s-0, waiting
Found pod kotsadm-postgres-0 in default mounting to-be-migrated PVC kotsadm-postgres-kotsadm-postgres-0, waiting
Found pod kotsadm-postgres-0 in default mounting to-be-migrated PVC kotsadm-postgres-kotsadm-postgres-0, waiting
Found pod kotsadm-postgres-0 in default mounting to-be-migrated PVC kotsadm-postgres-kotsadm-postgres-0, waiting
failed to scale down pods: pod prometheus-k8s-0 in monitoring mounting prometheus-k8s-db-prometheus-k8s-0 was created at 2022-02-07T23:00:51Z, after scale-down started at 2022-02-07T23:00:32Z. It is likely that there is some other operator scaling this back up
```

PR #2711, opened 2022-02-22
swap openssl libs package for compatible version on RHEL7

stefanrepl:
#### What type of PR is this?
type::bug

This PR does fixes an issue regarding conflicting openssl-lib packages when installing on RHEL7.  The implementation swaps out any installed openssl-lib package for the correct one required for RHEL7 installation.

Fixes #
https://github.com/replicated-collab/puppet-kots/issues/208

Tested by deploying to a RHEL7 VM in GCP with no issue, confirmed installation completed and all pods came up.  Running VM can be viewed [here](https://console.cloud.google.com/compute/instancesDetail/zones/us-central1-c/instances/stefan-rhel7?authuser=0&project=replicated-qa).

```release-note
     Fixes a bug that caused RHEL7 installations to fail because of conflicting openssl-lib package versions.
```

### DIFF
--- a/scripts/common/host-packages.sh
+++ b/scripts/common/host-packages.sh
@@ -164,6 +164,14 @@ EOF
     yum makecache --disablerepo=* --enablerepo=kurl.local
 
     # shellcheck disable=SC2086
+    if [[ "${packages[*]}" == *"openssl"* && -n $(uname -r | grep "el7") ]]; then
+        installed_version=$(yum list available | grep "openssl-libs" | awk '{print $2}' | cut -c 3-)
+        # if there is already an openssl-libs package installed, swap with the package version needed for RHEL7
+        if [[ -n "${installed_version}" ]]; then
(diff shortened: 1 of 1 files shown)


scripts/common/host-packages.sh
```diff
@@ -164,6 +164,13 @@ EOF
     yum makecache --disablerepo=* --enablerepo=kurl.local
 
     # shellcheck disable=SC2086
```
laverya, line 166: could you copy this comment so that it's present for both the openssl and containerd.io `if` lines?
```
# shellcheck disable
```
only affects the line immediately after it

divolgin (approved):

scripts/common/host-packages.sh
laverya: (also, we should really look at improving our shellcheck checking in general and maybe add it to CI)

stefanrepl: done

divolgin (approved):

PR #2723, opened 2022-02-28
feat(longhorn): add storage overprovisioning field to installer

DanStough:
#### What type of PR is this?
type::feature

This adds a new field in the Longhorn add to set the [Storage Over Provisioning Percentage](https://longhorn.io/docs/1.2.3/references/settings/#storage-over-provisioning-percentage). For certain workloads, like ES and Hadoop, they rely on the PVC getting full to know when to garbage collect.

Fixes [SC-43203](https://app.shortcut.com/replicated/story/43203/in-default-config-longhorn-does-not-guarantee-space-requested-by-pvc)

```release-note
- Adds the option to the Longhorn add-on to specify the [Storage Over Provisioning Percentage](https://longhorn.io/docs/1.2.3/references/settings/#storage-over-provisioning-percentage).
```

[Kurl.sh PR](https://github.com/replicatedhq/kurl.sh/pull/715)


kurlkinds/config/crds/v1beta1/cluster.kurl.sh_installers.yaml
```diff
@@ -1704,10 +1708,6 @@ spec:
                     type: string
                   noProxy:
                     type: boolean
-                  preflightIgnore:
```
stefanrepl, line 1707: Why were these removed?

addons/longhorn/1.2.2/install.sh
```diff
@@ -1,12 +1,23 @@
...
     fi
     if [ -z "$LONGHORN_UI_REPLICA_COUNT" ]; then
         LONGHORN_UI_REPLICA_COUNT="0"
     fi
+    # Error check that someone didn't add the percent sign to this integer
+    if [ -n "$LONGHORN_STORAGE_OVER_PROVISIONING_PERCENTAGE" ] && ! [[ $LONGHORN_STORAGE_OVER_PROVISIONING_PERCENTAGE =~ $re ]]; then
+        bail "You entered ${LONGHORN_STORAGE_OVER_PROVISIONING_PERCENTAGE} for LONGHORN_STORAGE_OVER_PROVISIONING_PERCENTAGE, but it must be an integer. e.g. 200"
+    fi
+    if [ -z "$LONGHORN_STORAGE_OVER_PROVISIONING_PERCENTAGE" ]; then
+        LONGHORN_STORAGE_OVER_PROVISIONING_PERCENTAGE="200"
```
stefanrepl, line 19: Just curious, can you explain to me why 200 is a sane default?

DanStough: Mostly just that's what the default from Longhorn is and I didn't want to change the existing behavior. The theory is that most workloads don't use up there volumes so it's fine to over-provision.  Some more discussion here: https://longhorn.io/docs/1.2.3/references/settings/#storage-over-provisioning-percentage

addons/longhorn/template/base/install.sh
```diff
@@ -139,7 +151,7 @@ function maybe_init_hosts() {
         local desired=$(kubectl get daemonsets -n longhorn-system longhorn-manager --no-headers | tr -s ' ' | cut -d ' ' -f2)
         local ready=$(kubectl get daemonsets -n longhorn-system longhorn-manager --no-headers | tr -s ' ' | cut -d ' ' -f4)
 
-        if [ "$desired" = "$ready" ] && [ -n "$desired" ] && [ "$desired" > "1" ]; then
+        if [ "$desired" = "$ready" ] && [ -n "$desired" ] && [ "$desired" -ge "1" ]; then
```
DanStough, line 154: This was a shellcheck error.

kurlkinds/config/crds/v1beta1/cluster.kurl.sh_installers.yaml
DanStough: The code generator wasn't run the last time these were changed, so that's why they are showing up in this PR.

Turns out it was my fault on the previous work for host preflights, there were renamed to the flags above: https://github.com/replicatedhq/kurl/blob/60b7cd77f2fbf869becb7b780c8f639c68a1e0ee/kurlkinds/pkg/apis/cluster/v1beta1/installer_types.go#L123-L126

addons/longhorn/template/base/install.sh
DanStough: I also don't think it was doing anything before, because the value of "1" is totally valid.

kurlkinds/config/crds/v1beta1/cluster.kurl.sh_installers.yaml
stefanrepl: gotchya cool thanks!

stefanrepl (approved):

PR #2724, opened 2022-03-02
add resource limits for envoy

jala-dx:
During high CPU operations, envoy pod may not get scheduled to run and hence does not get to run liveness probes and hangs for a long time. Specifying CPU limits may help.

https://github.com/replicated-collab/ferolabs-replicated/issues/1

type::bug


laverya (approved):
Looks good to me! Please wait for testgrid to pass before merging this though - it should be all green.

(I just added improved script-based testing to this addon, so I want to get some use out of it!)

PR #2727, opened 2022-03-04
Add EKCO version 0.18.0 to fix overriding pods images

sgalsaleh:
#### What type of PR is this?

type::bug

Adds EKCO version 0.18.0 which fixes a bug that was causing overriding pods images to not work.

```release-note
Adds EKCO version 0.18.0 which fixes a bug that was causing overriding pods images to not work.
```

NONE


web/src/installers/versions.js
```diff
@@ -306,6 +306,8 @@ module.exports.InstallerVersions = {
     "0.0.1",
   ],
   ekco: [
+    "0.18.0",
```
jala-dx, line 309: nice thank you

jala-dx (approved):

PR #2747, opened 2022-03-14
set admin.conf ownership to root:root

garcialuis:
#### What type of PR is this?
type::feature (SC-31335)

Part of CIS Benchmark Compliance Remediation

Fixes #

```release-note

```


emosbaugh (commented):
Does this affect the outro in any way? What I can't figure out is why it was like this before https://github.com/replicatedhq/kURL/blob/dfd58e08e8ad5d300f3d983b7d0aeb13dc794fa2/scripts/common/common.sh#L511-L537

scripts/common/common.sh
```diff
@@ -327,11 +327,9 @@ exportKubeconfig() {
     local kubeconfig
     kubeconfig="$(${K8S_DISTRO}_get_kubeconfig)"
 
-    current_user_sudo_group
-    if [ -n "$FOUND_SUDO_GROUP" ]; then
-        chown root:$FOUND_SUDO_GROUP ${kubeconfig}
-    fi
-    chmod 440 ${kubeconfig}
+    chown root:root ${kubeconfig}
+    chmod 444 ${kubeconfig}
```
laverya, line 331: I just know someone's going to complain to us about this, and I think a much better long term solution would be to make a copy of the kubeconfig file that's owned by the sudo user group

but this'll work for the moment

laverya (dismissed):

laverya:
> Does this affect the outro in any way? What I can't figure out is why it was like this before
> 
> https://github.com/replicatedhq/kURL/blob/dfd58e08e8ad5d300f3d983b7d0aeb13dc794fa2/scripts/common/common.sh#L511-L537

Thaaaat is actually exactly the logic that I wanted to see
@garcialuis can we actually set the permisions for admin.conf to `440` and always create a kubeconfig for the user instead?

garcialuis:
> Thaaaat is actually exactly the logic that I wanted to see @garcialuis can we actually set the permisions for admin.conf to `440` and always create a kubeconfig for the user instead?

Okay, yeah - I'll change that back and create a separate kubeconfig.

emosbaugh:
> > Thaaaat is actually exactly the logic that I wanted to see @garcialuis can we actually set the permisions for admin.conf to `440` and always create a kubeconfig for the user instead?
> 
> Okay, yeah - I'll change that back and create a separate kubeconfig.

We will have to maintain that additional copy of the config and update it when the admin.conf changes. I know that certs eventually get rotated before they expire and the admin.conf may change.

emosbaugh:
> Thaaaat is actually exactly the logic that I wanted to see @garcialuis can we actually set the permisions for admin.conf to `440` and always create a kubeconfig for the user instead?

Actually @laverya wont that affect other users other that may log into the machine other than the one that actually installs kurl?

laverya:
> Actually @laverya wont that affect other users other that may log into the machine other than the one that actually installs kurl?

It will, but I think it's still better than letting _every_ user of the instance run kubectl commands as admin

Wampir0990 (dismissed):

scripts/common/common.sh
emosbaugh: I think we should put this behind the feature flag that Lavery added in the spec.

garcialuis: I updated this file now to check for the CIS_COMPLIANCE flag that Andrew introduced

emosbaugh (commented):
This looks like it's bringing in a lot of changes.

garcialuis:
Closing this outdate pr, replaced by [2756](https://github.com/replicatedhq/kURL/pull/2756).

PR #2756, opened 2022-03-16
Change ownership of admin.conf to root:root if CIS_COMPLIANCE flag is set

garcialuis:
#### What type of PR is this?
type::feature (SC-31335)

Part of CIS Benchmark Compliance Remediation

Fixes #

```release-note

```

### DIFF
--- a/scripts/common/common.sh
+++ b/scripts/common/common.sh
@@ -327,11 +327,17 @@ exportKubeconfig() {
     local kubeconfig
     kubeconfig="$(${K8S_DISTRO}_get_kubeconfig)"
 
-    current_user_sudo_group
-    if [ -n "$FOUND_SUDO_GROUP" ]; then
-        chown root:$FOUND_SUDO_GROUP ${kubeconfig}
+    if [ "$CIS_COMPLIANCE" == "1" ]; then
+        chown root:root ${kubeconfig}
+        chmod 444 ${kubeconfig}
+    else
+        current_user_sudo_group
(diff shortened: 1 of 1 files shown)


laverya (approved):

scripts/common/common.sh
```diff
@@ -327,11 +327,17 @@ exportKubeconfig() {
     local kubeconfig
     kubeconfig="$(${K8S_DISTRO}_get_kubeconfig)"
 
-    current_user_sudo_group
-    if [ -n "$FOUND_SUDO_GROUP" ]; then
-        chown root:$FOUND_SUDO_GROUP ${kubeconfig}
+    if [ "$CIS_COMPLIANCE" == "1" ]; then
+        chown root:root ${kubeconfig}
+        chmod 444 ${kubeconfig}
```
emosbaugh, line 332: i feel like in this case the `kubeconfig_setup_outro` is no longer accurate since `FOUND_SUDO_GROUP` is no longer necessary for kubectl access.

emosbaugh (approved):

scripts/common/common.sh
garcialuis: @emosbaugh I added a comment and updated kubeconfig_setup_outro to check for CIS_COMPLIANCE as well. If it's set it won't need FOUND_SUDO_GROUP.

emosbaugh: thanks. lgtm

PR #2757, opened 2022-03-16
CIS audit logging

laverya:

This will enable audit logging for k8s, using a policy derived from https://github.com/kubernetes/kubernetes/blob/7b5c718/cluster/gce/gci/configure-helper.sh#L1176. 

Log files will be stored for at most three days, and may be 100MB in size (x3 for three files).

This PR stores logs in `/var/log/apiserver`.

Fixes #

```release-note
kURL will now enable k8s audit logging, storing logs within `/var/log/apiserver`.
```

### DIFF
--- a/scripts/install.sh
+++ b/scripts/install.sh
@@ -184,7 +184,7 @@ function init() {
 
     # ensure that /etc/kubernetes/audit.yaml exists
     cp $kustomize_kubeadm_init/audit.yaml /etc/kubernetes/audit.yaml
-    mkdir -p /etc/kubernetes/auditlog
+    mkdir -p /var/log/apiserver
 
     set -o pipefail
     kubeadm init \
--- a/scripts/install.sh
+++ b/scripts/install.sh
@@ -184,6 +184,7 @@ function init() {
 
     # ensure that /etc/kubernetes/audit.yaml exists
     cp $kustomize_kubeadm_init/audit.yaml /etc/kubernetes/audit.yaml
+    mkdir -p /etc/kubernetes/auditlog
 
     set -o pipefail
     kubeadm init \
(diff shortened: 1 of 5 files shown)


scripts/install.sh
```diff
@@ -182,6 +182,10 @@ function init() {
         mv -f /etc/kubernetes/pki/apiserver.key /tmp/
     fi
 
+    # ensure that /etc/kubernetes/audit.yaml exists
+    cp $kustomize_kubeadm_init/audit.yaml /etc/kubernetes/audit.yaml
+    mkdir -p /etc/kubernetes/auditlog
```
jala-dx, line 187: the above two lines need to be swapped? also the destination dir path is `/etc/kubernetes/auditlog/audit.yaml`

scripts/join.sh
```diff
@@ -73,6 +73,10 @@ function join() {
     $DIR/bin/yamlutil -r -fp $KUBEADM_CONF_DIR/kubeadm_conf_copy_in -yp metadata
     mv $KUBEADM_CONF_DIR/kubeadm_conf_copy_in $KUBEADM_CONF_FILE
 
+    # ensure that /etc/kubernetes/audit.yaml exists
+    cp $kustomize_kubeadm_join/audit.yaml /etc/kubernetes/audit.yaml
+    mkdir -p /etc/kubernetes/auditlog
```
jala-dx, line 78: same as above

scripts/install.sh
laverya: no, they're two different things - the mkdir is for where we're logging to, and the audit.yaml is the spec of what to log

jala-dx (approved):

PR #2758, opened 2022-03-17
add pod-security-policy for CIS compliance

pavansokkenagaraj:
#### What type of PR is this?
type::security

add pod-security-policy for cis compliance

Fixes # [sc-31335](https://app.shortcut.com/replicated/story/31335/cis-benchmark-compliance-remediation)

```release-note
Fixes CIS Compliance - Ensure the admission control plugin PodSecurityPolicy is set
Fixes CIS Compliance - Ensure the --insecure-port argument is set to 0
```

### DIFF
--- a/scripts/common/common.sh
+++ b/scripts/common/common.sh
@@ -219,6 +219,17 @@ function has_default_namespace() {
     kubectl get ns | grep -q '^default' 2>/dev/null
 }
 
+function wait_for_kube_apiserver(){
+        if ! spinner_until 120 get_kube_apiserver_succeeds ; then
+        # this should exit script on non-zero exit code and print error message
+        kubectl get pod -n kube-system -l component=kube-apiserver 1>/dev/null
+    fi
+}
+
+function get_kube_apiserver_succeeds() {
+    kubectl get pod -n kube-system -l component=kube-apiserver >/dev/null 2>&1
+}
+
 # Label nodes as provisioned by kurl installation
 # (these labels should have been added by kurl installation.
 #  See kubeadm-init and kubeadm-join yaml files.
--- a/scripts/common/common.sh
+++ b/scripts/common/common.sh
@@ -219,17 +219,6 @@ function has_default_namespace() {
     kubectl get ns | grep -q '^default' 2>/dev/null
 }
 
-function wait_for_kube_apiserver(){
-        if ! spinner_until 120 get_kube_apiserver_succeeds ; then
-        # this should exit script on non-zero exit code and print error message
-        kubectl get pod -n kube-system -l component=kube-apiserver 1>/dev/null
-    fi
-}
-
-function get_kube_apiserver_succeeds() {
-    kubectl get pod -n kube-system -l component=kube-apiserver >/dev/null 2>&1
-}
-
 # Label nodes as provisioned by kurl installation
 # (these labels should have been added by kurl installation.
 #  See kubeadm-init and kubeadm-join yaml files.
--- a/scripts/common/kubernetes.sh
+++ b/scripts/common/kubernetes.sh
@@ -289,6 +289,7 @@ function spinner_containerd_is_healthy() {
 # With AWS NLB kubectl commands may fail to connect to the Kubernetes API immediately after a single
 # successful health check
 function spinner_kubernetes_api_stable() {
+    echo "Waiting for kubernetes api health to report ok"
     for i in {1..10}; do
         sleep 1
         spinner_kubernetes_api_healthy
--- a/scripts/install.sh
+++ b/scripts/install.sh
@@ -210,12 +210,16 @@ function init() {
     spinner_kubernetes_api_stable
 
     exportKubeconfig
+    KUBEADM_TOKEN_CA_HASH=$(cat /tmp/kubeadm-init | grep 'discovery-token-ca-cert-hash' | awk '{ print $2 }' | head -1)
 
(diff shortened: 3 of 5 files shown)


scripts/common/common.sh
```diff
@@ -219,6 +219,17 @@ function has_default_namespace() {
...
+function wait_for_kube_apiserver(){
+        if ! spinner_until 120 get_kube_apiserver_succeeds ; then
+        # this should exit script on non-zero exit code and print error message
+        kubectl get pod -n kube-system -l component=kube-apiserver 1>/dev/null
+    fi
+}
+
+function get_kube_apiserver_succeeds() {
+    kubectl get pod -n kube-system -l component=kube-apiserver >/dev/null 2>&1
+}
```
laverya, line 231: is this the same as kubeadm_api_is_healthy?

https://github.com/replicatedhq/kURL/blob/2f3e46c9c4618f539f50fd99d9dc589034901368/scripts/distro/kubeadm/distro.sh#L185-L187

scripts/install.sh
```diff
@@ -208,6 +212,15 @@ function init() {
     exportKubeconfig
     KUBEADM_TOKEN_CA_HASH=$(cat /tmp/kubeadm-init | grep 'discovery-token-ca-cert-hash' | awk '{ print $2 }' | head -1)
 
+    if [ "$CIS_COMPLIANCE" == "1" ]; then
+        kubectl apply -f $kustomize_kubeadm_init/pod-security-policy-privileged.yaml
+        # patch 'PodSecurityPolicy' to kube-apiserver and wait for kube-apiserver to reconcile
+        old_admission_plugins='--enable-admission-plugins=NodeRestriction'
+        new_admission_plugins='--enable-admission-plugins=NodeRestriction,PodSecurityPolicy'
+        sed -i "s%$old_admission_plugins%$new_admission_plugins%g"  /etc/kubernetes/manifests/kube-apiserver.yaml
+        wait_for_kube_apiserver
```
laverya, line 221: is there a race condition here?
What happens if the pod hasn't gone down yet before the test starts?

pavansokkenagaraj: Once we patch the kube-apiserver, the reconciler will instantly bring the pod down as it tries to match the new kube-apiserver spec and create a new pod/container.
In, wait_for_kube_apiserver there is kube-apiserver to send requests to, it will timeout and keep on retrying till the kube-apiserver is up and running.

I didn't see a latency issue when testing, but if the kube-apiserver isn't reconciled quickly, the wait_for_kube_apiserver may return old pod status.
Should we add a sleep 1 or sleep 2 to cover the latency???

scripts/common/common.sh
pavansokkenagaraj: yes, this one is better as it will check for api-server to be in running state

emosbaugh:
@pavansokkenagaraj can you explain this PSP or where you got it from? Thanks!

scripts/install.sh
```diff
@@ -208,6 +212,15 @@ function init() {
     exportKubeconfig
     KUBEADM_TOKEN_CA_HASH=$(cat /tmp/kubeadm-init | grep 'discovery-token-ca-cert-hash' | awk '{ print $2 }' | head -1)
 
+    if [ "$CIS_COMPLIANCE" == "1" ]; then
+        kubectl apply -f $kustomize_kubeadm_init/pod-security-policy-privileged.yaml
+        # patch 'PodSecurityPolicy' to kube-apiserver and wait for kube-apiserver to reconcile
+        old_admission_plugins='--enable-admission-plugins=NodeRestriction'
+        new_admission_plugins='--enable-admission-plugins=NodeRestriction,PodSecurityPolicy'
+        sed -i "s%$old_admission_plugins%$new_admission_plugins%g"  /etc/kubernetes/manifests/kube-apiserver.yaml
+        spinner_kubernetes_api_stable
```
laverya, line 221: we still need a wait before starting this to avoid races, and a log to tell us what we're waiting for

pavansokkenagaraj:
@emosbaugh 
Since rke2 is already cis complaint, I used their unrestricted psp https://github.com/rancher/rke2/blob/master/pkg/rke2/psp_templates.go#L84-L112 
If we want to make it more strict we can use/ eventually while fixing CIS benchmarks Pod 
we will be using https://github.com/rancher/rke2/blob/master/pkg/rke2/psp_templates.go#L32-L68

scripts/install.sh
pavansokkenagaraj: https://github.com/replicatedhq/kURL/blob/978744d11c332ecda957fedd15e2de86e9f9781a/scripts/common/kubernetes.sh#L289-L296

`spinner_kubernetes_api_stable` does that internally

pavansokkenagaraj: + added logging
https://github.com/replicatedhq/kURL/blob/a83988476583d417acede5ebb6bda173a7a669a7/scripts/common/kubernetes.sh#L289-L297

scripts/common/kubernetes.sh
```diff
@@ -289,6 +289,7 @@ function spinner_containerd_is_healthy() {
 # With AWS NLB kubectl commands may fail to connect to the Kubernetes API immediately after a single
 # successful health check
 function spinner_kubernetes_api_stable() {
+    echo "Waiting for kubernetes api health to report ok"
```
laverya, line 292: generally we do this in the calling function, not the spinner code itself, but this'll be fine

laverya (approved):

PR #2814, opened 2022-04-04
Use kurl version from spec if provided

laverya:
#### What type of PR is this?
type::bug

This fixes the bug preventing the kurlVersion that is specified in the installer spec from being fully applied. 

Fixes #
[SC-42119](https://app.shortcut.com/replicated/story/42119/kurlversion-isn-t-fully-applied-when-it-s-set-as-part-of-the-kurl-manifest)

```release-note
NONE
```

NONE

### DIFF
--- a/web/src/controllers/Scripts.ts
+++ b/web/src/controllers/Scripts.ts
@@ -52,6 +52,11 @@ export class Installers {
       response.status(404);
       return notFoundResponse;
     }
+
+    if (!kurlVersion && installer.spec.kurl) {
+      kurlVersion = installer.spec.kurl.installerVersion
+    }
+
     installer = await installer.resolve();
 
     response.set("X-Kurl-Hash", installer.hash());
@@ -79,6 +84,11 @@ export class Installers {
       response.status(404);
       return notFoundResponse;
     }
+
+    if (!kurlVersion && installer.spec.kurl) {
+      kurlVersion = installer.spec.kurl.installerVersion
+    }
+
     installer = await installer.resolve();
 
     response.set("X-Kurl-Hash", installer.hash());
@@ -100,6 +110,11 @@ export class Installers {
       response.status(404);
       return notFoundResponse;
     }
+
+    if (!kurlVersion && installer.spec.kurl) {
+      kurlVersion = installer.spec.kurl.installerVersion
+    }
+
     installer = await installer.resolve();
 
     response.set("X-Kurl-Hash", installer.hash());
--- a/web/src/controllers/Scripts.ts
+++ b/web/src/controllers/Scripts.ts
@@ -118,7 +118,7 @@ export class Installers {
 
     response.set("X-Kurl-Hash", installer.hash());
     response.status(200);
-    return this.templates.renderInstallScript(installer, kurlVersion);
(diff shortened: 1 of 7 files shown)


web/src/controllers/Scripts.ts
```diff
@@ -118,7 +118,7 @@ export class Installers {
 
     response.set("X-Kurl-Hash", installer.hash());
     response.status(200);
-    return this.templates.renderInstallScript(installer, kurlVersion);
+    return this.templates.renderInstallScript(installer, installer.spec.kurl?.installerVersion);
```
emosbaugh, line 121: wont this make kurlversion in the url no longer work?

laverya: no, because of the code change to have the version in the URL inserted into the installer spec

emosbaugh: so what takes precedence? what should take precedence?

garcialuis: This handler is not used to get installers for custom specs, this just gets the latest installer or a specific version that is specified in the URL. With this change the kurlVersion from the URL will be added into the spec (line 117), at the end of the day these two would be the same: 
`return this.templates.renderInstallScript(installer, kurlVersion);`
`return this.templates.renderInstallScript(installer, installer.spec.kurl?.installerVersion);`

web/src/controllers/Scripts.ts
```diff
@@ -145,6 +145,11 @@ export class Installers {
       response.status(404);
       return notFoundResponse;
     }
+
+    if (!kurlVersion && installer.spec.kurl) {
```
garcialuis, line 149: @emosbaugh Here's where we use the custom spec that may have the installerVersion. I added this condition to check for the installerVersion in the spec if there wasn't one provided in the URL.

laverya:
Could you add a daily test that 1. specifies an old kurl installer version, and 2. somehow checks that that old version is used?
(maybe grep the install.sh script for something that was removed?)

web/src/util/services/templates.ts
```diff
@@ -136,6 +136,7 @@ export async function manifestFromInstaller(i: Installer, kurlUrl: string, repli
   if (kurlVersion) {
     kurlUtilImage = `replicated/kurl-util:${kurlVersion}`;
     kurlBinUtils = `kurl-bin-utils-${kurlVersion}.tar.gz`;
+    i.spec.kurl = {additionalNoProxyAddresses: [], installerVersion: kurlVersion}
```
laverya, line 139:
```suggestion
    if (i.spec.kurl) {
      i.spec.kurl.installerVersion = kurlVersion;
    } else {
      i.spec.kurl = {additionalNoProxyAddresses: [], installerVersion: kurlVersion}
    }
```
that way you don't clobber other kurl settings here

emosbaugh (commented):
I feel like it would be valuable to add tests to the handlers in Scripts.ts. im happy to help with that if you need.

web/src/util/services/templates.ts
garcialuis: Good catch, I've added your suggestion. Thank you

emosbaugh (approved):
This is awesome. So glad to have some controller tests!

laverya (commented):
lgtm

PR #2816, opened 2022-04-05
Fixes an issue where migrating from Docker to Containerd was failing when upgrading two minor kubernetes versions at the same time.

stefanrepl:
#### What type of PR is this?
type::bug

Delay uninstallation of docker in HA cluster until nodes are fully upgraded.

Fixes #
https://github.com/replicated-collab/swimlane-replicated/issues/248
https://github.com/replicated-collab/swimlane-support/issues/19

```release-note
Fixes an issue where migrating from Docker to Containerd was failing when upgrading two minor kubernetes versions at the same time.
```

### DIFF
--- a/scripts/upgrade.sh
+++ b/scripts/upgrade.sh
@@ -113,6 +113,10 @@ function main() {
     download_util_binaries
     get_machine_id
     merge_yaml_specs
+    # Parse yaml into bash variables so we can get final kubernetes version for cases where we are upgrading 2 minor versions
+    # Must run this prior to applying bash flag overrides since kubernetes-version gets overwritten on CLI to enforce single step minor upgrades
+    parse_yaml_into_bash_variables
+    local finalK8sVersion=${KUBERNETES_VERSION}
     apply_bash_flag_overrides "$@"
     parse_yaml_into_bash_variables
     parse_kubernetes_target_version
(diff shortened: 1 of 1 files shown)


sgalsaleh (approved):

scripts/upgrade.sh
```diff
@@ -113,6 +113,8 @@ function main() {
     download_util_binaries
     get_machine_id
     merge_yaml_specs
+    parse_yaml_into_bash_variables
+    local finalK8sVersion=${KUBERNETES_VERSION}
     apply_bash_flag_overrides "$@"
     parse_yaml_into_bash_variables
```
laverya, line 119: you're calling parse_yaml_into_bash_variables twice and I'd really like to know why

stefanrepl: The first is before flag overrides are enforced and one after.  We need to parse into variables and extract the "final k8s version" prior to flag overrides because in a 2 minor version upgrade the bash variable of KUBERNETES_VERSION gets overwritten to enforce step upgrades.

laverya: okay
this really needs a comment then
because I would have read that as a bug

laverya (dismissed):

laverya (approved):

sgalsaleh (approved):

PR #2817, opened 2022-04-06
lower requests for contour

stefanrepl:
#### What type of PR is this?
type::bug

This PR lowers the requests and limits for contour.  Previously they were set to higher values that would cause "Insufficient CPU" errors when trying to schedule nodes.  I used: https://projectcontour.io/guides/resource-limits/ as a guide to choose these limits.

Fixes #
https://github.com/replicated-collab/swaggerhub-kots/issues/41
https://github.com/replicated-collab/swaggerhub-kots/issues/70

I was able to install their installer yaml and their application with no issues.

```release-note
     Fixed an issue where contour and envoy were requesting too much CPU, causing other pods to not get scheduled in 4 CPU single node installations.
```

NONE

### DIFF
--- a/addons/contour/1.20.1/patches/resource-limits.yaml
+++ b/addons/contour/1.20.1/patches/resource-limits.yaml
@@ -11,6 +11,6 @@ spec:
       - name: envoy
         resources:
           limits:
-            cpu: "1"
+            cpu: "0.4"
           requests:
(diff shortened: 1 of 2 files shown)


ajp-io:
That's a very nice reduction in the request there :)

jala-dx (approved):

PR #2830, opened 2022-04-09
Automated Containerd version update 1.6.4

replicated-ci-kurl:
Automated changes by the [cron-containerd-update](https://github.com/replicatedhq/kURL/blob/main/.github/workflows/update-containerd.yaml) GitHub action


laverya:
We need to fix the name detection here 😆

PR #2843, opened 2022-04-16
Fix prometheus alerts

emosbaugh:
#### What type of PR is this?

type::feature

Fixes alerts

"KubeScheduler has disappeared from Prometheus target discovery."
"KubeControllerManager has disappeared from Prometheus target discovery."
"KubeProxy has disappeared from Prometheus target discovery."

Fixes NONE

```release-note
TODO
```

NONE


emosbaugh:
I need to test kube-bench to see if this violates cis benchmark

xavpaice:
rebased against current main branch

xavpaice:
IIUC, this change is effective for k8s 1.21?  Wonder if we can split out the content of addons/prometheus/0.55.0-34.10.0 from the change to templates & patches, and have a separate commit to apply that to the various addon versions?

PR #2855, opened 2022-04-21
Fix containerd migrate from docker on airgapped server

emosbaugh:
#### What type of PR is this?

type::bug

Fixes 3 issues:

- Airgapped migration from docker to containerd fails because of missing images
- Docker to containerd migration fails and reverts back to docker if script fails after migration but before removal of docker
- Airgapped Kubernetes upgrade fails with image pull errors from kubeadm

Fixes NONE

I backported the changes from the template to most older containerd addon versions. You can just review the template.

Tests are failing because there are some changes outside of the add-on that do not get run as part of the pr test.

```release-note
Fixed an issue causing a migration from Containerd to Docker to fail on airgapped instances with image pull errors.
```

NONE

### DIFF
--- a/addons/containerd/template/base/install.sh
+++ b/addons/containerd/template/base/install.sh
@@ -1,5 +1,7 @@
 
 CONTAINERD_NEEDS_RESTART=0
+CONTAINERD_DID_MIGRATE_FROM_DOCKER=0
+
 function containerd_install() {
     local src="$DIR/addons/containerd/$CONTAINERD_VERSION"
 
@@ -43,6 +45,12 @@ function containerd_install() {
         systemctl daemon-reload
         restart_systemd_and_wait containerd
         CONTAINERD_NEEDS_RESTART=0
+    fi
+
+    if [ "$AIRGAP" = "1" ] && [ "$CONTAINERD_DID_MIGRATE_FROM_DOCKER" = "1" ]; then
+        logStep "Migrating images from Docker to Containerd..."
+        containerd_migrate_images_from_docker
+        logSuccess "Images migrated successfully"
     fi
 
     # Explicitly configure kubelet to use containerd instead of detecting dockershim socket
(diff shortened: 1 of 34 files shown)


testgrid/specs/full.yaml
```diff
@@ -1025,6 +1059,99 @@
...
+    longhorn:
+      version: latest
+    registry:
+      version: latest
+    kotsadm:
+      version: latest
+      s3Disabled: true
+    containerd:
+      version: latest
+  airgap: true
```
emosbaugh, line 1154: Only the minimal case should succeed after merge. There are other issues causing failures on the more robust test cases.

addons/containerd/template/base/install.sh
```diff
@@ -1,5 +1,25 @@
 
 CONTAINERD_NEEDS_RESTART=0
+CONTAINERD_DID_MIGRATE_FROM_DOCKER=0
+
+function containerd_pre_init() {
+    local src="$DIR/addons/containerd/$CONTAINERD_VERSION"
+
+    # Explicitly configure kubelet to use containerd instead of detecting dockershim socket
+    if [ -d "$DIR/kustomize/kubeadm/init-patches" ]; then
```
laverya, line 9: what happens if there's not an init-patches directory yet?

Should this create it?

addons/containerd/template/testgrid/k8s-docker.yaml
```diff
@@ -38,4 +38,73 @@
...
+      version: latest
+    longhorn:
+      version: latest
+    registry:
+      version: latest
+    ekco:
+      version: latest
+    kotsadm:
+      version: latest
+      disableS3: true
```
laverya, line 56: IMO our on-pr tests should generally avoid including unneeded addons like this, so that they run faster

laverya (approved):

addons/containerd/template/base/install.sh
emosbaugh: I'm only copying this code but my hunch is that it is a way to detect if this is kubeadm addon init and not join

addons/containerd/template/testgrid/k8s-docker.yaml
emosbaugh: I like kotsadm addon cause it validates that things are working

emosbaugh: and it uses a pvc

laverya (dismissed):

scripts/upgrade.sh
```diff
@@ -127,12 +127,13 @@ function main() {
     configure_no_proxy
     ${K8S_DISTRO}_addon_for_each addon_fetch
     host_preflights "${MASTER:-0}" "1" "1"
-    install_cri
+    install_host_dependencies
```
emosbaugh, line 130: im making the order here consistent with install. setup_kubeadm_kustomize must come before addon_join which must come before maybe upgrade

scripts/join.sh
```diff
@@ -149,11 +149,11 @@ function main() {
     configure_no_proxy
     ${K8S_DISTRO}_addon_for_each addon_fetch
     host_preflights "${MASTER:-0}" "1" "0"
-    install_cri
+    install_host_dependencies
```
emosbaugh, line 152: im making the order here consistent with install

laverya (approved):

laverya:
actually this has a REALLY high testgrid failure rate - we should fix that first

laverya:
I missed that. Ok, let's see how it does on staging then

PR #2904, opened 2022-05-14
fix less command on amazon linux 2

garcialuis:
#### What type of PR is this?

type::bug

This PR is to fix the less command on amazon linux 2 after installing kURL. This will also fix existing installations with the bug once the installer is rerun.

Fixes # [SC-43509](https://app.shortcut.com/replicated/story/43509/kurl-installation-in-amazon-linux-2-breaks-less-command)

## Steps to reproduce

Details to reproduce the issue are listed on the SC ticket linked above. 

```release-note
NONE
```
#### Release note:
```release-note
- Fixes a bug where the less command breaks after installing kURL on amazon linux 2.
```

NONE

### DIFF
--- a/scripts/common/kubernetes.sh
+++ b/scripts/common/kubernetes.sh
@@ -92,6 +92,11 @@ function kubernetes_install_host_packages() {
 
     if kubernetes_host_commands_ok "$k8sVersion"; then
         logSuccess "Kubernetes host packages already installed"
+        # less command is broken if libtinfo.so.5 is missing in amazon linux 2
+        if [ "$LSB_DIST" == "amzn" ] && ! file_exists "/usr/lib64/libtinfo.so.5"; then
+            install_host_packages "${DIR}/packages/kubernetes/${k8sVersion}" ncurses-compat-libs
+        fi
+
         return
     fi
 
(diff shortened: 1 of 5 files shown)


scripts/common/kubernetes.sh
```diff
@@ -92,6 +92,11 @@ function kubernetes_install_host_packages() {
 
     if kubernetes_host_commands_ok "$k8sVersion"; then
         logSuccess "Kubernetes host packages already installed"
+        # less command is broken if libtinfo.so.5 is missing in amazon linux 2
+        if [ "$LSB_DIST" == "amzn" ] && ! file_exists "/usr/lib64/libtinfo.so.5"; then
+            install_host_packages "${DIR}/packages/kubernetes/${k8sVersion}" ncurses-compat-libs
```
emosbaugh, line 97: Are you sure you still need this?

testgrid/specs/full.yaml
```diff
@@ -1450,3 +1450,12 @@
       shouldDisableRebootServices: false
       shouldDisableClearNodes: false
       shouldEnablePurgeNodes: false
+- name: less_command
+  installerSpec:
+    kubernetes:
+      version: "1.23.5"
+    docker:
+      version: "20.10.5"
+  postInstallScript: |
```
emosbaugh, line 1459: Did you run this to see if it fails prior to the fix?

scripts/common/kubernetes.sh
emosbaugh: ah i see for existing installs

emosbaugh (approved):

scripts/common/kubernetes.sh
garcialuis: Yeah, I left that in for existing installs.

testgrid/specs/full.yaml
emosbaugh: Would it be red and say fail in the browser? If so yes

PR #2917, opened 2022-05-20
Add capability to have multiple primary node to testgrid

stascool:

Fixes #

```release-note

```

### DIFF
--- a/testgrid/tgrun/hack/test-spec.yaml
+++ b/testgrid/tgrun/hack/test-spec.yaml
@@ -1,6 +1,7 @@
 - name: hack
   numPrimaryNodes: 1
   numSecondaryNodes: 0
+  flags: "ha"
   installerSpec:
     kubernetes:
       version: "latest"
--- a/testgrid/tgrun/hack/test-spec.yaml
+++ b/testgrid/tgrun/hack/test-spec.yaml
@@ -23,6 +23,7 @@
       version: "latest"
     ekco:
       version: "latest"
+      enableInternalLoadBalancer: true # this must be present if you need to install with ha
     fluentd:
       version: "latest"
     minio:
--- a/testgrid/tgrun/pkg/runner/vmi/embed/common.sh
+++ b/testgrid/tgrun/pkg/runner/vmi/embed/common.sh
@@ -50,7 +50,7 @@ function wait_for_join_commandready()
         secondaryJoin=$(echo "$joinCommand" | sed 's/{.*secondaryJoin":"*\([0-9a-zA-Z]*\)"*,*.*}/\1/')
         primaryJoin=$(echo "$joinCommand" | sed 's/{.*primaryJoin":"*\([0-9a-zA-Z]*\)"*,*.*}/\1/')
         if [ "$secondaryJoin" != "" ] && [ "$primaryJoin" != "" ]; then
-            echo "$secondaryJoin"
+            echo "join command is ready"
             break
         fi
         i=$((i+1))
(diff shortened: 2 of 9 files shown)


testgrid/tgrun/pkg/runner/vmi/embed/common.sh
```diff
@@ -0,0 +1,86 @@
...
+
+function wait_for_join_commandready()
+{
+  i=0
+  while true
+    do
+        joinCommand=$(get_join_command)
+        secondaryJoin=$(echo "$joinCommand" | sed 's/{.*secondaryJoin":"*\([0-9a-zA-Z]*\)"*,*.*}/\1/')
+        if [ "$secondaryJoin" != "" ]; then
+            echo "$secondaryJoin"
```
emosbaugh, line 52: do you need this echo? seems specific to secondary anyway

testgrid/tgrun/hack/test-spec.yaml
```diff
@@ -1,6 +1,7 @@
 - name: hack
   numPrimaryNodes: 1
   numSecondaryNodes: 0
+  flags: "ha"
```
stascool, line 4: this flag isn't being used by the bash script

testgrid/tgrun/pkg/runner/vmi/embed/runcmd.sh
```diff
@@ -436,21 +442,29 @@ function wait_for_cluster_ready() {
...
+        if [ $i -gt 20 ]; then
             report_failure "cluster_not_ready"
             exit 0
         fi
-        sleep 120
+        sleep 60
     done
 }
 
+function useHA() {
```
stascool, line 453: @emosbaugh need your suggestion here, when I run the installation script  function with ``ha`` it give me this error
```
kubernetes.loadBalancerAddress required
```
SO from what I read in the doc I have add this to the spec
```
    ekco:
      version: "latest"
      enableInternalLoadBalancer: true  
```
and then I added this function and everything started to work fine but If you have a better way let me know.

testgrid/tgrun/hack/test-spec.yaml
emosbaugh: it is being used here

https://github.com/replicatedhq/kURL/blob/c54f2d37ce4f537e6f7212eb17f234392baa5c0d/testgrid/tgrun/pkg/runner/vmi/vmi.go#L122

https://github.com/replicatedhq/kURL/blob/98ac102b74ab9c043dc2533ca4d87afff820922e/testgrid/tgrun/pkg/runner/vmi/embed/runcmd.sh#L62

testgrid/tgrun/pkg/runner/vmi/embed/runcmd.sh
emosbaugh: yes please use `enableInternalLoadBalancer: true`

testgrid/tgrun/pkg/runner/vmi/embed/runcmd.sh
```diff
@@ -436,21 +442,29 @@ function wait_for_cluster_ready() {
...
+    if [ "$NUM_PRIMARY_NODES" -gt 1 ]; then
+        echo "kurl install finished with ha"
+        cat install.sh | timeout 30m bash -s $AIRGAP_FLAG ${KURL_FLAGS[@]} ha
```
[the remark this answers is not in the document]
emosbaugh: i like this approach rather than requiring it in the spec

testgrid/tgrun/hack/test-spec.yaml
stascool: I've fixed that and tested it with the following flags 
```
flags: "labels=disktype=ssd,gpu=enabled exclude-builtin-host-preflights"
```

testgrid/tgrun/pkg/runner/vmi/embed/runcmd.sh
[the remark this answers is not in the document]
stascool: So I will keep it this way, but also it will depend on ``ekco:`` need to be there in the cluster

PR #2932, opened 2022-05-25
containerd versions 1 5 x and 1.6.x

divolgin:
#### What type of PR is this?

type::feature

This adds 1.5.x and 1.6.x versions of containerd addon

Fixes #

```release-note
* Adds support for containerd 1.5.10, 1.5.11, and 1.6.4
```


web/src/installers/index.ts
```diff
@@ -1378,6 +1378,13 @@ export class Installer {
       }
     }
 
+    // Containerd 1.6 is not incompatible with Weave
```
laverya, line 1381: Is incompatible, or is not compatible

testgrid/specs/full.yaml
```diff
@@ -252,7 +252,7 @@
     contour:
       version: latest
     containerd:
-      version: 1.5.11
+      version: 1.6.04
```
laverya, line 255: Pretty sure this will break, needs to not have the zero

web/src/installers/index.ts
```diff
@@ -812,7 +812,7 @@ export class Installer {
 
     i.id = "latest";
     i.spec.kubernetes = { version: "1.23.x" };
-    i.spec.containerd = { version: this.toDotXVersion(installerVersions.containerd[0]) };
+    i.spec.containerd = { version: "1.5.x" }; // 1.6.x is not compatible with Weave
```
laverya, line 815: I would prefer to mark containerd 1.6.x as NOT latest, so that we don't have people attempting to upgrade run into it

web/src/installers/versions.js
```diff
@@ -78,7 +78,7 @@ module.exports.InstallerVersions = {
     "19.03.4",
     "18.09.8",
   ],
-  containerd: ["1.5.11", "1.5.10", "1.4.13", "1.4.12", "1.4.11", "1.4.10", "1.4.9", "1.4.8", "1.4.6", "1.4.4", "1.4.3", "1.3.9", "1.3.7", "1.2.13"], // cron-containerd-update
+  containerd: ["1.6.4", "1.5.11", "1.5.10", "1.4.13", "1.4.12", "1.4.11", "1.4.10", "1.4.9", "1.4.8", "1.4.6", "1.4.4", "1.4.3", "1.3.9", "1.3.7", "1.2.13"], // cron-containerd-update
```
laverya, line 81: I know we used to have code to make 1.4.6 the first entry here despite not being the most recent - should we do the same with 1.6.x to put it at the end?

Or a fix to weave, one of the two

laverya (dismissed):

laverya (dismissed):

laverya (dismissed):

addons/containerd/template/base/install.sh
```diff
@@ -78,7 +78,8 @@ function containerd_install() {
         # If using the internal load balancer the Kubernetes API server will be unavailable until
         # kubelet starts the HAProxy static pod. This check ensures the Kubernetes API server
         # is available before proceeeding.
-        try_5m kubectl --kubeconfig=/etc/kubernetes/kubelet.conf get nodes
+        # ".." is needed becasue addons can have a CRD names "nodes", like nodes.longhorn.io
+        try_5m kubectl --kubeconfig=/etc/kubernetes/kubelet.conf get nodes..
```
divolgin, line 82: This fixes ```cannot list resource "nodes" in API group "longhorn.io" in the namespace "default"```

laverya (approved):

PR #2936, opened 2022-05-25
use amazonlinux and install ncurses-compat-libs to fix less command

garcialuis:
#### What type of PR is this?

type::bug

This PR is to fix the less command on amazon linux 2 after installing kURL. 

Fixes # [SC-43509](https://app.shortcut.com/replicated/story/43509/kurl-installation-in-amazon-linux-2-breaks-less-command)

Most of these changes are being added back from the previous [revert](https://github.com/replicatedhq/kURL/pull/2915/files) that was done so it wouldn't block the release.
Only new changes are in [scripts/common/kubernetes.sh](https://github.com/replicatedhq/kURL/compare/luisgarcia3795/sc-43509/amazon-less-cmd?expand=1#diff-289c11f03a3d8d7b38bab6cc5e4c5c1e5590dc90fe7e25623acb9ad9efe26302):
- Now checking that it's not an airgap install.
- Checking that the kubernetes package directory exists.

## Steps to reproduce

Details to reproduce the issue are listed on the SC ticket linked above. 

```release-note
NONE
```
#### Release note:
```release-note
- Fixes a bug where the less command breaks after installing kURL on amazon linux 2.
```

NONE

### DIFF
--- a/scripts/common/kubernetes.sh
+++ b/scripts/common/kubernetes.sh
@@ -92,6 +92,13 @@ function kubernetes_install_host_packages() {
 
     if kubernetes_host_commands_ok "$k8sVersion"; then
         logSuccess "Kubernetes host packages already installed"
+        # less command is broken if libtinfo.so.5 is missing in amazon linux 2
+        if [ "$LSB_DIST" == "amzn" ] && [ "$AIRGAP" != "1" ] && ! file_exists "/usr/lib64/libtinfo.so.5"; then
+            if [ -d "$DIR/packages/kubernetes/${k8sVersion}" ]; then
+                install_host_packages "${DIR}/packages/kubernetes/${k8sVersion}" ncurses-compat-libs
+            fi
+        fi
+        
         return
     fi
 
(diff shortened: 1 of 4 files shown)


scripts/common/kubernetes.sh
```diff
@@ -92,6 +92,13 @@ function kubernetes_install_host_packages() {
 
     if kubernetes_host_commands_ok "$k8sVersion"; then
         logSuccess "Kubernetes host packages already installed"
+        # less command is broken if libtinfo.so.5 is missing in amazon linux 2
+        if [ "$LSB_DIST" == "amzn" ] && [ "$AIRGAP" != "1" ] && ! file_exists "/usr/lib64/libtinfo.so.5"; then
+            if [ -d "$DIR/packages/kubernetes/${k8sVersion}" ]; then
+                install_host_packages "${DIR}/packages/kubernetes/${k8sVersion}" ncurses-compat-libs
```
emosbaugh, line 98: will this fail if `ncurses-compat-libs` is not in the bundle?

garcialuis: @emosbaugh 

No, it won't fail. I tested this on my codeserver and it was able to install the package even though it wasn't on the `build/packages/kubernetes/1.23.5/rhel-7-force` directory. 

I added the check for the directory because I saw this message on staging last week:
```
realpath: ‘./packages/kubernetes/1.23.5’: No such file or directory
An error occurred on line
```
when I ran it on an existing install. It may be because the packages weren't rebuilt but I'm hoping that this way the installer won't break if the package directory is not there.

emosbaugh (approved):

PR #2992, opened 2022-06-06
local-path-storage provisioner

laverya:

This PR adds the local-path-provisioner from Rancher, https://github.com/rancher/local-path-provisioner. For single node installations and those that do not need data replication, it is far more lightweight than using openebs in an equivalent fashion.

Fixes #

```release-note
Add local-path-storage addon as an additional storage provisioner
```

Yes, WIP

### DIFF
--- a/addons/local-path-storage/0.0.22/install.sh
+++ b/addons/local-path-storage/0.0.22/install.sh
@@ -1,7 +1,13 @@
 
+function local-path-storage_join() {
+  local-path-storage_host_init
+}
+
 function local-path-storage() {
     local src="$DIR/addons/local-path-storage/__releasever__"
     local dst="$DIR/kustomize/local-path-storage"
+
+    local-path-storage_host_init
 
     cp "$src/local-path-storage.yaml" "$dst/"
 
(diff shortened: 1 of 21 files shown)


addons/local-path-storage/0.0.22/install.sh
```diff
@@ -0,0 +1,49 @@
...
+    local src="$DIR/addons/local-path-storage/__releasever__"
+    local dst="$DIR/kustomize/local-path-storage"
+
+    local-path-storage_host_init
+
+    cp "$src/local-path-storage.yaml" "$dst/"
+
+    cp "$src/tmpl-kustomization.yaml" "$dst/"
+
+    if has_default_storageclass && ! local_path_storage_is_default_storageclass ; then
```
emosbaugh, line 16: should we make this a property of the addon so that its less racey? it can be true by default

emosbaugh (changes requested):

addons/local-path-storage/0.0.22/install.sh
emosbaugh: I just wonder if there is a better way to specify priority. This is a bit circumstantial as they happen to be in this order today and could be more explicit. I'll approve with it but we should think of a better way to support this "feature" in the future.

emosbaugh (approved):

PR #3020, opened 2022-06-17
Dont fail if the excludeStorageClass is set to true

jala-dx:
Fixes# SC-49441

#### What type of PR is this?
type::bug

### DIFF
--- a/addons/aws/0.1.0/install.sh
+++ b/addons/aws/0.1.0/install.sh
@@ -9,7 +9,12 @@ function aws() {
         insert_resources "$dst/kustomization.yaml" storageclass.yaml
     fi
 
-    kubectl apply -k "$DIR/kustomize/aws/"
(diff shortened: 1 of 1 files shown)


addons/aws/0.1.0/install.sh
```diff
@@ -9,7 +9,12 @@ function aws() {
         insert_resources "$dst/kustomization.yaml" storageclass.yaml
     fi
 
-    kubectl apply -k "$DIR/kustomize/aws/"
+    if  [ -s "$DIR/kustomize/aws/kustomization.yaml" ]; then
```
jala-dx, line 12: basically checking if the file is empty , if the file is empty `kubectl apply -k` errors with `no objects`

laverya (approved):

emosbaugh (approved):

PR #3022, opened 2022-06-17
update weave add-on to use patched images

garcialuis:
#### What type of PR is this?
type::chore

Updates images for weave add-on versions 2.6.5 and 2.8.1. New images are patched for CVEs.

Fixes #
[SC-49336](https://app.shortcut.com/replicated/story/49336/weave-2-8-1-cves)

```release-note
Fixes CVEs for weave 2.8.1 plugin
```

NONE

### DIFF
--- a/scripts/tasks.sh
+++ b/scripts/tasks.sh
@@ -395,9 +395,15 @@ function get_weave_version() {
     local weave_version=$(KUBECONFIG=/etc/kubernetes/admin.conf kubectl get daemonset -n kube-system weave-net -o jsonpath="{..spec.containers[0].image}" | sed 's/^.*://')
     if [ -z "$weave_version" ]; then
         if [ -n "$DOCKER_VERSION" ]; then
-            weave_version=$(docker image ls | grep weaveworks/weave-npc | awk '{ print $2 }' | head -1)
+            weave_version=$(docker image ls | grep kurlsh/weave-npc | awk '{ print $2 }' | head -1)
+            if [ -z "$weave_version" ]; then
+                weave_version=$(docker image ls | grep weaveworks/weave-npc | awk '{ print $2 }' | head -1)
+            fi
         else
-            weave_version=$(crictl images list | grep weaveworks/weave-npc | awk '{ print $2 }' | head -1)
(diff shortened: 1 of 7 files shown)


emosbaugh (changes requested):
you will need to add EXEC_IMAGE env var to pods like so https://github.com/replicatedhq/replicated-installer/blob/06a2cb9fc32731386e23afe87456b23386565610/install_scripts/templates/kubernetes/yml-generate.sh#L966-L967

can you find out if we have the ability to run airgap tests on pr from @laverya ? if so can you add? i would think we would have caught this without my review

scripts/tasks.sh
```diff
@@ -396,8 +396,14 @@ function get_weave_version() {
     if [ -z "$weave_version" ]; then
         if [ -n "$DOCKER_VERSION" ]; then
             weave_version=$(docker image ls | grep weaveworks/weave-npc | awk '{ print $2 }' | head -1)
+            if [ -z "$weave_version" ]; then
+                weave_version=$(docker image ls | grep kurlsh/weave-npc | awk '{ print $2 }' | head -1)
+            fi
```
trelore, line 401: Curious what @emosbaugh thinks about this, could `docker image ls` show both images (I'm thinking in cases of an upgrade)? If so, do we want to default to the kurlsh image as that should be the more 'up-to-date' on?

emosbaugh: good point. agreed

garcialuis: I swapped them, this is now checking for the kurlsh image first.

garcialuis:
> you will need to add EXEC_IMAGE env var to pods like so https://github.com/replicatedhq/replicated-installer/blob/06a2cb9fc32731386e23afe87456b23386565610/install_scripts/templates/kubernetes/yml-generate.sh#L966-L967
> 
> can you find out if we have the ability to run airgap tests on pr from @laverya ? if so can you add? i would think we would have caught this without my review

@emosbaugh The env var and airgap test have been added now.

emosbaugh (approved):

PR #3074, opened 2022-06-30
Feature: Prefix each line of output in logs with timestamp

stascool:

Fixes #

```release-note

```

### DIFF
--- a/testgrid/tgrun/pkg/runner/vmi/embed/mainscript.sh
+++ b/testgrid/tgrun/pkg/runner/vmi/embed/mainscript.sh
@@ -1,0 +1,16 @@
+#!/usr/bin/env bash
+
+function adddate() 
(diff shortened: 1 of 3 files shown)


testgrid/tgrun/pkg/runner/vmi/embed/mainscript.sh
```diff
@@ -0,0 +1,16 @@
+#!/usr/bin/env bash
+
+function adddate() 
+{
+  while IFS= read -r line; do
+      printf '%s %s\n' "$(date)" "$line";
```
emosbaugh, line 6: can you use date in the format specified in the task? `date --rfc-3339=seconds`

emosbaugh (changes requested):

PR #3114, opened 2022-07-11
move deployment_fully_updated function to common

laverya:

This is a function that should be used a lot - it returns false if a deployment is not yet fully running. A lot of addons use `spinner pod_running` instead, which does not actually wait for the pod to be available and invites race conditions when rolling out an update.

Fixes #

```release-note

```


laverya:
I've enabled auto-merge because I do not expect this to pass testgrid, so there's no reason to wait for it (since testgrid only does addon changes, not main changes)

laverya:
lol github beat me to it

PR #3142, opened 2022-07-15
Create new Longhorn version 1.3.1

emosbaugh:

Fixes NONE

```release-note
- Adds [Longhorn add-on](https://kurl.sh/docs/add-ons/longhorn) version 1.3.1.
```

Yes. Must document limitation can only upgrade 1 minor version at a time.


emosbaugh:
[sc-53236]

addons/longhorn/template/testgrid/k8s-ctrd.yaml
```diff
@@ -137,7 +137,8 @@
     validate_testfile rwtest testfile.txt
     validate_read_write_object_store postupgrade upgradefile.txt
 
-- name: upgrade-from-oldest-longhorn
+# can only upgrade 1 minor version at a time
+- name: upgrade-from-oldest-longhorn-1.2.x
```
laverya, line 141: this just duplicates the existing 'latest' test then
I'd prefer for there to be code within the addon that skips the upgrade if it's too large

addons/longhorn/template/base/install.sh
```diff
@@ -188,11 +192,38 @@ function maybe_init_hosts() {
...
+
+    semverParse "${LONGHORN_VERSION}"
+    local next_version_minor="${minor}"
+
+    local previous_version_minor="$((next_version_minor-1))"
+
+    if [ "$current_version_minor" -lt "$previous_version_minor" ]; then
+        logFail "Upgrades to Longhorn version ${LONGHORN_VERSION} from versions prior to 1.${previous_version_minor}.x are unsupported."
+        bail "Please upgrade to the previous minor version first."
+    fi
```
laverya, line 221: I'm genuinely unsure - is this the behavior we want? For instance, in rook, we just skip the upgrade and continue on with the rest of the kurl install instead of exiting with an error

addons/longhorn/template/base/template/patch-defaults-cm.yaml
```diff
@@ -0,0 +1,10 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: longhorn-default-setting
+  namespace: longhorn-system
+data:
+  default-setting.yaml: |-
+    storage-over-provisioning-percentage: ${LONGHORN_STORAGE_OVER_PROVISIONING_PERCENTAGE}
+    upgrade-checker: false
+    priority-class: system-node-critical
```
laverya, line 10: won't this override the _entire_ `default-setting.yaml` key, and not just these three items?

addons/longhorn/template/generate.sh
```diff
@@ -25,34 +25,82 @@ function get_ksplit() {
...
     curl --silent "https://raw.githubusercontent.com/longhorn/longhorn/v$VERSION/deploy/longhorn-images.txt" | sed 's/\(.*\)\/\(.*\):\([^"]*\)/image \2 \1\/\2:\3/' >> "../${VERSION}/Manifest"
 
 }
 
+function split_resources() {
+    local combined="$1"
+    local outdir="$2"
+    local kustomization_file="$3"
+    local tmpdir="$(mktemp -d -p $outdir)"
+    csplit --quiet --prefix="$tmpdir/out" -b ".%03d.yaml" $combined "/^---$/+1" "{*}"
```
laverya, line 62: I believe ksplit will do this, and with better names - we use it in other addons

addons/longhorn/template/base/template/patch-defaults-cm.yaml
emosbaugh: yes but defaults are empty. id rather patch so as not to override/duplicate things like metadata.labels

addons/longhorn/template/generate.sh
emosbaugh: id rather leave this as it is already tested

laverya (dismissed):

addons/longhorn/template/testgrid/k8s-ctrd.yaml
emosbaugh: latest will upgrade from 1.3.1 once a new version is released. even so i addressed this

addons/longhorn/template/base/install.sh
emosbaugh: i will make the behavior mimic rook

addons/longhorn/template/base/install.sh
```diff
@@ -20,10 +21,21 @@ function longhorn_pre_init() {
     fi
 
     # Can only upgrade 1 minor version at a time
-    longhorn_bail_upgrade_unsupported
+    if ! longhorn_can_upgrade ; then
+        if ! confirmY ; then
```
laverya, line 27: I don't think this confirmy is very clear - should the user press y to continue, or to bail?
the message is:
```
        logWarn "Upgrades to Longhorn version ${LONGHORN_VERSION} from versions prior to 1.${previous_version_minor}.x are unsupported."
```

laverya (approved):

PR #3164, opened 2022-07-19
report basic metrics from testgrid runners

laverya:

Fixes #

```release-note

```

### DIFF
--- a/testgrid/tgrun/pkg/runner/metrics.go
+++ b/testgrid/tgrun/pkg/runner/metrics.go
@@ -1,0 +1,94 @@
+package runner
+
+import (
+	"bytes"
+	"context"
+	"encoding/json"
+	"fmt"
+	"net/http"
+	"os"
+
+	"github.com/pkg/errors"
+	tghandlers "github.com/replicatedhq/kurl/testgrid/tgapi/pkg/handlers"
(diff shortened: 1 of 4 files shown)


testgrid/tgrun/pkg/runner/metrics.go
```diff
@@ -0,0 +1,91 @@
...
+		return errors.Wrap(err, "failed to get hostname")
+	}
+
+	runnerStatusRequest := tghandlers.RunnerStatusRequest{
+		FreeCPU:      freeCPU,
+		FreeRAM:      freeRAM,
+		RunningTests: float64(runningVMIs),
+		Hostname:     hostName,
+	}
+	b, err := json.Marshal(runnerStatusRequest)
```
emosbaugh, line 40: handle error here?

emosbaugh (dismissed):

testgrid/tgrun/pkg/runner/metrics.go
laverya: ah yeah, we should
not that there's much chance of an error with that object though (no chance of circular dependencies)

laverya:
things I'm not 100% sure of:
will this be interpreted properly by datadog? (eg will it be split by tag, and not by API server instance?)

emosbaugh (approved):

PR #3165, opened 2022-07-19
Weave images install iptables 1.8.7

emosbaugh:
#### What type of PR is this?

type::bug

Pins iptables to 1.8.7 in patched weave images to fix issue with host incompatibility.

Fixes NONE

```release-note
- Adds [Weave add-on](https://kurl.sh/docs/add-ons/weave) versions 2.6.5-20220720 and 2.8.1-20220720.
- Fixes an issue that causes iptables commands run on the host to report error `table `filter' is incompatible, use 'nft' tool`.
```

NONE


laverya (changes requested):
I am aghast that this is all it took to fix, but happy that it IS fixed.

That being said, I would prefer for this to become a new version - eg 2.8.1-20220719 - instead of editing 2.8.1-20220616

addons/weave/template/base/patch-daemonset.yaml
```diff
@@ -18,5 +18,7 @@ spec:
               value: "--log-level=info" # default log level is debug
             - name: CHECKPOINT_DISABLE
               value: "1"
+            - name: EXEC_IMAGE
+              value: "weaveworks/weaveexec:__WEAVE_VERSION__"
```
laverya, line 22: is this useful? weave does this by default

laverya: oh nvm you're sedding it out

laverya (approved):

laverya (approved):

PR #3179, opened 2022-07-22
Add flag kubernetes-enable-internal-load-balancer

emosbaugh:
#### What type of PR is this?

type::feature

We cant automate ha without ekco. Script fails with

```
kubernetes.loadBalancerAddress required
```

https://testgrid.kurl.sh/run/pr-3177-a307e4b-ekco-0.19.3-k8s-docker-2022-07-22T18:31:53Z

Fixes NONE

```release-note
The property `kubernetes.loadBalancerUseFirstPrimary`, and equivalent flag `kubernetes-load-balancer-use-first-primary`, has been added to automatically use the first primary address as the cluster control plane endpoint. It is recommended to use the property `ekco.enableInternalLoadBalancer` instead.
```

I will not document this


kurl_util/cmd/bashmerge/main.go
```diff
@@ -145,15 +145,21 @@ func parseBashFlags(installer *kurlv1beta1.Installer, bashFlags string) error {
 			}
 			installer.Spec.Kubernetes.LoadBalancerAddress = split[1]
 		case "ekco-enable-internal-load-balancer":
-			if installer.Spec.Ekco != nil {
-				installer.Spec.Ekco.EnableInternalLoadBalancer = true
+			if installer.Spec.Ekco == nil {
+				installer.Spec.Ekco = &kurlv1beta1.Ekco{}
+			}
+			installer.Spec.Ekco.EnableInternalLoadBalancer = true
```
laverya, line 151: does this have an effect if there's no ekco version supplied?

I don't like changing this unless ekco is no longer required

emosbaugh: changed back

web/src/installers/index.ts
```diff
@@ -51,6 +52,7 @@ export const kubernetesConfigSchema = {
     masterAddress: { type: "string", flag: "kuberenetes-master-address", description: "The address of the internal Kubernetes API server, used during join scripts (read-only)" },
     clusterName: { type: "string", flag: "kubernetes-cluster-name", description: "The name of the Kubernetes cluster (default \"kubernetes\")"},
     loadBalancerAddress: { type: "string", flag: "load-balancer-address", description: "Used for High Availability installs, indicates the address of the external load balancer" },
+    loadBalancerUseFirstPrimary: { type: "boolean", flag: "kubernetes-load-balancer-use-first-primary" , description: "Use first primary address as control plane endpoint in HA cluster" },
```
laverya, line 55: can we put "do not use" in here somewhere?

laverya: or "for testing upgrades only, use ekco's internal LB instead"

emosbaugh: done

laverya (approved):

PR #3192, opened 2022-07-26
Update EKCO action doesn't need go

emosbaugh:


.github/workflows/update-longhorn.yaml
```diff
@@ -12,10 +12,6 @@ jobs:
       with:
         fetch-depth: 0
 
-    - uses: actions/setup-go@v3
-      with:
-        go-version: 1.17
-
```
laverya, line 18: longhorn uses this:

https://github.com/replicatedhq/kURL/blob/dd0ed9d2b0a3500b7d6e60419339f1d3375e1fac/addons/longhorn/template/generate.sh#L14-L23

laverya (changes requested):

.github/workflows/update-longhorn.yaml
emosbaugh: ah. got it

laverya (approved):

PR #3204, opened 2022-07-29
Testgrid additional nodes must wait until test is done

emosbaugh:
#### What type of PR is this?

type::tests

Additional nodes must stick around as part of the cluster until the test is complete otherwise multi node tests fail

### DIFF
--- a/testgrid/tgrun/pkg/runner/vmi/embed/common.sh
+++ b/testgrid/tgrun/pkg/runner/vmi/embed/common.sh
@@ -84,6 +84,32 @@ function wait_for_join_commandready()
   done
 }
 
+function wait_for_initprimary_done()
+{
+  i=0
+  while true; do
+    primaryNodeStatus=$(get_initprimary_status)
+    if [[ "$primaryNodeStatus" = "success" ]]; then
+      echo "initprimary status finsihed the test"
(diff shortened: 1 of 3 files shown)


testgrid/tgrun/pkg/runner/vmi/embed/common.sh
```diff
@@ -84,6 +84,32 @@ function wait_for_join_commandready()
...
+    fi
+    echo "initprimary not ready"
+    i=$((i+1))
+    if [ $i -gt 20 ]; then
+      echo "wait_for_initprimary_done timeout"
+      report_status_update "failed"
+      send_logs
+      exit 1
+    fi
+    sleep 60
```
rrpolanco, line 109: Worst case, we could wait up to 20 mins before a node ready. Could this negatively affect test run times?

rrpolanco (approved):

testgrid/tgrun/pkg/runner/vmi/embed/common.sh
emosbaugh: it has to wait until the initial primary is done with its tests otherwise the test will fail cause etcd will fall our of quorom if its a primary

PR #3247, opened 2022-08-05
Fix a directory naming issue

syntastical:
Fix the directory naming issue. After enabling a new pipeline, there was a typo in the directory name causing a failure.  This PR will fix that issue.

Fixes #

```release-note
NONE
```

NONE


marccampbell:
@syntastical mind updating the PR description with more context on this? 
What's the problem, how was this solution validated, is there an internal story or GitHub issue that reports the problem?  Adding context will really be helpful in the future when we see this commit in the history.

emosbaugh (approved):

PR #3268, opened 2022-08-09
Seanrester/sc 55220/add external addon support to kurl api

syntastical:
#### What type of PR is this?
Update api to pull external addon data and merge it with internal addons.

Fixes #55220

```release-note
NONE
```

NONE


.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -3,18 +3,7 @@
 1. If this is your first time, please read our contributor guidelines here:
 https://github.com/replicatedhq/kURL/blob/main/CONTRIBUTING.md.
 2. If the PR is unfinished, please mark it as a draft.
--->
```
e3b0c442, line 6: Collateral damage/rebase required?

web/src/installers/installer-versions.ts
```diff
@@ -4,14 +4,41 @@ import Bugsnag from "@bugsnag/js";
...
 
 const installerVersionsCache: { [url: string]: IInstallerVersions } = {};
+let externalAddons = {};
+let externalAddonTimer;
 
-export async function getInstallerVersions(distUrl: string, kurlVersion?: string): Promise<IInstallerVersions> {
+function mergeAddonVersions(internalAddonVersions: IInstallerVersions, kurlVersion?: string) {
+  const addons: IInstallerVersions = {};
+  Object.keys(externalAddons).forEach(externalAddonName => {
+    const fileName = externalAddonName.slice(0, externalAddonName.length - 7);
```
e3b0c442, line 21: What is the significance of magic number `7` here? Can we make that a constant somewhere so it's a bit more self-documenting?

web/src/installers/installer-versions.ts
```diff
@@ -4,14 +4,41 @@ import Bugsnag from "@bugsnag/js";
...
+    const [name, version] = fileName.split("-");
```
e3b0c442, line 22: Can we guarantee there are no dashes in any addon filename nor will ever be?

web/src/installers/installer-versions.ts
```diff
@@ -41,3 +68,25 @@ export async function getInstallerVersions(distUrl: string, kurlVersion?: string
   installerVersionsCache[url] = installerVersions;
   return installerVersionsCache[url];
 }
+
+async function externalAddonHandler() {
+  try {
+    const response = await fetch("https://kurl-sh.s3.amazonaws.com/external/addon-registry.json");
```
e3b0c442, line 74: Is this URL something we want to make configurable in case it needs to be changed in the future for some reason?

web/src/installers/installer-versions.ts
```diff
@@ -41,3 +68,25 @@ export async function getInstallerVersions(distUrl: string, kurlVersion?: string
...
+    externalAddons = await response.json();
+  } catch (error) {
+    logger.error(error, "failed to pull external addon registry.");
```
e3b0c442, line 77: Just making sure here, we're intending for this function to be a best-effort and just log an error?

web/src/commands/serve.ts
```diff
@@ -26,6 +27,7 @@ exports.handler = (argv: any) => {
 
 export async function main(argv: any): Promise<void> {
   sourceMapSupport.install();
+  await startExternalAddonPulling();
```
e3b0c442, line 30: Making sure I'm understanding here, we are intending to prevent startup from continuing until the poller has started and a first attempt has been made to pull the addons?

e3b0c442 (commented):
I've added some comments, some are likely just a lack of understanding on my part. I would like an application SME to review this too.

web/src/commands/serve.ts
syntastical: That's the thought, I feel like it would be bad if an instance went live after a deployment but all the external addons were unavailable.  I figure its better to wait for the application to become fully initialized.

web/src/installers/installer-versions.ts
syntastical: Stripping off ".tar.gz", I could either make a constant or a function to strip all file extensions.  Node's path API has functionality to get file name without extension but it considers the file name for "test.tar.gz" to be "test.tar"

web/src/installers/installer-versions.ts
syntastical: Since we're importing the addon, we can make this always the case but I would need to tweak the import process, which might be good.  Since we're only looking to do this for kots right now I've been keeping things more simplistic.

web/src/installers/installer-versions.ts
syntastical: I thought about it, but I figured we can cross that bridge if we ever come to it

web/src/installers/installer-versions.ts
syntastical: Yeah, I'm not 100% sure what to do here.  If the file isn't available that's reallllly bad but if the application already has data cached, it could already be current and i wouldn't want to fail a request because of that.

.github/PULL_REQUEST_TEMPLATE.md
laverya: No this looks like a good change
Not sure it should be integrated into this pr, but I'm happy to see the change itself

web/src/commands/serve.ts
laverya: can we do this after bugsnag and newrelic init?

web/src/controllers/InstallerAPI.ts
```diff
@@ -49,12 +49,6 @@ const unauthenticatedResponse = {
   },
 };
 
-const forbiddenResponse = {
-  error: {
-    message: "Forbidden",
-  },
-};
```
laverya, line 56: Why drop this?
(It's presumably not used anywhere?)

laverya (approved):

PR #3279, opened 2022-08-12
Use release git tag plus commit sha for image tags in KURL

jala-dx:
This PT attempts to `Use release git tag for image tags in KURL`

https://app.shortcut.com/replicated/story/55090/use-release-git-tag-for-image-tags-in-kurl

Yet to find a way to test this as this is a GH action change.
May be create a Nightly Tag and verify change after merging PR -- is one way I can think of.


emosbaugh (approved):

emosbaugh:
This will not work for staging as it reuses the version tag. You will have
to append the sha or change imagepullpolicy to always

PR #3291, opened 2022-08-16
Fix mismatched containerd versions between docker and installer spec

jala-dx:
When migrating from Docker to Containerd add-on, Docker is packaged with a higher version of Containerd. We must downgrade Containerd to the version specified as we package the corresponding version of the pause image. If we do not downgrade Containerd, Kubelet will fail to start in airgapped installations with pause image not found.

https://app.shortcut.com/replicated/story/55019/migration-to-containerd-airgap-fails

Fixes #

## Steps to reproduce
Run the Installer spec and the upgrade spec in the ticket

No

```release-note

Fixes a bug where Dockerd brings in a higher version of containerd than the one present in the Installer spec

```

No


scripts/common/host-packages.sh
```diff
@@ -172,6 +172,21 @@ EOF
         fi
     fi
     # shellcheck disable=SC2086
+    # When migrating from Docker to Containerd add-on, Docker is packaged with a higher version of
```
emosbaugh, line 175: Move the shellcheck comment after this new block of code

emosbaugh (changes requested):
Testgrid does not test changes in common.sh. How did you test this?

addons/containerd/1.5.11/install.sh
```diff
@@ -250,3 +250,75 @@ function _containerd_migrate_images_from_docker() {
         (set -x; ctr -n=k8s.io images import $image)
     done
 }
+function _yum_install_host_packages() {
```
jala-dx, line 253: this function will be removed before the commit. this is just to get the testgrid to run.

rrpolanco: I'll approve after this func is removed. Testgrid [runs](https://testgrid.kurl.sh/run/pr-3291-3c7091f-containerd-1.5.11-k8s-docker-2022-08-16T00:48:56Z ) for this PR looks satisfactory per our thread on slack.

rrpolanco (approved):

addons/containerd/1.5.11/install.sh
jala-dx: Added back this code to re-run testgrid, will remove before commit.

addons/containerd/1.5.11/install.sh
```diff
@@ -78,8 +78,8 @@ function containerd_install() {
         # If using the internal load balancer the Kubernetes API server will be unavailable until
         # kubelet starts the HAProxy static pod. This check ensures the Kubernetes API server
         # is available before proceeeding.
-        # ".." is needed becasue addons can have a CRD names "nodes", like nodes.longhorn.io
-        try_5m kubectl --kubeconfig=/etc/kubernetes/kubelet.conf get nodes..
+        # "nodes.v1." is needed becasue addons can have a CRD names "nodes", like nodes.longhorn.io ("lhn")
```
emosbaugh, line 81:
```suggestion
        # "nodes.v1." is needed becasue addons can have a CRD names "nodes", like nodes.longhorn.io
```

addons/containerd/1.5.11/install.sh
```diff
@@ -250,3 +250,77 @@ function _containerd_migrate_images_from_docker() {
         (set -x; ctr -n=k8s.io images import $image)
     done
 }
+function _yum_install_host_packages() {
```
emosbaugh, line 253: remove this

emosbaugh (changes requested):

addons/containerd/1.5.11/install.sh
jala-dx: removed

addons/containerd/1.5.11/install.sh
```diff
@@ -249,4 +249,4 @@ function _containerd_migrate_images_from_docker() {
     for image in $tmpdir/* ; do
         (set -x; ctr -n=k8s.io images import $image)
     done
-}
+}
```
emosbaugh, line 252: fix this

emosbaugh (changes requested):

emosbaugh (approved):

addons/containerd/1.5.11/install.sh
```diff
@@ -78,8 +78,8 @@ function containerd_install() {
         # If using the internal load balancer the Kubernetes API server will be unavailable until
         # kubelet starts the HAProxy static pod. This check ensures the Kubernetes API server
         # is available before proceeeding.
-        # ".." is needed becasue addons can have a CRD names "nodes", like nodes.longhorn.io
-        try_5m kubectl --kubeconfig=/etc/kubernetes/kubelet.conf get nodes..
+        # "nodes.v1." is needed becasue addons can have a CRD names "nodes", like nodes.longhorn.io
+        try_5m kubectl --kubeconfig=/etc/kubernetes/kubelet.conf get nodes.v1.
```
laverya, line 82: why has this not been changed in the addon template?

laverya: And why 1.5.11 but not 1.6, or the other 1.5, and so on?

jala-dx: yeah made a different PR to fix all the older versions and template for the upcoming versions https://github.com/replicatedhq/kURL/pull/3310/files#diff-8a7133184a76cd0383e373a9cff3a1b6c7c5f68233b702bc3ae0e101d2f003ffL81-L82

laverya: handled in https://github.com/replicatedhq/kURL/pull/3310

laverya (approved):

PR #3305, opened 2022-08-18
improve the reliability and output of rook 1.0 to 1.4 migration

laverya:

Fixes #

```release-note

```

### DIFF
--- a/scripts/common/rook.sh
+++ b/scripts/common/rook.sh
@@ -263,6 +263,19 @@ function rook_10_to_14() {
     curl -sSL https://raw.githubusercontent.com/rook/rook/v1.1.9/cluster/examples/kubernetes/ceph/upgrade-from-v1.0-apply.yaml \
       | sed 's/ROOK_SYSTEM_NAMESPACE/rook-ceph/g' | sed 's/ROOK_NAMESPACE/rook-ceph/g' | kubectl apply -f -
 
+    # change the default osd pool size from 3 to 1
+    kubectl apply -f - << EOM
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: rook-config-override
+  namespace: rook-ceph
+data:
+  config: |
+    [global]
+    osd pool default size = 1
+EOM
+
     kubectl delete crd volumesnapshotclasses.snapshot.storage.k8s.io volumesnapshotcontents.snapshot.storage.k8s.io volumesnapshots.snapshot.storage.k8s.io || true # resources may not be present
     kubectl -n rook-ceph set image deploy/rook-ceph-operator rook-ceph-operator=rook/ceph:v1.1.9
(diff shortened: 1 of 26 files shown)


scripts/common/rook.sh
```diff
@@ -323,14 +291,13 @@ function rook_10_to_14() {
     # todo store these files ourselves
     curl -sSL https://raw.githubusercontent.com/rook/rook/v1.1.9/cluster/examples/kubernetes/ceph/upgrade-from-v1.0-crds.yaml | kubectl apply -f -
 
-    # todo upgrade ceph image to 14.2.5 here
     echo "Upgrading ceph to v14.2.5"
     kubectl -n rook-ceph patch CephCluster rook-ceph --type=merge -p '{"spec": {"cephVersion": {"image": "ceph/ceph:v14.2.5-20191210"}}}'
-    spinner_until -1 is_ceph_rollout_complete "14.2.5"
+    $DIR/bin/kurl rook wait-for-ceph-version "14.2.5"
```
emosbaugh, line 296: when something goes wrong with upgrades it is most often the case that scripts fail waiting for ceph to roll out with a timeout of 10 minutes. should we do something similar here or have the user manually kill the script? i do not have a strong opinion either way.

scripts/common/rook.sh
```diff
@@ -399,9 +364,76 @@ EOM
     curl -sSL https://raw.githubusercontent.com/rook/rook/v1.4.9/cluster/examples/kubernetes/ceph/upgrade-from-v1.3-apply.yaml | kubectl apply -f -
     curl -sSL https://raw.githubusercontent.com/rook/rook/v1.4.9/cluster/examples/kubernetes/ceph/upgrade-from-v1.3-crds.yaml | kubectl apply -f -
     kubectl -n rook-ceph set image deploy/rook-ceph-operator rook-ceph-operator=rook/ceph:v1.4.9
-    kubectl -n rook-ceph set image deploy/rook-ceph-tools rook-ceph-tools=rook/ceph:v1.4.9
+    kubectl apply -f << EOM
```
emosbaugh, line 367: now that we have an add-on can we move this and other yaml to separate files?

emosbaugh (changes requested):

scripts/common/rook.sh
laverya: I can certainly see putting in a time limit, but I wasn't planning to until I had seen an issue.

scripts/common/rook.sh
laverya: Yep, I'm working on that at the moment. (along with "actually downloading the addon and loading images from it")

laverya: (I already have the yaml committed, I just don't have the way to access it yet)

emosbaugh (approved):

PR #3310, opened 2022-08-18
Fix name collisions nodes.v1.core with longhorn crd name 'nodes'

jala-dx:
Longhorn has its own CRD names like `nodes.longhorn.io`. And its short name is `nodes`.
This causes trouble listing `kubectl get nodes`. 
Using long hand name like "nodes.v1." fixes it.

Fixes #

```release-note

Fixed a bug where migration to containerd from docker will fail due to incorrectly listing node using incorrect Group. 

```

No

### DIFF
--- a/addons/containerd/1.3.7/install.sh
+++ b/addons/containerd/1.3.7/install.sh
@@ -78,8 +78,8 @@ function containerd_install() {
         # If using the internal load balancer the Kubernetes API server will be unavailable until
         # kubelet starts the HAProxy static pod. This check ensures the Kubernetes API server
         # is available before proceeeding.
-        # ".." is needed becasue addons can have a CRD names "nodes", like nodes.longhorn.io
-        try_5m kubectl --kubeconfig=/etc/kubernetes/kubelet.conf get nodes..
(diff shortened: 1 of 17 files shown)


rrpolanco:
I'm not sure why this would clash? Shouldn't Kubernetes api-discovery resolve the `core/v1` apigroup/version _first_ when evaluating resources? This seems like a bug in kube to me.

rrpolanco:
Also, how do you reproduce this? I have longhorn installed on a cluster with Kube 1.22 and I don't see this issue.

rrpolanco:
Relevant: 
- There's a proposal to enable `kubectl` to warn the user when there is a resource name/type conflict: https://github.com/kubernetes/kubernetes/pull/110218

laverya (approved):

jala-dx:
> Also, how do you reproduce this? I have longhorn installed on a cluster with Kube 1.22 and I don't see this issue.

Migrating to containerd causes it.

jala-dx:
> I'm not sure why this would clash. Shouldn't Kubernetes api-discovery resolve the `core/v1` apigroup/version _first_ when evaluating resources? This seems like a bug in kube to me.

Probably a bug on kube. and interestingly core is an empty string :)

PR #3315, opened 2022-08-19
block rook 10-to-14 upgrade on airgap

laverya:

Fixes #

```release-note

```

### DIFF
--- a/scripts/common/rook.sh
+++ b/scripts/common/rook.sh
@@ -205,6 +205,11 @@ function should_upgrade_rook_10_to_14() {
         return 1
     fi
 
+    # this upgrade process does not yet support airgap
+    if [ "$AIRGAP" = "1" ]; then
+        return 1
+    fi
+
     # rook is not currently installed, so no upgrade
     if ! is_rook_1 ; then
         return 1
--- a/scripts/tasks.sh
+++ b/scripts/tasks.sh
@@ -19,7 +19,7 @@ DIR=.
 . $DIR/scripts/distro/interface.sh
 . $DIR/scripts/distro/kubeadm/distro.sh
 . $DIR/scripts/distro/rke2/distro.sh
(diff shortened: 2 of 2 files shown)


scripts/common/rook.sh
```diff
@@ -205,6 +205,11 @@ function should_upgrade_rook_10_to_14() {
         return 1
     fi
 
+    # this upgrade process does not yet support airgap
+    if [ "$AIRGAP" = "1" ]; then
+        return 1
```
emosbaugh, line 210: can you reword the warning a bit?

```
        echo "Not upgrading rook because it is not installed or the currently installed version is not 1.0.x"
```

scripts/tasks.sh
```diff
@@ -19,7 +19,7 @@ DIR=.
 . $DIR/scripts/distro/interface.sh
 . $DIR/scripts/distro/kubeadm/distro.sh
 . $DIR/scripts/distro/rke2/distro.sh
-. $DIR/scripts/distro/rke2/addon.sh
```
emosbaugh, line 22: why are you removing this?

emosbaugh (changes requested):

scripts/tasks.sh
laverya: because I added it 20 minutes ago and meant to add this instead

laverya: `distro/rke2/addon.sh` doesn't even exist

emosbaugh: ok

scripts/common/rook.sh
laverya: no because this won't actually come up in tasks.sh yet - it doesn't parse the airgap flag and so will attempt to reach out to the internet

emosbaugh (approved):

PR #3341, opened 2022-08-23
Create new OpenEBS version 3.3.0

emosbaugh:

Create new OpenEBS version 3.3.0
Workaround [chart issue](https://github.com/openebs/charts/issues/309) installing version 3.3.0.

Fixes NONE

```release-note
Adds new [OpenEBS add-on](https://kurl.sh/docs/add-ons/openebs) version 3.3.0.
```

NONE

### DIFF
--- a/addons/openebs/template/generate.sh
+++ b/addons/openebs/template/generate.sh
@@ -23,6 +23,11 @@ function generate() {
         --set defaultStorageConfig.enabled=false \
         --set localprovisioner.enableDeviceClass=false \
         --set localprovisioner.enableHostpathClass=false \
+        --set release.version="$chart_version" \
+        --set localprovisioner.imageTag="$chart_version" \
+        --set helper.imageTag="$chart_version" \
+        --set ndm.imageTag=2.0.0 \
+        --set ndmOperator.imageTag=2.0.0 \
         > "$tmpdir/openebs.tmpl.yaml"
 
     $ksplit_path crdsplit "$tmpdir/"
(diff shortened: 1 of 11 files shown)


addons/openebs/template/generate.sh
```diff
@@ -23,6 +23,11 @@ function generate() {
         --set defaultStorageConfig.enabled=false \
         --set localprovisioner.enableDeviceClass=false \
         --set localprovisioner.enableHostpathClass=false \
+        --set release.version="$chart_version" \
```
emosbaugh, line 26: this hardcodes imageTags from README.md
https://github.com/openebs/charts/blob/openebs-3.3.0/charts/openebs/README.md?plain=1#L90

emosbaugh:
[sc-56120]

addons/openebs/template/generate.sh
laverya: do we want to hardcode images, though?
won't this prevent us from getting a new ndm image if one comes out in the future?

we're not regenerating old versions, so I don't see the problem with letting them float

rrpolanco (approved):

addons/openebs/template/generate.sh
emosbaugh: I am working around an upstream issue I created. They forgot to update the images for the new version of the chart.

https://github.com/openebs/charts/issues/309

laverya (approved):

PR #3349, opened 2022-08-23
test-addon-pr.yaml workflow broken calling reusable workflow

emosbaugh:

Fixes cannot run reusable workflow as step, only job.


e3b0c442 (approved):
`jq` is available in the runner images if things get any more complex re: building the JSON object in bash.

PR #3350, opened 2022-08-23
Output table rather than sarif for daily-scan-images workflow

emosbaugh:

Sarif results are inputs to github security code scanning alerts. Because there is no file and line number to associate with the images we are scanning, the alerts never go away and must be manually dismissed. Because of this limitation this pr will output the alerts in human readable format until we find a better platform.


emosbaugh:
[sc-54516]

laverya (approved):

PR #3367, opened 2022-08-26
fix race condition between deleting VMI PVCs and sending logs from the PVC

laverya:
also cleanup VMIs with all status conditions, not just running and completed

Fixes #

```release-note
NONE
```


laverya:
The PVC was being cleaned up after 3h, so running something that required that same pvc after 3h wasn't going to work

trelore (approved):
lgtm

PR #3383, opened 2022-08-29
include rook upgrade in main script

laverya:

Fixes #

First thing after this gets merged, this is going on testgrid

```release-note
Rook can now be upgraded from 1.0 to 1.4 during the install script.
```

### DIFF
--- a/scripts/common/rook.sh
+++ b/scripts/common/rook.sh
@@ -184,17 +184,16 @@ function current_rook_version() {
 # checks if rook should be upgraded before upgrading k8s. If it should, and the user confirms this, reports that as an addon, and starts the upgrade process.
 function maybe_report_upgrade_rook_10_to_14() {
     if should_upgrade_rook_10_to_14; then
-        printf "Upgrading Rook will take some time and will place additional load on your server.\n"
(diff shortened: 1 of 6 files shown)


scripts/common/rook.sh
```diff
@@ -181,9 +181,20 @@ function current_rook_version() {
...
+        if ! [ "$ROOK_BLOCK_STORAGE_ENABLED" = "1" ]; then # this check should probably be moved to golang (as a 'are there actually disks' check instead of a 'does the spec require disks' check
+            printf "In order to complete this migration, you will need to attach a blank disk to each node in the cluster for Rook to use.\n"
+        fi
+        printf "Would you like to continue? "
+
+        if ! confirmN; then
+            printf "Not upgrading Rook\n"
+            return 0
+        fi
+
```
emosbaugh, line 197:
```suggestion
        echo "Upgrading Rook will take some time and will place additional load on your server."
        if [ "$ROOK_BLOCK_STORAGE_ENABLED" != "1" ]; then # this check should probably be moved to golang (as a 'are there actually disks' check instead of a 'does the spec require disks' check
            echo "In order to complete this migration, you will need to attach a blank disk to each node in the cluster for Rook to use."
        fi
        printf "Would you like to continue? "

        if ! confirmN; then
            echo "Not upgrading Rook"
            return 0
        fi
```

emosbaugh (approved):
approved with a suggestion

emosbaugh (approved):

PR #3386, opened 2022-08-29
always validate rook scale when rerunning the installer

laverya:
When EKCO is not installed, the only opportunity we have to control the number of replicas for data in Rook is when running install.sh. For this to happen, that function needs to be called in the _already_applied function.

Fixes #

```release-note

```

### DIFF
--- a/addons/rook/template/base/install.sh
+++ b/addons/rook/template/base/install.sh
@@ -89,6 +89,7 @@ function rook_join() {
 
 function rook_already_applied() {
     rook_object_store_output
+    rook_set_ceph_pool_replicas
 }
 
 function rook_operator_crds_deploy() {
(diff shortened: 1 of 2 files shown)


addons/rook/template/base/install.sh
```diff
@@ -88,6 +88,7 @@ function rook_join() {
 }
 
 function rook_already_applied() {
+    rook_set_ceph_pool_replicas
```
emosbaugh, line 91: the lines are revered

emosbaugh (changes requested):

addons/rook/template/base/install.sh
laverya: 🤦

emosbaugh (approved):

rrpolanco (approved):

PR #3389, opened 2022-08-30
Make kurlkinds separate go module, go 1.19, no more controller-gen

emosbaugh:

I'm splitting out api and testgrid to separate repos. kurlkinds is used by both. I'd making it its own module so the other projects can pull in fewer dependencies.


.github/workflows/daily-testgrid.yaml
```diff
@@ -15,7 +15,7 @@ jobs:
     - name: setup-go
       uses: actions/setup-go@v3
       with:
-        go-version: 1.17
```
laverya, line 18: I swear that there's a way to have dependabot make these prs

emosbaugh:
[sc-53251]

laverya (approved):

PR #3404, opened 2022-08-31
Testgrid longhorn should use k8s 1.23.x

emosbaugh:

No need to pin longhorn to k8s latest (1.19.x) since that is only for rook.


rrpolanco (approved):

emosbaugh:
Tests are failing with kubernetes incompatibilities. Trying to find the highest versions possible to reduce failures.

laverya (approved):

rrpolanco (approved):

PR #3412, opened 2022-09-02
Preflight check for supported docker versions on ubuntu 22.04

jala-dx:
Ubuntu 22.04 does not support docker versions less than 20.10.17. 
Added Host built in preflight check so that the Install bails early on.

sc-57514 https://app.shortcut.com/replicated/story/57514/write-a-preflight-check-that-does-not-allow-any-docker-versions-less-than-20-10-17-on-ubuntu-22-04

Did basic testing. 
. Without docker version in the spec.
. With docker version in the spec.
. semvers like 20.01.06 (prefixed with 0's on Minor/Patch numbers

## Steps to reproduce
Install kurl with Docker version say 20.10.5 on Ubuntu 22.04 

No
```release-note
Added Host Preflight check for supported docker versions on Ubuntu 22.04
```

No

### DIFF
--- a/pkg/preflight/assets/host-preflights.yaml
+++ b/pkg/preflight/assets/host-preflights.yaml
@@ -394,3 +394,11 @@ spec:
           - fail:
               when: "ubuntu = 22.04"
               message: "Collectd addon not supported on ubuntu 22.04"
+    - hostOS:
+        checkName: "Docker Support"
+        exclude: '{{kurl and .Installer.Spec.Docker.Version (semverCompare "< 20.10.17" .Installer.Spec.Docker.Version) | not }}'
+        outcomes:
+          - fail:
+              when: "ubuntu = 22.04"
+              message: "Docker version < 20.10.17 not supported on ubuntu 22.04"
+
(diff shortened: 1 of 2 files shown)


pkg/preflight/assets/host-preflights.yaml
```diff
@@ -394,3 +394,11 @@ spec:
           - fail:
               when: "ubuntu = 22.04"
               message: "Collectd addon not supported on ubuntu 22.04"
+    - hostOS:
+        checkName: "Docker Support"
+        exclude: '{{kurl and .Installer.Spec.Docker.Version (semverCompare "< 20.10.17" .Installer.Spec.Docker.Version) | not }}'
```
emosbaugh, line 399: can you flip this so its as you would read it

"exclude if (docker version does not exist) or (docker version >= 20.10.17)"

jala-dx: Is this better? if () or () seems complicated.

emosbaugh (approved):
Thanks. This is what I was looking for

PR #3422, opened 2022-09-06
Validate version tag in CI and bail if necessary

jala-dx:
 If the Version tag is incorrect (say by accidentally pushing a wrong tag) or does not follow the convention of `<vYYYY.MM.DD-[0..n]>` ,  fail CI to avoid storing images with incorrect tags.

https://app.shortcut.com/replicated/story/55228/ci-should-fail-for-tags-with-dates-in-the-future

Verified this on private repo. Need to verify this during Release. I will do a kURL release to verify. 

## Steps to reproduce
Push a tag that has future or past dated. 

No

No

### DIFF
--- a/.github/workflows/deploy-prod.yaml
+++ b/.github/workflows/deploy-prod.yaml
@@ -6,6 +6,21 @@ on:
     - "v*.*.*"
 
 jobs:
+  verify-tag:
+    runs-on: ubuntu-20.04
+    outputs:
+        result: ${{ steps.verify-tag-id.outputs.result }}
+    steps:
+    - name: Checkout
+      uses: actions/checkout@v2
+
+    - id: verify-tag-id
+      name: Verify version tag
+      run: |
+          export VERSION_TAG=$(echo $GITHUB_REF | awk -F'/' '{ print $NF }')
+          OUTPUT=`bin/verify_tag.sh "${VERSION_TAG}"`
+          echo "::set-output name=result::$OUTPUT"
+
   generate-kurl-release-notes-pr:
     runs-on: ubuntu-20.04
     steps:
--- a/.github/workflows/deploy-prod.yaml
+++ b/.github/workflows/deploy-prod.yaml
@@ -18,7 +18,7 @@ jobs:
       name: Verify version tag
       run: |
           export VERSION_TAG=$(echo $GITHUB_REF | awk -F'/' '{ print $NF }')
-          OUTPUT=`bin/verify_tag.sh "${VERSION_TAG}"`
+          OUTPUT=`bin/verify-tag.sh "${VERSION_TAG}"`
           echo "::set-output name=result::$OUTPUT"
 
   generate-kurl-release-notes-pr:
--- a/.github/workflows/deploy-prod.yaml
+++ b/.github/workflows/deploy-prod.yaml
@@ -15,7 +15,7 @@ jobs:
     - name: Verify version tag
       run: |
         export VERSION_TAG=$(echo $GITHUB_REF | awk -F'/' '{ print $NF }')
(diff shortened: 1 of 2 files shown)


.github/workflows/deploy-prod.yaml
```diff
@@ -6,6 +6,21 @@ on:
...
+    steps:
+    - name: Checkout
+      uses: actions/checkout@v2
+
+    - id: verify-tag-id
+      name: Verify version tag
+      run: |
+          export VERSION_TAG=$(echo $GITHUB_REF | awk -F'/' '{ print $NF }')
+          OUTPUT=`bin/verify-tag.sh "${VERSION_TAG}"`
+          echo "::set-output name=result::$OUTPUT"
```
emosbaugh, line 22: no need to output here as you are not using this as an input https://docs.github.com/en/actions/using-jobs/defining-outputs-for-jobs

.github/workflows/deploy-prod.yaml
```diff
@@ -6,6 +6,21 @@ on:
     - "v*.*.*"
 
 jobs:
+  verify-tag:
+    runs-on: ubuntu-20.04
+    outputs:
+        result: ${{ steps.verify-tag-id.outputs.result }}
```
emosbaugh, line 12: no need to output here as you are not using this as an input https://docs.github.com/en/actions/using-jobs/defining-outputs-for-jobs

.github/workflows/deploy-prod.yaml
```diff
@@ -6,6 +6,21 @@ on:
     - "v*.*.*"
 
 jobs:
+  verify-tag:
```
emosbaugh, line 9: Can you make the other jobs depend (`needs`) this job so that the workflow will not start if this fails.

bin/verify-tag.sh
```diff
@@ -0,0 +1,26 @@
...
+    exit 1
+}          
+
+function main() {
+    VERSION_TAG=$1
+    tag_arr=(${VERSION_TAG//-/ })
+    if [ "${#tag_arr[@]}" != 2 ]; then
+        bail "Tag must have a release sequence, <vYYYY.MM.DD-[0..n]>"
+    fi
+    todays_date=$(date +'v%Y.%m.%d')
```
emosbaugh, line 20: make this utc today as we have [here](https://github.com/replicatedhq/kURL/blob/bef801b49c7b36e17d7d6f7ad52972bd2a70b1c3/bin/tag-and-release.sh#L72)

```suggestion
    todays_date=$(date -u +'v%Y.%m.%d')
```

emosbaugh (changes requested):

.github/workflows/deploy-prod.yaml
jala-dx: yes thanks. all of them depends on production-docker-image or kurl-util-image. Added in those.

.github/workflows/deploy-prod.yaml
```diff
@@ -6,8 +6,21 @@ on:
...
+    runs-on: ubuntu-20.04
+    steps:
+    - name: Checkout
+      uses: actions/checkout@v2
+
+    - name: Verify version tag
+      run: |
+        export VERSION_TAG=$(echo $GITHUB_REF | awk -F'/' '{ print $NF }')
+        OUTPUT=`bin/verify-tag.sh "${VERSION_TAG}"`
```
emosbaugh, line 18:
```suggestion
        ./bin/verify-tag.sh "${VERSION_TAG}"
```

emosbaugh (changes requested):

emosbaugh (approved):

PR #3423, opened 2022-09-07
Add cert-manager 1.9.1

jnewland:

This PR adds a version of cert-manager compatible with Kubernetes 1.20 → 1.24. The current version [has been EOL for more than a year](https://cert-manager.io/docs/installation/supported-releases/#old-releases) and is incompatible with supported versions of Kubernetes.

I was able to confirm that the latest cert-manager functions as intended  by running a Kurl installer without cert-manager immediately followed by [the single kubectl command from their docs](https://cert-manager.io/docs/installation/kubectl/#steps), so I adapted the existing 1.0.3 installer and added the combined YAML file from the kubectl instructions in this PR.

Related to [SC-50012](https://app.shortcut.com/replicated/story/50012/support-latest-available-cert-manager-1-9-x) (which I can't see as a non replicatedhq org member).

## Steps to reproduce

TODO

```release-note
Add cert-manager 1.9.1
```

TODO

### DIFF
--- a/addons/cert-manager/1.9.1/Manifest
+++ b/addons/cert-manager/1.9.1/Manifest
@@ -1,0 +1,5 @@
+image cainjector quay.io/jetstack/cert-manager-cainjector:v1.9.1
+image cert-manager-controller quay.io/jetstack/cert-manager-controller:v1.9.1
+image cert-manager-webhook quay.io/jetstack/cert-manager-webhook:v1.9.1
+
+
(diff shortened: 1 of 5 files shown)


laverya:
I'm going to test this out in a different PR - there's a few secret keys that CI needs to actually run things in testgrid, and I want to improve that test suite anyways.

(we also may need to hold off on making this 'latest' depending on whether applying this on top of 1.0.3 works or not - it looks plausible that we'll need to do https://cert-manager.io/docs/installation/upgrading/#reinstalling-cert-manager for those instances, but if so we can still get this out as a with some 'don't upgrade in broken ways' checks while that's in flight)

jnewland:
> I'm going to test this out in a different PR

@laverya 🙏 thank you, makes sense. Please keep me in the loop.

laverya:
Will do, I think I'm just going to go with the 'no supported upgrades from v1.0.3' route for the moment and figure out what's needed for upgrades later 😄

PR #3429, opened 2022-09-08
Allow upgrading certmanager 1.0 to 1.9

laverya:

Fixes #

```release-note
Allows upgrading cert-manager 1.0.3 to 1.9.1
```

### DIFF
--- a/addons/cert-manager/1.9.1/install.sh
+++ b/addons/cert-manager/1.9.1/install.sh
@@ -8,12 +8,16 @@ cert-manager() {
 
   # check if the current cert-manager version is v1.0.3 - if it is, skip the upgrade
   if kubectl get ns | grep -q cert-manager; then
-      local certManagerImage=
-      certManagerImage=$(kubectl get deployment -n cert-manager cert-manager -o jsonpath='{.spec.template.spec.containers[0].image}')
-      if [ "$certManagerImage" == "quay.io/jetstack/cert-manager-controller:v1.0.3" ]; then
-          printf "not upgrading cert-manager from v1.0.3 at this time\n"
-          return 0
-      fi
+    local certManagerImage=
+    certManagerImage=$(kubectl get deployment -n cert-manager cert-manager -o jsonpath='{.spec.template.spec.containers[0].image}')
+    if [ "$certManagerImage" == "quay.io/jetstack/cert-manager-controller:v1.0.3" ]; then
+      kubectl delete MutatingWebhookConfiguration cert-manager-webhook
+      kubectl delete namespace cert-manager
+
+      kubectl get issuer -A -o yaml > "$dst/existingIssuers.yaml"
+
+      kubectl delete crd certificaterequests.cert-manager.io certificates.cert-manager.io challenges.acme.cert-manager.io clusterissuers.cert-manager.io installers.cluster.kurl.sh issuers.cert-manager.io orders.acme.cert-manager.io || true
+    fi
   fi
 
   kubectl apply -k  "$dst"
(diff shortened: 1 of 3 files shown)


addons/cert-manager/1.9.1/install.sh
```diff
@@ -8,21 +8,30 @@ cert-manager() {
...
+
+      kubectl delete crd certificaterequests.cert-manager.io certificates.cert-manager.io challenges.acme.cert-manager.io clusterissuers.cert-manager.io installers.cluster.kurl.sh issuers.cert-manager.io orders.acme.cert-manager.io || true
+    fi
   fi
 
   kubectl apply -k  "$dst"
 
   # wait for deployments to be ready
-  printf "awaiting cert-manager deployment"
+  printf "awaiting cert-manager deployment\n"
```
emosbaugh, line 26: why not just change these to echo?

addons/cert-manager/1.9.1/install.sh
```diff
@@ -8,21 +8,30 @@ cert-manager() {
...
   spinner_until 120 deployment_fully_updated cert-manager cert-manager
-  printf "awaiting cert-manager-cainjector deployment"
+  printf "awaiting cert-manager-cainjector deployment\n"
   spinner_until 120 deployment_fully_updated cert-manager cert-manager-cainjector
-  printf "awaiting cert-manager-webhook deployment"
+  printf "awaiting cert-manager-webhook deployment\n"
   spinner_until 120 deployment_fully_updated cert-manager cert-manager-webhook
+
+  if [ -f "$dst/existingIssuers.yaml" ]; then
+    kubectl apply -f "$dst/existingIssuers.yaml" || true
```
emosbaugh, line 34: why ignore error here? can you at least justify with a comment?

addons/cert-manager/template/testgrid/k8s-docker.yaml
```diff
@@ -38,9 +38,12 @@
     sleep 30s
     kubectl get crds
     kubectl get all -n cert-manager
+    kubectl create -f https://raw.githubusercontent.com/cert-manager/website/master/content/docs/tutorials/acme/example/staging-issuer.yaml
+    kubectl get issuers
```
emosbaugh, line 42: i dont think this will ever return an error unless you get a specific issuer

addons/cert-manager/template/testgrid/k8s-docker.yaml
```diff
@@ -38,9 +38,12 @@
...
   postUpgradeScript: |
     # todo validate that cert-manager is actually working (TBD how)
     kubectl get pods -n cert-manager
+    kubectl get issuers
```
emosbaugh, line 46: same here

addons/cert-manager/template/testgrid/k8s-docker.yaml
```diff
@@ -38,9 +38,12 @@
     sleep 30s
```
emosbaugh, line 38: there really should be a mechanism that waits in the install script rather than here

emosbaugh (changes requested):

addons/cert-manager/template/testgrid/k8s-docker.yaml
laverya: right, I'll change it

addons/cert-manager/template/testgrid/k8s-docker.yaml
laverya: I didn't want to change the 1.0.3 installer, but I could probably add the same wait mechanism there

addons/cert-manager/1.9.1/install.sh
laverya: yeah fair enough

addons/cert-manager/1.9.1/install.sh
laverya: I am not 100% positive that this will work with every issuer that might exist (v1beta1, etc)

addons/cert-manager/template/testgrid/k8s-docker.yaml
laverya: I've changed the 1.0.3 installer now

emosbaugh (approved):

PR #3430, opened 2022-09-08
Deploy workflow to set current version in s3

emosbaugh:

Since we will be moving the kurl api to a separate repo we can no longer pin the version in the kurl release workflow. I will use this file in s3 to identify the current version from the api.

### DIFF
--- a/.github/workflows/deploy-prod.yaml
+++ b/.github/workflows/deploy-prod.yaml
@@ -297,11 +297,28 @@ jobs:
         upload_url: ${{ steps.create_release.outputs.upload_url }}
         asset_path: sbom/assets/*
 
+  set-current-version:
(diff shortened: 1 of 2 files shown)


.github/workflows/deploy-prod.yaml
```diff
@@ -297,6 +297,22 @@ jobs:
         upload_url: ${{ steps.create_release.outputs.upload_url }}
         asset_path: sbom/assets/*
 
+  set-current-version:
```
laverya, line 300: should we make testgrid-run depend on this?

emosbaugh: done

laverya (approved):

PR #3432, opened 2022-09-09
chore: fix testgrid for k8s 1.24 docker incompatibility

ricardomaraschini:

This PR makes a series of changes in the test grids:

- As Kubernetes 1.24.x does not support Docker, all references to `kubernetes:latest` are now using `containerd`
- Ensures that we never try to run OpenEBS 1.12 on Kubernetes newer than 1.21
- Ensures we never attempt to skip more than one version when upgrading Kubernetes
- Creates an extra test to verify if we can migrate from Docker to Containerd while moving also from Kubernetes `1.23` to `1.24`
- Ensures we never try to run Rook `1.0.x` in any Kubernetes newer than `1.19.x`

Fixes #54878

To review this one can be a little bit overwhelming but please take special care while doing it. The whole process of changing and verifying this is error prone therefore I passed through the whole test grid more than once.

```release-note
NONE
```

NONE

### DIFF
--- a/addons/openebs/template/testgrid/k8s-docker-localpv.yaml
+++ b/addons/openebs/template/testgrid/k8s-docker-localpv.yaml
@@ -24,7 +24,7 @@
 - name: localpv upgrade from 2.6.0
   installerSpec:
     kubernetes:
-      version: "1.21.x"
+      version: "1.23.x"
     weave:
       version: "latest"
     containerd:
@@ -104,7 +104,7 @@
       version: "1.12.0"
   upgradeSpec:
     kubernetes:
-      version: "1.24.x"
+      version: "1.21.x"
     weave:
       version: "latest"
     containerd:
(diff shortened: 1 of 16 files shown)


emosbaugh (dismissed):

addons/openebs/template/testgrid/k8s-docker-localpv.yaml
```diff
@@ -104,10 +104,10 @@
       version: "1.12.0"
   upgradeSpec:
     kubernetes:
-      version: "1.24.x"
+      version: "1.21.x"
     weave:
       version: "latest"
-    docker:
+    containerd:
```
laverya, line 110:
```suggestion
    docker:
```

because otherwise we're also testing docker->containerd here

emosbaugh (approved):

addons/openebs/template/testgrid/k8s-docker-localpv.yaml
ricardomaraschini: yeah, makes total sense.

laverya (approved):

PR #3445, opened 2022-09-12
cis: remove apiserver --insecure-port command line flag

ricardomaraschini:

From kubernetes v1.24 onwards the kubeapi flag --insecure-port has been removed. In CIS compliant environment this flag was in use, with it the apiserver does not come up.

This flag is a no-op since kubernetes [v1.20](https://github.com/kubernetes/kubeadm/issues/2156)

Fixes #54878

Please pay special attention to my shell script foo as it is quite rusty.

## Steps to reproduce

Install any cluster with kubernetes v1.24  and `cisCompliance` enabled.

```release-note
NONE
```

NONE

### DIFF
--- a/scripts/install.sh
+++ b/scripts/install.sh
@@ -115,9 +115,15 @@ function init() {
             $kustomize_kubeadm_init/kustomization.yaml \
             patch-kubelet-cis-compliance.yaml
         
-        insert_patches_strategic_merge \
-            $kustomize_kubeadm_init/kustomization.yaml \
-            patch-cluster-config-cis-compliance.yaml
+        if [ "$KUBERNETES_TARGET_VERSION_MINOR" -ge "20" ]; then
+            insert_patches_strategic_merge \
+                $kustomize_kubeadm_init/kustomization.yaml \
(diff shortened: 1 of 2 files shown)


laverya (approved):
honestly we should probably be splitting this into multiple files (one for insecure-port, only to be applied for old k8s, and the other for all the other settings that is applied every time)

But this will let us not break 1.20 and earlier in the future

scripts/install.sh
```diff
@@ -115,9 +115,15 @@ function init() {
...
             patch-kubelet-cis-compliance.yaml
         
-        insert_patches_strategic_merge \
-            $kustomize_kubeadm_init/kustomization.yaml \
-            patch-cluster-config-cis-compliance.yaml
+        if [ "$KUBERNETES_TARGET_VERSION_MINOR" -ge "20" ]; then
+            insert_patches_strategic_merge \
+                $kustomize_kubeadm_init/kustomization.yaml \
+                patch-cluster-config-cis-compliance.yaml
+	else
```
emosbaugh, line 122: Looks like the indentation isn't quite right here

PR #3451, opened 2022-09-13
fix: ignore PDB on single node upgrades

rrpolanco:
This PR adds the [disable-eviction](https://kubernetes.io/docs/reference/generated/kubectl/kubectl-commands#drain) flag to the `kubectl drain` command during a Kubernetes upgrade which allows longhorn instance-manager pods to be restarted and therefore allows statefulset pods to be recreated.

**Background**

When upgrading Kubernetes with the `longhorn` add-on installed, the existing `kubectl drain` command never completes because of the podDisruptionBudget for the `instance-manager` pods:
```
error when evicting pods/"instance-manager-r-78f51205" -n "longhorn-system" (will retry after 5s): Cannot evict pod as it would violate the pod's disruption budget.
evicting pod longhorn-system/instance-manager-e-8a5dd362
evicting pod longhorn-system/instance-manager-r-78f51205
There are pending pods in node "rafael-sc55652-k8s121-122-and-kots-upgrade" when an error occurred: [error when evicting pods/"instance-manager-e-8a5dd362" -n "longhorn-system": global timeout reached: 2m0s, error when evicting pods/"instance-manager-r-78f51205" -n "longhorn-system": global timeout reached: 2m0s]
pod/instance-manager-e-8a5dd362
pod/instance-manager-r-78f51205
error: unable to drain node "rafael-sc55652-k8s121-122-and-kots-upgrade", aborting command...
```

This caused statefulset pods such as `kotsadm-postgres` to never rollout successfully as the pod stayed stuck in `Terminating` status due to mount volume failures like the ones below:
```
MountVolume.WaitForAttach failed for volume "pvc-7d821e41-7c9f-46db-9498-387b47922cfb" : volume pvc-7d821e41-7c9f-46db-9498-387b47922cfb has GET error for volume attachment csi-3fd98854818e88981e49ebd22a7dbb054b5056e0123647ec1e044483e80ab5e6: volumeattachments.storage.k8s.io "csi-3fd98854818e88981e49ebd22a7dbb054b5056e0123647ec1e044483e80ab5e6" is forbidden: User "system:node:rafael-sc55652-k8s121-122-and-kots-upgrade" cannot get resource "volumeattachments" in API group "storage.k8s.io" at the cluster scope: no relationship found between node 'rafael-sc55652-k8s121-122-and-kots-upgrade' and this object
```

Fixes #
[sc-55652]

## Steps to reproduce

1. Install Kubernetes via this kURL spec: https://kurl.sh/8d91156
2. Deploy the app as described in https://app.shortcut.com/replicated/story/55652/kurl-upgrade-to-kots-1-79-0-fails-and-kills-the-application-too
3. Perform an upgrade of Kubernetes to `1.22.13` and kots to `1.79.0`.
4. See the kURL install script fail when waiting for `kotsadm` rollout to finish 
 
kURL upgrade spec:
```yaml
apiVersion: cluster.kurl.sh/v1beta1
kind: Installer
metadata:
  name: a11f243
spec:
  kubernetes:
    version: 1.22.13
  docker:
    version: 20.10.5
  weave:
    version: 2.6.5
  minio:
    version: 2020-01-25T02-50-51Z
  contour:
    version: 1.20.1
  registry:
    version: 2.7.1
  prometheus:
    version: 0.53.1-30.1.0
  kotsadm:
    applicationSlug: jama-k8s/cguzman-2-kots
    version: 1.79.0
  velero:
    version: 1.7.1
  ekco:
    version: 0.13.0
  kurl:
    additionalNoProxyAddresses: []
    installerVersion: v2022.09.08-1
  certManager:
    version: 1.9.1
  metricsServer:
    version: 0.3.7
  longhorn:
    version: 1.2.4
```

```release-note
NONE
```

NONE

### DIFF
--- a/scripts/common/kubernetes.sh
+++ b/scripts/common/kubernetes.sh
@@ -316,13 +316,26 @@ function spinner_kubernetes_api_stable() {
 }
 
 function kubernetes_drain() {
-    kubectl drain "$1" \
+    if kubernetes_has_remotes ; then
+      kubectl drain "$1" \
         --delete-local-data \
         --ignore-daemonsets \
         --force \
         --grace-period=30 \
         --timeout=120s \
-        --pod-selector 'app notin (rook-ceph-mon,rook-ceph-osd,rook-ceph-osd-prepare,rook-ceph-operator,rook-ceph-agent),k8s-app!=kube-dns' || true
+        --pod-selector 'app notin (rook-ceph-mon,rook-ceph-osd,rook-ceph-osd-prepare,rook-ceph-operator,rook-ceph-agent),k8s-app!=kube-dns, name notin (restic)' || true
+    else
+      # On single node installs ignore pod disruption budgets
+      kubectl drain "$1" \
+        --delete-local-data \
+        --ignore-daemonsets \
+        --force \
+        --grace-period=30 \
+        --timeout=120s \
+        --disable-eviction \
+        --pod-selector 'app notin (rook-ceph-mon,rook-ceph-osd,rook-ceph-osd-prepare,rook-ceph-operator,rook-ceph-agent),k8s-app!=kube-dns, name notin (restic)' || true
+
+    fi
 }
 
 function kubernetes_node_has_version() {
--- a/scripts/common/kubernetes.sh
+++ b/scripts/common/kubernetes.sh
@@ -316,26 +316,26 @@ function spinner_kubernetes_api_stable() {
 }
 
 function kubernetes_drain() {
-  if kubernetes_has_remotes ; then
-    kubectl drain "$1" \
-      --delete-local-data \
-      --ignore-daemonsets \
-      --force \
-      --grace-period=30 \
-      --timeout=120s \
-      --pod-selector 'app notin (rook-ceph-mon,rook-ceph-osd,rook-ceph-osd-prepare,rook-ceph-operator,rook-ceph-agent),k8s-app!=kube-dns, name notin (restic)' || true
-  else
-    # On single node installs force drain to delete pods or
-    # else the command will timeout when evicting pods with pod disruption budgets
-    kubectl drain "$1" \
-      --delete-local-data \
-      --ignore-daemonsets \
-      --force \
-      --grace-period=30 \
-      --timeout=120s \
-      --disable-eviction \
-      --pod-selector 'app notin (rook-ceph-mon,rook-ceph-osd,rook-ceph-osd-prepare,rook-ceph-operator,rook-ceph-agent),k8s-app!=kube-dns, name notin (restic)' || true
-  fi
+    if kubernetes_has_remotes ; then
+        kubectl drain "$1" \
+            --delete-local-data \
+            --ignore-daemonsets \
+            --force \
+            --grace-period=30 \
+            --timeout=120s \
+            --pod-selector 'app notin (rook-ceph-mon,rook-ceph-osd,rook-ceph-osd-prepare,rook-ceph-operator,rook-ceph-agent),k8s-app!=kube-dns, name notin (restic)' || true
+    else
+        # On single node installs force drain to delete pods or
+        # else the command will timeout when evicting pods with pod disruption budgets
+        kubectl drain "$1" \
+            --delete-local-data \
+            --ignore-daemonsets \
+            --force \
+            --grace-period=30 \
+            --timeout=120s \
+            --disable-eviction \
+            --pod-selector 'app notin (rook-ceph-mon,rook-ceph-osd,rook-ceph-osd-prepare,rook-ceph-operator,rook-ceph-agent),k8s-app!=kube-dns, name notin (restic)' || true
+    fi
 }
 
 function kubernetes_node_has_version() {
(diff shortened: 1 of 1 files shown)


scripts/common/kubernetes.sh
```diff
@@ -316,13 +316,26 @@ function spinner_kubernetes_api_stable() {
 }
 
 function kubernetes_drain() {
-    kubectl drain "$1" \
+    if kubernetes_has_remotes ; then
```
emosbaugh, line 319: could you simplify the code here and omit this conditional if you use `enable-deletion` in conjunction with `skip-wait-for-delete-timeout > 0`?

The flag is hard to understand but this post does a good job of explaining

https://github.com/rancher/system-upgrade-controller/discussions/148

> If you would prefer to honor PDB up to a maximum-ish number of seconds, then you should set your plan .spec.drain.skipWaitForDeleteTimeout.

emosbaugh: Actually I'm finding that when I'm upgrading kubernetes on 4 node cluster I'm still hitting PDBs when rolling the 4th node since the other 3 have just been upgraded. Eventually they do get drained but I still wonder if setting `skip-wait-for-delete-timeout > 0` as a precaution would be best?

```
...
evicting pod longhorn-system/instance-manager-r-4f2ebd0b
evicting pod longhorn-system/instance-manager-e-884b7b96
error when evicting pods/"instance-manager-e-884b7b96" -n "longhorn-system" (will retry after 5s): Cannot evict pod as it would violate the pod's disruption budget.
error when evicting pods/"instance-manager-r-4f2ebd0b" -n "longhorn-system" (will retry after 5s): Cannot evict pod as it would violate the pod's disruption budget.
evicting pod longhorn-system/instance-manager-r-4f2ebd0b
evicting pod longhorn-system/instance-manager-e-884b7b96
error when evicting pods/"instance-manager-e-884b7b96" -n "longhorn-system" (will retry after 5s): Cannot evict pod as it would violate the pod's disruption budget.
error when evicting pods/"instance-manager-r-4f2ebd0b" -n "longhorn-system" (will retry after 5s): Cannot evict pod as it would violate the pod's disruption budget.
evicting pod longhorn-system/instance-manager-e-884b7b96
evicting pod longhorn-system/instance-manager-r-4f2ebd0b
error when evicting pods/"instance-manager-r-4f2ebd0b" -n "longhorn-system" (will retry after 5s): Cannot evict pod as it would violate the pod's disruption budget.
error when evicting pods/"instance-manager-e-884b7b96" -n "longhorn-system" (will retry after 5s): Cannot evict pod as it would violate the pod's disruption budget.
evicting pod longhorn-system/instance-manager-e-884b7b96
evicting pod longhorn-system/instance-manager-r-4f2ebd0b
error when evicting pods/"instance-manager-r-4f2ebd0b" -n "longhorn-system" (will retry after 5s): Cannot evict pod as it would violate the pod's disruption budget.
error when evicting pods/"instance-manager-e-884b7b96" -n "longhorn-system" (will retry after 5s): Cannot evict pod as it would violate the pod's disruption budget.
evicting pod longhorn-system/instance-manager-e-884b7b96
evicting pod longhorn-system/instance-manager-r-4f2ebd0b
pod/instance-manager-r-4f2ebd0b evicted
pod/instance-manager-e-884b7b96 evicted
node/ethanm-migrate-4 drained
```

rrpolanco: > if you use enable-deletion

What is `enable-deletion`?

I understand what you're saying: we can still respect PDBs (for multi node scenarios) and fall back to deleting the pods using one command. I'll make this change and test it.

emosbaugh: Sorry I meant `disable-eviction`

rrpolanco: >skip-wait-for-delete-timeout > 0 as a precaution would be best

What's the benefit of adding this flag? The main benefit I see is that it ensures `kubectl drain` command completes instead of timing out.

emosbaugh: It will try normal eviction until the wait for delete timeout and then move on to delete. Evict is the normal drain behavior. I haven't looked into the difference between delete and evict. Do you think its ok to default to delete rather than evict?

https://stackoverflow.com/questions/62277852/whats-the-difference-between-pod-deletion-and-pod-eviction

rrpolanco: Deleting a pod _should_ be fine since they are backed by a controller. As I stated before, the benefit of falling back to normal delete behavior is to prevent `kubectl drain` command from timing out.

emosbaugh: In both cases it'll fall back to delete. It will fall back if it can't evict after some timeout. I think you may want to remove the conditional and always delete as I'm seeing issues (see the above comment) evicting longhorn pods when upgrading kubernetes as it requires us to roll the entire cluster.

emosbaugh (changes requested):
Sorry code looks good here but can you please fix the indentation? Thanks!

scripts/common/kubernetes.sh
emosbaugh: After discussing IRL we agreed that the original approach from @rrpolanco was best.

emosbaugh (approved):

PR #3452, opened 2022-09-13
Fix docker to containerd migration when upgrading to k8s 1.24

emosbaugh:

Fix kubelet fails to start with "Error: failed to parse kubelet flag: unknown flag: --network-plugin" when migrating from containerd to docker and upgrading kubernetes to 1.24.x.

Various other improvements around the migration from Docker to Containerd.

Fixes NONE

```release-note
- Fixes an issue that prevents upgrading Kubernetes to 1.24.x if the CRI has previously been migrated from Docker to Containerd.
```

NONE


scripts/upgrade.sh
```diff
@@ -116,10 +116,6 @@ function main() {
     download_util_binaries
     get_machine_id
     merge_yaml_specs
-    # Parse yaml into bash variables so we can get final kubernetes version for cases where we are upgrading 2 minor versions
```
emosbaugh, line 119: [This code](https://github.com/replicatedhq/kurl/commit/4d41a44e75ddc606eddc10553740dd77832b1060) is no longer necessary as it was fixed by [this commit](https://github.com/replicatedhq/kURL/commit/a6bcc0cbdc762e539134387475c4be330bc38843).

scripts/Manifest
```diff
@@ -8,5 +8,5 @@ REPLICATED_APP_URL=https://replicated.app
 KURL_UTIL_IMAGE=replicated/kurl-util:alpha
 KURL_BIN_UTILS_FILE=
 # STEP_VERSIONS array is generated by the server and injected at runtime based on supported k8s versions
-STEP_VERSIONS=("" "" "" "" "" "" "" "" "" "" "" "" "" "" "" "1.15.3" "1.16.4" "1.17.13" "1.18.20" "1.19.15", "1.20.11")
-INSTALLER_YAML=
\ No newline at end of file
+STEP_VERSIONS=(0.0.0 0.0.0 0.0.0 0.0.0 0.0.0 0.0.0 0.0.0 0.0.0 0.0.0 0.0.0 0.0.0 0.0.0 0.0.0 0.0.0 0.0.0 0.0.0 1.16.4 1.17.13 1.18.20 1.19.16 1.20.15 1.21.14 1.22.13 1.23.10 1.24.4)
```
laverya, line 11: I don't think this should have been updated

laverya (approved):

scripts/Manifest
emosbaugh: It's unrelated but it is helpful for the next developer working on upgrades of kubernetes.

rrpolanco: @emosbaugh Why do we need all the `0.0` versions?

emosbaugh: The relevant code is [here](https://github.com/replicatedhq/kURL/blob/7376cc9d7441f22b650d22209be9eec8ab223011/scripts/common/discover.sh#L247-L249). I needed to add the newer versions. This is placeholder for each minor version. 0.0.0 isn't necessary its just what I copied from production. See [here](view-source:https://k8s.kurl.sh/latest)

ricardomaraschini: @emosbaugh may be ok but: have you intentionally skipped 1.15.3 ?

ricardomaraschini (approved):

PR #3469, opened 2022-09-16
Fix upgrade install fails after partial install

emosbaugh:

After my initial install failed when installing rook I tried to re-run the installer. In between running the installer a second time a new k8s version was released. When i ran a second time my installer tried to upgrade k8s and failed with the following error since the kurl-config had not yet been written.

```
Drain local node and apply upgrade? (Y/n) y
...
Waiting for kubernetes api health to report ok
Error from server (NotFound): configmaps "kurl-config" not found
Error from server (NotFound): configmaps "kurl-config" not found
...
Error from server (NotFound): configmaps "kurl-config" not found
Error from server (NotFound): configmaps "kurl-config" not found
Error from server (NotFound): configmaps "kurl-config" not found
Kubernetes API failed to report healthy
```

Fixes NONE

```release-note
Fixes an edge case that causes Kubernetes upgrades to fail after an initial installation of kURL is attempted without running to completion.
```


emosbaugh:
I still need to test this scenario https://github.com/replicatedhq/kURL/commit/b3d4339dd20000450c5f634635b6a5a2205957a9

PR #3482, opened 2022-09-19
test rook-ceph 1.0 to 1.4 (or 1.5) upgrade in airgap

laverya:
This adds a daily testgrid test to make sure we don't inadvertently break airgap rook 1.0 to 1.4 upgrades. The run can be seen here: https://testgrid.kurl.sh/run/laverya%20rook%201.0.4%20to%201.4%20airgap%20nok8s%20allos.

This partially reincludes things reverted in https://github.com/replicatedhq/kURL/pull/3456.

Fixes #

```release-note
NONE
```

### DIFF
--- a/testgrid/specs/full.yaml
+++ b/testgrid/specs/full.yaml
@@ -1679,6 +1679,62 @@
 
     k8sVersion=$(kubectl get nodes -o jsonpath='{.items[0].status.nodeInfo.kubeletVersion}')
     echo $k8sVersion | grep 1.20
+- name: rook-upgrade-to-1.5-airgap
+  flags: "yes"
+  airgap: true
+  installerSpec:
+    kubernetes:
+      version: 1.19.x
+    containerd:
+      version: "latest"
+    weave:
+      version: "latest"
+    rook:
+      version: "1.0.4"
+    prometheus:
+      version: "latest"
+    minio:
+      version: "latest"
+  upgradeSpec:
+    kubernetes:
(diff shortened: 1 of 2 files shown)


testgrid/specs/full.yaml
```diff
@@ -1679,6 +1679,62 @@
...
+  postInstallScript: |
+    source /opt/kurl-testgrid/testhelpers.sh
+    rook_ecph_object_store_info
+    validate_read_write_object_store rwtest testfile.txt
+
+    minio_object_store_info
+    validate_read_write_object_store rwtest minio.txt
+  postUpgradeScript: |
+    source /opt/kurl-testgrid/testhelpers.sh
+    rook_ecph_object_store_info
```
rrpolanco, line 1725: Nit: Looks like there's a typo in the script name: `rook_ecph_...` instead of `rook_ceph_...`

rrpolanco (approved):

testgrid/specs/full.yaml
laverya: 1. yes, there is a typo in the name
2. unfortunately, I didn't catch that [before committing the script](https://github.com/replicatedhq/kURL-testgrid/blob/e34f686743c37f8a7fc6d5d6cc25a5af0298d733/tgrun/pkg/runner/vmi/embed/testhelpers.sh#L85) a few months ago and haven't found it worth it to fix since then

PR #3520, opened 2022-09-27
Fix import external addons action

emosbaugh:

Fix the action
Explicitly fetching external add-ons from s3Override rather than falling back on 404s

### DIFF
--- a/scripts/common/addon.sh
+++ b/scripts/common/addon.sh
@@ -62,7 +62,7 @@ function addon_fetch() {
     if [ "$AIRGAP" != "1" ]; then
         if [ -n "$s3Override" ]; then
             rm -rf $DIR/addons/$name/$version # Cleanup broken/incompatible addons from failed runs
-            addon_fetch_no_cache "$s3Override"
+            addon_fetch_cache "$name-$version.tar.gz" "$s3Override"
         elif [ -n "$DIST_URL" ]; then
             rm -rf $DIR/addons/$name/$version # Cleanup broken/incompatible addons from failed runs
             addon_fetch_cache "$name-$version.tar.gz"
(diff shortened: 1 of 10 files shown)


scripts/common/addon.sh
```diff
@@ -62,7 +62,7 @@ function addon_fetch() {
     if [ "$AIRGAP" != "1" ]; then
         if [ -n "$s3Override" ]; then
             rm -rf $DIR/addons/$name/$version # Cleanup broken/incompatible addons from failed runs
-            addon_fetch_no_cache "$s3Override"
+            addon_fetch_cache "$name-$version.tar.gz" "$s3Override"
```
emosbaugh, line 65: is it ok to fetch s3Override with cache?

scripts/common/common.sh
```diff
@@ -53,28 +41,31 @@ function package_download() {
...
+    checksum="$(grep -F "${package}" assets/Manifest | awk 'NR == 1 {print $3}')"
 
     if [ -n "${etag}" ] && ! package_matches_checksum "${package}" "${checksum}" ; then
         etag=
     fi
 
     local package_url=
-    package_url="$(get_dist_url)/${package}"
-
-    # if the package does not exist as a "normal" add-on, try the external add-on cache
```
emosbaugh, line 66: this was awkward

scripts/common/addon.sh
laverya: I think it's fine
we repull if the sha changed, right?

laverya (approved):

scripts/common/addon.sh
emosbaugh: ya. looks like we check the etag in s3

PR #3522, opened 2022-09-28
Document kURL's external add-on capability

emosbaugh:

Document kURL's external add-on capability


rrpolanco (approved):
I think it will be beneficial to document this as an ADR (retro ADR).

PR #3527, opened 2022-09-29
Allow image and asset Manifest assets as local files

emosbaugh:

Enable testing pull request and on merge builds of kots

### DIFF
--- a/bin/save-manifest-assets.sh
+++ b/bin/save-manifest-assets.sh
@@ -183,18 +183,28 @@ while read -r line || [ -n "$line" ]; do
 
     case "$kind" in
         image)
-            filename=$(echo $line | awk '{ print $2 }')
-            image=$(echo $line | awk '{ print $3 }')
-            try_5_times docker pull $image
-            mkdir -p $OUT_DIR/images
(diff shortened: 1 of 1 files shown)


bin/save-manifest-assets.sh
```diff
@@ -183,18 +183,28 @@ while read -r line || [ -n "$line" ]; do
...
-            url=$(echo $line | awk '{ print $3 }')
-            curl -fL -o "$OUT_DIR/assets/$filename" "$url"
+            mkdir -p "$OUT_DIR/assets"
+            filename=$(echo "$line" | awk '{ print $2 }')
+            asset=$(echo "$line" | awk '{ print $3 }')
+            # we support both http and local assets
+            if echo "$asset" | grep -q '^https://' ; then
+                curl -fL -o "$OUT_DIR/assets/$filename" "$asset"
+            else
+                cp "$asset" "$OUT_DIR/assets/$filename"
```
rrpolanco, line 206: Do we have an example of a local asset?

rrpolanco (approved):

bin/save-manifest-assets.sh
emosbaugh: no we dont use this but i will in the kots repo

PR #3541, opened 2022-10-05
Use default name for data pool

emosbaugh:


addons/rook/template/base/cluster/patches/filesystem-Json6902.yaml
```diff
@@ -1,4 +1,7 @@
 ---
+- op: replace
+  path: /spec/dataPools/0/name
+  value: data0
```
rrpolanco, line 4: In Rook 1.7, the dataPool name is `rook-shared-fs-data0` but in Rook 1.8 it is set to the string `replicated`. The value should match the Rook 1.7 name.

rrpolanco: The cephfs storageclass `parameters.pool` should match the name in Cephfilesystem CR `dataPools.name` or else you'll get the following error when using the storageclass:

```bash
# osd pools
kubectl exec deploy/rook-ceph-tools -n rook-ceph -- ceph osd lspools
1 device_health_metrics
2 replicapool
3 rook-ceph-store.rgw.control
4 rook-shared-fs-metadata
5 rook-ceph-store.rgw.meta
6 rook-shared-fs-replicated
7 rook-ceph-store.rgw.log
8 rook-ceph-store.rgw.buckets.index
9 rook-ceph-store.rgw.buckets.non-ec
10 .rgw.root
11 rook-ceph-store.rgw.buckets.data
```

PVC stays in pending status due to:
```
Warning  ProvisioningFailed    64s (x12 over 6m20s)  rook-ceph.cephfs.csi.ceph.com_csi-cephfsplugin-provisioner-64d744d7ff-5g2mp_18586b06-b2d9-4237-aeac-7e7aa64866e3  failed to provision volume with StorageClass "rook-cephfs": rpc error: code = Internal desc = rados: ret=-22, Invalid argument: "invalid pool layout 'rook-shared-fs-data0' -- need a valid data pool"
```

PR #3562, opened 2022-10-10
feat: adding kurl installer object lint capabilities

ricardomaraschini:

This PR introduces a single source of truth for Installer object validations. It provides a set of rego rules and a complete list of Add-Ons and respective supported versions, all centralised in rego policies and variables.

Fixes #60178

Pay special attention to the `regotests` directory as it contains inputs (on `installer:` property) and outputs (on the `output` property). In essence, given `installer` we expect to se `output` errors.

```release-note
NONE
```

NONE


kurlkinds/pkg/lint/rego/vars.rego
```diff
@@ -0,0 +1,585 @@
+package kurl.installer
+
+# known_versions holds all addons and their respective versions, latest
+# is kept on its own variable.
+default known_versions = {
```
emosbaugh, line 5: is there some way to auto-generate this? i think we will also need to integrate this with the automated add-on version updates.

kurlkinds/pkg/lint/rego/vars.rego
```diff
@@ -0,0 +1,585 @@
...
+	"k3s": {
+		"latest": "v1.23.3+k3s1",
+		"versions": [
+			"v1.23.3+k3s1",
+			"v1.19.7+k3s1"
+		]
+	},
+	"kotsadm": {
+		"latest": "1.86.2",
+		"versions": [
```
emosbaugh, line 170: how will this work with kots as an external add-on?

kurlkinds/pkg/lint/rego/vars.rego
emosbaugh: an alternative and i think what i would prefer is to query the [api](https://k8s.kurl.sh/installer) with `http.send` https://www.openpolicyagent.org/docs/latest/policy-reference/#http

emosbaugh (commented):
In general I like the direction this is headed. One comment

kurlkinds/pkg/lint/rego/vars.rego
ricardomaraschini: I was seeing this repo (with the changes) as the single source of truth and everything else would be, eventually, migrated to start using it (including the `/installer` endpoint). Wasn't this the idea ? I may have got it wrong.

As for doing an external request that would be fine but we are planning to use this validation also during the host-preflight checks and that would not work in an air-gapped environment (ie. we gonna need to think in a solution to make this work without internet connection as well).

kurlkinds/pkg/lint/rego/vars.rego
ricardomaraschini: What I have done so far is: if the add-on version is invalid but it has an s3override property set then it is considered to be valid.

emosbaugh (approved):
lgtm!

emosbaugh (approved):

PR #3569, opened 2022-10-12
Apply Prometheus Monitor and Import Ceph Grafana Dashboard

rrpolanco:

1. Enable Rook metrics collection when the Prometheus add-on is installed
2. Discovery Grafana Ceph Dashboard and auto import
3. Add post-init hook to addon lifecycle in the kURL installer script

Fixes #

**Ceph Grafana Dashboard**
The Ceph Grafana dashboard was taken from sysdig/promcat [repository](https://github.com/sysdiglabs/promcat-resources/blob/master/resources/ceph/include/ceph_grafana.json). It was then modified so that it can be imported into our Grafana instance.

screenshot of Ceph dashboard after import:
image

```release-note
Beginning with Rook 1.9.12 and later, Ceph metrics collection and a Ceph Grafana dashboard will be enabled when the Prometheus add-on is also installed.
```


scripts/common/monitoring.sh
```diff
@@ -0,0 +1,10 @@
+
+function maybe_apply_prometheus_monitor() {
+    if [ -n "$PROMETHEUS_VERSION" ]; then
+        if [ -n "$ROOK_VERSION" ]; then
+            echo "Applying Prometheus service monitor custom resource"
+            curl "https://raw.githubusercontent.com/rook/rook/v${ROOK_VERSION}/deploy/examples/monitoring/service-monitor.yaml" \
```
emosbaugh, line 6: We cant do this since we have to take into account airgap. instead you should add this to generate.sh and save in the add-on bundle. Also you should write this as a patch not apply it directly. I would follow [this](https://github.com/replicatedhq/kURL/blob/66491a41f4b6b1919e4f10c9fd8c73d1bb0078eb/addons/kotsadm/1.86.0/install.sh#L41-L43) pattern. That being said this code should be moved to the Rook add-on.

emosbaugh (changes requested):

scripts/common/monitoring.sh
emosbaugh: Good point. I would package it with prometheus.

emosbaugh (approved):

emosbaugh:
Would some documentation of this be appropriate here? https://kurl.sh/docs/add-ons/rook

PR #3581, opened 2022-10-15
ADR: Host Packages

emosbaugh:

Proposal

### DIFF
--- a/docs/arch/adr-005-host-packages.md
+++ b/docs/arch/adr-005-host-packages.md
@@ -29,6 +29,8 @@ If not the script will exit early with a friendly message to the user detailing 
 
 kURL will treat add-ons that are host packages themselves a bit differently (Containerd, Docker, Kubernetes Kubelet, Collectd).
 These will continue to be bundled along with the add-on archive.
+We will attempt to install these host packages from their hosted package repositories when possible for online installations.
+If not possible, we will install the package distributed in the add-on archive.
 Sub-dependencies will no longer be packaged along with the add-on archive and will be installed from remote repositories.
 
 We will roll this solution out for new host OS versions.
--- a/docs/arch/adr-005-host-packages.md
+++ b/docs/arch/adr-005-host-packages.md
@@ -1,0 +1,53 @@
+# ADR 5: Host Packages
+
+## Context
+
+Today, kURL bundles all host OS packages and their sub-dependencies in the add-on packages it generates and installs these packages in the end customer environment, rather than directly from the package repositories.
+
+For RPM based distributions, kURL sets up a local "kurl.local" repository and overrides the official OS mirrors for dependency resolution when installing these packages.
+Even worse, for APT based distributions, kURL uses the `dpkg` command rather than `apt`, forgoes dependency resolution, and force installs the packages, often breaking the dependency graph.
+
+This negatively affects the end customer, as dependency incompatibilites can result in broken OS packages, and often prevent upgrading the host OS when applying upgrades and security patches.
+Additionally, this is an unsupported means of installing, likely nullifying enterprise support contracts with the host OS provider.
+Finally, packaging an entirely new dependency chain for each host OS and version results in a significant amount of data (on the order of gigabytes) that must be downloaded and stored by the end customer each time they want to run a kURL install or upgrade.
+
+For us, the maintainers of kURL, this "feature" has its own set of challenges.
+Since host packages are targeted to OS and version, this requires additional work from the team to add support for each new OS and version.
+Additionally, as kURL is taking ownership of these OS packages, each supported OS must be added to the test matrix, adding significant time and cost to Testgrid.
+
+## Decision
+
+Rather than bundle host packages in the add-on archive, kURL will install the required packages using the official repos already available to the host.
+
+## Solution
+
+It is likely that the server already has access to official mirrors or satellite package repositories.
+If the packages are available in a repository, then kURL will install the packages from there.
+
+A set of preflights will be run at the start of installation to determine if any of the packages are not available.
+If not the script will exit early with a friendly message to the user detailing the missing packages, including a command to install these packages.
+
+kURL will treat add-ons that are host packages themselves a bit differently (Containerd, Docker, Kubernetes Kubelet, Collectd).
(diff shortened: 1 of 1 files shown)


docs/arch/adr-005-host-packages.md
```diff
@@ -0,0 +1,53 @@
...
+
+kURL will install the most up-to-date packages available rather than the ones available at the time the add-on archive was built.
+
+Significantly smaller add-on packages and airgap bundles, resulting in reduced download times for the end customer.
+
+Reduced testing surface area, resulting in savings of both time and cost from Testgrid.
+
+Airgapped environments must have access to package repositories or the end customer will have to manually install the required packages, possibly resulting in longer time to a live installation.
+
+Any unforseen challenges installing host packages from remote repositories in a secure environemnt.
```
ricardomaraschini, line 53:
```suggestion
Any unforseen challenges installing host packages from remote repositories in a secure environment.
```

docs/arch/adr-005-host-packages.md
```diff
@@ -0,0 +1,53 @@
...
+
+kURL will install packages from official mirrors rather than ones included in the add-on archive.
+
+kURL will install the most up-to-date packages available rather than the ones available at the time the add-on archive was built.
+
+Significantly smaller add-on packages and airgap bundles, resulting in reduced download times for the end customer.
+
+Reduced testing surface area, resulting in savings of both time and cost from Testgrid.
+
+Airgapped environments must have access to package repositories or the end customer will have to manually install the required packages, possibly resulting in longer time to a live installation.
```
rrpolanco, line 51: We should update our kURL airgap installation pre-req docs to mention that the customer will have to open IP addresses and ports in their corporate/gateway/network allowlist to enable the OS package manager to download required packages.

rrpolanco (dismissed):
Totally agree with this proposal.

docs/arch/adr-005-host-packages.md
emosbaugh: added a consequence below.

docs/arch/adr-005-host-packages.md
```diff
@@ -0,0 +1,54 @@
...
+
+A set of preflights will be run at the start of installation to determine if any of the packages are not available.
+If not the script will exit early with a friendly message to the user detailing the missing packages, including a command to install these packages.
+
+kURL will treat add-ons that are host packages themselves a bit differently (Containerd, Docker, Kubernetes Kubelet, Collectd).
+These will continue to be bundled along with the add-on archive.
+Sub-dependencies will no longer be packaged along with the add-on archive and will be installed from remote repositories.
+
+We will roll this solution out for new host OS versions.
+We will continue to support existing host OS versions as we did prior to this proposal.
```
laverya, line 35: will we ever change current host OS versions to use packages provided by the host?

Or will we just wait the 4 years or so for them all to drop out of support?

laverya (approved):

docs/arch/adr-005-host-packages.md
emosbaugh: I suppose we can add another adr in the future for dropping our old way of supporting host packages rather than address that here.

docs/arch/adr-005-host-packages.md
```diff
@@ -0,0 +1,54 @@
...
+
+It is likely that the server already has access to official mirrors or satellite package repositories.
+If the packages are available in a repository, then kURL will install the packages from there.
+
+A set of preflights will be run at the start of installation to determine if any of the packages are not available.
+If not the script will exit early with a friendly message to the user detailing the missing packages, including a command to install these packages.
+
+kURL will treat add-ons that are host packages themselves a bit differently (Containerd, Docker, Kubernetes Kubelet, Collectd).
+These will continue to be bundled along with the add-on archive.
+Sub-dependencies will no longer be packaged along with the add-on archive and will be installed from remote repositories.
```
chris-sanders, line 32: Can or should we check to see if these packages are available and use them from the host if they are? I like that we're using the packaging system for su-dependencies.

Have you considered if we're shipping these that we statically compile them and avoid any host level sub-dependency?

docs/arch/adr-005-host-packages.md
chris-sanders: Just a note Ubuntu versions tend to live on forever, you probably don't want to wait for them to 'roll out of support' :)

chris-sanders (commented):
Just one small comment, otherwise this looks as expected.

docs/arch/adr-005-host-packages.md
emosbaugh: It makes sense to check for online installations. I will add that.

If there are statically compiled versions out there we can consider using them but I would not want to maintain our own. I suggest a proposal in an alternative ADR.

emosbaugh: Added a couple sentences about checking the hosted repo first.

AmberAlston (approved):
Approval on the content. We still need to solicit feedback from internal folks and vendors before making the change.

jdewinne:
I fully approve this 👍 
But I'm also concerned how this will go in practice. Mainly airgap installations, where they will not have mirrors. I'm a little bit afraid that this will break those installs, and the end-customer will look at us to resolve. But I could also be wrong on this. On the other hand: If your environment is airgapped, I do expect the customer be at that enterprise level to do have a mirror in place 🤞

xavpaice:
I don't imagine any airgap installations will be able to install the OS without a package repository accessible by the server.

I think I would take this further still - surely we do not expect to support the OS packages themselves even for packages like docker and containerd. The supported versions for the customer's platform should be in the repo mirror already.  If the package and it's deps are in a supported repo (e.g. not Universe if it's Ubuntu) then we don't need to supply the package at all, merely make sure that the _supported_ combinations are available for the target OS.  That would narrow down the list of packages we absolutely need to include in the bundle.

emosbaugh:
Packages like docker and kubernetes are available via 3rd party repositories 

https://docs.docker.com/engine/install/ubuntu/

https://kubernetes.io/docs/setup/production-environment/tools/kubeadm/install-kubeadm/#installing-kubeadm-kubelet-and-kubectl

PR #3583, opened 2022-10-16
Add external add-on isPrerelease support

emosbaugh:

Right now, an external add-on version becomes available in on kurl.sh as soon as it is added to the versions.json file. However, there may be scenarios where a new version is published, but it is still a pre-release, so we would not want that showing up on kurl.sh.

Add support for an isPrerelease field that would allow filtering out particular add-on versions. This field would be managed by the external add-on.

Fixes NONE

Related pr https://github.com/replicatedhq/kURL-api/pull/44

```release-note
NONE
```

NONE

### DIFF
--- a/.github/actions/import-external-addons/index.js
+++ b/.github/actions/import-external-addons/index.js
@@ -2,7 +2,7 @@ import { getInput, info, setFailed } from '@actions/core'
 import { exec } from '@actions/exec'
 import { HttpClient } from '@actions/http-client';
 import fs from 'node:fs/promises';
-import { appendVersion, findVersion, generateChecksum } from './util.js';
+import { appendVersion, findVersion, generateChecksum, isVersionReleasing } from './util.js';
 
 const main = async () => {
   const debug = process.env.DEBUG;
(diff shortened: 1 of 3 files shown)


.github/actions/import-external-addons/index.js
```diff
@@ -45,12 +45,22 @@ const main = async () => {
...
       const existing = findVersion(addonRegistryKurl[addonName], next);
       if(existing) {
-        // At the moment we treat the version as immutable.
-        // In the future we could check the hash and overwrite.
-        info(`Skipping existing addon ${addonBundleName}.`);
+        if (isVersionReleasing(existing, next)) {
+          info(`Releasing addon ${addonBundleName}.`);
+  
+          // only allow changing isPrerelease property for now
+          existing.isPrerelease = false;
```
cbodonnell, line 56: Makes sense that we would only allow changing `isPrerelease`.  Do we want this to be one-way only (true -> false)? Or might is also be useful to allow the other way in case there's a need to undo a release (false -> true)?

cbodonnell (approved):

PR #3605, opened 2022-10-19
Testgrid use ekco latest with rook 1.9.x

emosbaugh:

Fixes https://testgrid.kurl.sh/run/PROD-daily-2022-10-19T05:34:01Z?kurlLogsInstanceId=sakacnaavcerzvbe&nodeId=sakacnaavcerzvbe-initialprimary#L302

```
Rook Pre-init: Rook 1.9.12 is only compatible with EKCO add-on version 0.23.0 and above.
Rook 1.9.12 will not be installed due to failed preflight checks.
```

### DIFF
--- a/testgrid/specs/full.yaml
+++ b/testgrid/specs/full.yaml
@@ -436,7 +436,7 @@
     kotsadm:
       version: 1.25.2
     ekco:
-      version: 0.7.0
+      version: latest
     containerd:
       version: 1.4.10
     certManager:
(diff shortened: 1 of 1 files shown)


testgrid/specs/full.yaml
```diff
@@ -589,6 +589,8 @@
       nameserver: 8.8.8.8
     collectd:
       version: v5
+    ekco:
+      version: latest
```
laverya, line 593: wait, does rook _require_ ekco now?

laverya (approved):

testgrid/specs/full.yaml
emosbaugh: no but it should

PR #3607, opened 2022-10-20
Ekco improvements

emosbaugh:

Be sure to delete bootstrap-lb pod as it can cause failures to reach the apiserver if it is not deleted.
Explicitly disable Rook and Contour operations if the add-ons are disabled.
Fixes an issue that causes the .kube/config to get removed on a Kubernetes upgrade.

Fixes NONE

```release-note
Fixes an issue that causes the .kube/config to get removed on a Kubernetes upgrade.
```

NONE

### DIFF
--- a/addons/ekco/template/base/install.sh
+++ b/addons/ekco/template/base/install.sh
@@ -337,7 +337,7 @@ function ekco_create_deployment() {
     cp "$src/rotate-certs-rbac.yaml" "$dst/rotate-certs-rbac.yaml"
 
     # is rook enabled
-    if kubectl get ns | grep -q rook-ceph; then
+    if [ -n "$ROOK_VERSION" ] && kubectl get ns rook-ceph >/dev/null 2>&1 ; then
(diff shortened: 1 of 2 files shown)


addons/ekco/template/base/install.sh
```diff
@@ -336,7 +337,7 @@ function ekco_create_deployment() {
     cp "$src/rotate-certs-rbac.yaml" "$dst/rotate-certs-rbac.yaml"
 
     # is rook enabled
-    if kubectl get ns | grep -q rook-ceph; then
+    if [ -n "$ROOK_VERSION" ]; then
```
emosbaugh, line 340: is there any drawback to changing this?

laverya: if someone used to have rook in the spec and so it's installed, but it's not in the most recent spec?

laverya (approved):

rrpolanco (approved):

laverya (dismissed):

PR #3674, opened 2022-11-04
feat: checking for disk space before migrating

ricardomaraschini:

pvmigrate now checks if there is enough disk space before starting to migrate volumes.

#### How this (roughly) works 

1. Determine if the destination storage is OpenEBS local volume or Rook
2. If neither of them, move on to the pvmigrate, no disk free verification
3. If any of them, instantiates a specific entity to execute the disk free check

For OpenEBS as destination:

1. Sums up the total of storage reserved by PVs per node and, possibly, space reserved by "detached" PVs.
2. "Per node" is determined by inspecting where the pod that mounts the PVC is running.
3. Spawns a pod in each of the "schedulable" nodes.
4. Each pod mounts (HostPath) the BasePath directory from the StorageClass configuration and calculates the free space through a call to `df`.
5. If the BasePath is part of the root filesystem only 85% of the disk is taken into account, otherwise 100%.
6. Verifies that each node has enough space to host the the PVs "running" on them.
7. Verifies that all nodes has enough space to host the "detached" PVs (we don't know where they are going to be migrated).
8. If there is not enough space in any node, bails out.

For Rook as destination:

1. Sums up the total of storage reserved by PVs per node and, possibly, space reserved by "detached" PVs.
2. Gets the Ceph cluster and CephPool from the StorageClass parameters.
3. Queries through Ceph CRDs to find out how much space is available.
4. Divides the total Ceph space by the number of replicas.
5. Verifies that the sum of the space reserved by all PVs per node and detached PVs fits into the space.

Fixes #61205

```release-note
PVMigrate now checks for available disk space before starting to migrate volumes.
```


kurl_util/cmd/pvmigrate/main.go
```diff
@@ -1,15 +1,124 @@
...
+	clusterspace "github.com/replicatedhq/kurl/pkg/cluster/space"
 	"github.com/replicatedhq/kurl/pkg/version"
 	"github.com/replicatedhq/pvmigrate/pkg/migrate"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/rest"
+	"sigs.k8s.io/controller-runtime/pkg/client/config"
 )
 
 func main() {
```
camilamacedo86, line 20: Should that to be a standalone command or subcommand of kurl?
When/who will call this command?

pkg/cluster/space/openebs.go
```diff
@@ -0,0 +1,414 @@
...
+		free, used, err := o.parseDFContainerOutput(out["df"])
+		if err != nil {
+			o.logPodInfo(out, status)
+			return nil, fmt.Errorf("unable to parse pod output: %w", err)
+		}
+
+		volumes, err := o.parseFstabContainerOutput(out["fstab"])
+		if err != nil {
+			o.logPodInfo(out, status)
+			return nil, fmt.Errorf("unable to parse node %s fstab: %w", node.Name, err)
```
emosbaugh, line 140:
```suggestion
			return nil, fmt.Errorf("failed to parse node %s fstab output: %w", node.Name, err)
```

emosbaugh (changes requested):
Some feedback in comments.

Also, I'd like to see some unit tests.

rrpolanco (commented):
Overall I think this looks good. However, I'm wondering if some or most of this could have been implemented in the core of [pvmigrate](https://github.com/replicatedhq/pvmigrate) compared to here (the client).

pkg/k8sutil/storage.go
```diff
@@ -0,0 +1,47 @@
...
+	return pvs, nil
+}
+
+// PVCSForPVs returns a pv to pvc mapping. the returned map is indexed by the pv name.
+func PVCSForPVs(ctx context.Context, cli kubernetes.Interface, pvs map[string]corev1.PersistentVolume) (map[string]corev1.PersistentVolumeClaim, error) {
+	pvcs := map[string]corev1.PersistentVolumeClaim{}
+	for pvidx, pv := range pvs {
+		cref := pv.Spec.ClaimRef
+		if cref == nil {
+			return nil, fmt.Errorf("pv %s without associated PVC", pvidx)
```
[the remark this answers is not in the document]
ricardomaraschini: According to pvmigrate [documentation](https://github.com/replicatedhq/pvmigrate#known-limitations) this causes an error: "PVs without associated PVCs are not handled, and will cause pvmigrate to exit with an error"

pkg/cluster/space/openebs.go
```diff
@@ -0,0 +1,414 @@
...
+					VolumeMounts: []corev1.VolumeMount{
+						{
+							MountPath: "/data",
+							Name:      "openebs",
+							ReadOnly:  true,
+						},
+					},
+					Resources: corev1.ResourceRequirements{
+						Requests: corev1.ResourceList{
+							corev1.ResourceMemory: resource.MustParse("32Mi"),
```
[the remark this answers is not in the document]
ricardomaraschini: Yes, no need. Removed.

pkg/cluster/space/openebs.go
```diff
@@ -0,0 +1,414 @@
...
+					VolumeMounts: []corev1.VolumeMount{
+						{
+							MountPath: "/node/etc/fstab",
+							Name:      "fstab",
+							ReadOnly:  true,
+						},
+					},
+					Resources: corev1.ResourceRequirements{
+						Requests: corev1.ResourceList{
+							corev1.ResourceMemory: resource.MustParse("32Mi"),
```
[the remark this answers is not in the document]
ricardomaraschini: Fixed.

pkg/k8sutil/pod.go
```diff
@@ -0,0 +1,83 @@
...
+func RunEphemeralPod(ctx context.Context, cli kubernetes.Interface, logger *log.Logger, timeout time.Duration, pod *corev1.Pod) (map[string][]byte, *corev1.PodStatus, error) {
+	podidx := fmt.Sprintf("%s/%s", pod.Namespace, pod.Name)
+	pod, err := cli.CoreV1().Pods(pod.Namespace).Create(ctx, pod, metav1.CreateOptions{})
+	if err != nil {
+		return nil, nil, fmt.Errorf("error creating pod %s: %w", podidx, err)
+	}
+
+	defer func() {
+		if err = cli.CoreV1().Pods(pod.Namespace).Delete(
+			ctx, pod.Name, metav1.DeleteOptions{},
```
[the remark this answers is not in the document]
ricardomaraschini: nice catch! fixed.

pkg/cluster/space/kutils.go
```diff
@@ -0,0 +1,100 @@
...
+
+// K8SUtils provides tooling for common operations used by multiple storage free checkers.
+type K8SUtils struct {
+	log *log.Logger
+	cli kubernetes.Interface
+	cfg *rest.Config
+}
+
+// PodUsesPVC returs true if provided pod has provided pvc among its volumes.
+func (k *K8SUtils) PodUsesPVC(pod corev1.Pod, pvc corev1.PersistentVolumeClaim) bool {
```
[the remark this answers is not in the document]
ricardomaraschini: Agreed, moved it to `k8sutil` package.

pkg/cluster/space/openebs.go
```diff
@@ -0,0 +1,414 @@
...
+	RootVolume bool
+}
+
+// openEBSVolumes attempts to gather the free and used disk space for the openebs volume in
+// all nodes in the cluster. this function creates a temporary pod in each of the nodes of
+// the cluster, the pod runs a "df" command and we parse its output.
+func (o *OpenEBSChecker) openEBSVolumes(ctx context.Context) (map[string]OpenEBSVolume, error) {
+	nodes, err := o.cli.CoreV1().Nodes().List(ctx, metav1.ListOptions{})
+	if err != nil {
+		return nil, fmt.Errorf("unable to list nodes: %w", err)
```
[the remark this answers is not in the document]
ricardomaraschini: fixed in all errors.

pkg/k8sutil/pod.go
```diff
@@ -0,0 +1,83 @@
...
+)
+
+// RunEphemeralPod starts provided pod and waits until it finishes. the values returned are the
+// pod logs, its last status, and an error. returned logs and pod status may be nil depending on
+// the type of failure. the pod is deleted at the end.
+func RunEphemeralPod(ctx context.Context, cli kubernetes.Interface, logger *log.Logger, timeout time.Duration, pod *corev1.Pod) (map[string][]byte, *corev1.PodStatus, error) {
+	podidx := fmt.Sprintf("%s/%s", pod.Namespace, pod.Name)
+	pod, err := cli.CoreV1().Pods(pod.Namespace).Create(ctx, pod, metav1.CreateOptions{})
+	if err != nil {
+		return nil, nil, fmt.Errorf("error creating pod %s: %w", podidx, err)
```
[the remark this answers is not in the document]
ricardomaraschini: Sounds good. Fixed.

pkg/k8sutil/deploy.go
```diff
@@ -0,0 +1,50 @@
...
+	"fmt"
+
+	appsv1 "k8s.io/api/apps/v1"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes"
+)
+
+// DeploymentPods return a list of pods for provided deployment.
+func DeploymentPods(ctx context.Context, cli kubernetes.Interface, ns, depname string) ([]corev1.Pod, error) {
```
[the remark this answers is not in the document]
ricardomaraschini: Yes, there is no reason to select by replicaset. I don't remember where I was with my head. Fixed.

pkg/cluster/space/kutils.go
```diff
@@ -0,0 +1,100 @@
...
+		return nil, 0, fmt.Errorf("unable to get pvcs for pvs: %w", err)
+	}
+
+	var detached int64
+	var attached = map[string]int64{}
+	for pvidx, pvc := range pvcs {
+		pv, ok := pvs[pvidx]
+		if !ok {
+			pvcidx := fmt.Sprintf("%s/%s", pvc.Namespace, pvc.Name)
+			return nil, 0, fmt.Errorf("pv for pvc %s not found", pvcidx)
```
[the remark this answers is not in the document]
ricardomaraschini: fixed.

pkg/k8sutil/pod.go
```diff
@@ -0,0 +1,83 @@
...
+	for {
+		var gotPod *corev1.Pod
+		if gotPod, err = cli.CoreV1().Pods(pod.Namespace).Get(
+			ctx, pod.Name, metav1.GetOptions{},
+		); err != nil {
+			return nil, nil, fmt.Errorf("failed getting pod %s: %w", podidx, err)
+		}
+
+		lastPodStatus = gotPod.Status
+		if gotPod.Status.Phase == corev1.PodSucceeded {
```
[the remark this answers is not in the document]
ricardomaraschini: Makes total sense. Fixed.

pkg/cluster/space/kutils.go
```diff
@@ -0,0 +1,100 @@
...
+		}
+
+		pods, err := k.cli.CoreV1().Pods(pvc.Namespace).List(ctx, metav1.ListOptions{})
```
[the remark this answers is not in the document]
ricardomaraschini: fixed.

pkg/k8sutil/pod.go
```diff
@@ -0,0 +1,83 @@
...
+
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes"
+)
+
+// RunEphemeralPod starts provided pod and waits until it finishes. the values returned are the
+// pod logs, its last status, and an error. returned logs and pod status may be nil depending on
+// the type of failure. the pod is deleted at the end.
+func RunEphemeralPod(ctx context.Context, cli kubernetes.Interface, logger *log.Logger, timeout time.Duration, pod *corev1.Pod) (map[string][]byte, *corev1.PodStatus, error) {
```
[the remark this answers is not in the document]
ricardomaraschini: Migrated to a job.

kurl_util/cmd/pvmigrate/main.go
```diff
@@ -1,15 +1,133 @@
...
+	}
+
+	if err = migrate.Migrate(ctx, logger, cli, opts); err != nil {
+		fmt.Printf("%s\n", err.Error())
+		os.Exit(1)
+	}
+}
+
+func checkFreeSpace(ctx context.Context, logger *log.Logger, cfg *rest.Config, cli kubernetes.Interface, opts migrate.Options) {
+	logger.Printf("Preparing to run migrations...")
```
laverya, line 64: Could we make this more explicitly storage space related? Since this isn't the migration process yet

laverya:
```suggestion
	logger.Printf("Checking if there is enough space to complete the storage migration")
```

kurl_util/cmd/pvmigrate/main.go
```diff
@@ -1,15 +1,133 @@
...
+		checkFreeSpace(ctx, logger, cfg, cli, opts)
+	}
+
+	if err = migrate.Migrate(ctx, logger, cli, opts); err != nil {
+		fmt.Printf("%s\n", err.Error())
+		os.Exit(1)
+	}
+}
+
+func checkFreeSpace(ctx context.Context, logger *log.Logger, cfg *rest.Config, cli kubernetes.Interface, opts migrate.Options) {
```
laverya, line 63: Why does this not return error? There's a lot of `os.Exit(1)` in it and that feels weird to me

kurl_util/cmd/pvmigrate/main.go
```diff
@@ -1,15 +1,133 @@
...
+		if err != nil {
+			logger.Printf("Failed to check nodes free space: %s", err)
+			os.Exit(1)
+		}
+
+		if len(nodesWithoutSpace) == 0 {
+			return
+		}
+
+		logger.Print("Some nodes do not have enough disk space for the migration:")
```
laverya, line 104: can we print the nodes that do not have disk space?

kurl_util/cmd/pvmigrate/main.go
```diff
@@ -1,15 +1,133 @@
...
 	"github.com/replicatedhq/kurl/pkg/version"
 	"github.com/replicatedhq/pvmigrate/pkg/migrate"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/rest"
+	"sigs.k8s.io/controller-runtime/pkg/client/config"
 )
 
 func main() {
+	ctx, stop := signal.NotifyContext(context.Background(), os.Interrupt)
```
camilamacedo86, line 22: Why are we not creating a subcommand for this one? (i.e https://github.com/replicatedhq/kURL/blob/main/pkg/cli/commands.go#L5)

kurl_util/cmd/pvmigrate/main.go
```diff
@@ -1,15 +1,133 @@
...
+			dstProvisioner = sc.Provisioner
+		}
+	}
+
+	// we skip free space check if any of the provided storage classes don't exist.
+	if srcProvisioner == "" || dstProvisioner == "" {
+		return
+	}
+
+	if dstProvisioner == "openebs.io/local" {
```
camilamacedo86, line 87: wdyt about to use const?

pkg/k8sutil/pod.go
```diff
@@ -0,0 +1,24 @@
+package k8sutil
+
+import (
+	corev1 "k8s.io/api/core/v1"
+)
+
+// PodUsesPVC returs true if provided pod has provided pvc among its volumes.
+func PodUsesPVC(pod corev1.Pod, pvc corev1.PersistentVolumeClaim) bool {
```
camilamacedo86, line 8: Nit: funcs that return a bool usually has as prefix has or is. wdyt about rename this one?

pkg/k8sutil/pod_test.go
```diff
@@ -0,0 +1,132 @@
...
+
+func TestPodUsesPVC(t *testing.T) {
+	for _, tt := range []struct {
+		name     string
+		expected bool
+		pod      corev1.Pod
+		pvc      corev1.PersistentVolumeClaim
+	}{
+		{
+			name:     "uses pvc",
```
camilamacedo86, line 18: wdyt about the name of the tests describes what should test and the expected result?
(i.e should pass when uses pvc and has enough disk space)

pkg/k8sutil/storage.go
```diff
@@ -0,0 +1,113 @@
...
+	"fmt"
+
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes"
+)
+
+// PVSByStorageClass returns a map of persistent volumes using the provided storage class name.
+// returned pvs map is indexed by pv's name.
+func PVSByStorageClass(ctx context.Context, cli kubernetes.Interface, scname string) (map[string]corev1.PersistentVolume, error) {
```
camilamacedo86, line 14: wdyt about we have a struct to store the data collect?
would not make it easier to manipulated instead of have these funcs that return a lot of values with maps?

pkg/cluster/space/rook.go
```diff
@@ -0,0 +1,154 @@
...
+	kcli  kubernetes.Interface
+	rcli  rookcli.Interface
+	cfg   *rest.Config
+	log   *log.Logger
+	srcSC string
+	dstSC string
+}
+
+// freeSpace attempts to get the ceph free space. returns the number of available bytes.
+func (r *RookChecker) freeSpace(ctx context.Context) (int64, error) {
```
camilamacedo86, line 32: wdyt about use the prefix get to make clear what the func does? 
```suggestion
func (r *RookChecker) getFreSpaceWith(ctx context.Context) (int64, error) {
```

pkg/cluster/space/rook.go
```diff
@@ -0,0 +1,154 @@
...
+	}
+
+	if pname == "" || cname == "" {
+		return "", "", fmt.Errorf("failed to read storage class %s pool/cluster", r.dstSC)
+	}
+	return pname, cname, nil
+}
+
+// Check verifies if there is enough ceph disk space to migrate from the source storage class.
+func (r *RookChecker) Check(ctx context.Context) (bool, error) {
```
camilamacedo86, line 104:
```suggestion
func (r *RookChecker) HasEnoughDiskSpace(ctx context.Context) (bool, error) {
```

pkg/cluster/space/rook.go
```diff
@@ -0,0 +1,154 @@
...
+
+	"github.com/replicatedhq/kurl/pkg/k8sutil"
+)
+
+const (
+	namespace = "rook-ceph"
+)
+
+// RookChecker checks if we have enough disk space to migrate volumes to rook.
+type RookChecker struct {
```
camilamacedo86, line 22:
```suggestion
type RockDiskSpaceValidator struct {
```

wdyt about something like ^

pkg/cluster/space/openebs.go
```diff
@@ -0,0 +1,599 @@
...
+		}
+
+		// pos is now the position where the actual used space is.
+		pos = len(words) - 4
+		usedBytes, err := strconv.ParseInt(words[pos], 10, 64)
+		if err != nil {
+			return 0, 0, fmt.Errorf("failed to parse %q as used space: %w", words[pos], err)
+		}
+
+		return freeBytes, usedBytes, nil
```
laverya, line 76: I would REALLY prefer a regex-based implementation here (with capture groups)

Regexes may be brittle and inscrutable, but so's using word order directly

pkg/cluster/space/openebs.go
```diff
@@ -0,0 +1,599 @@
...
+	for node, vol := range volumes {
+		var ok bool
+		var free int64
+		if free, ok = o.hasEnoughSpace(vol, reservedPerNode[node]); ok {
+			continue
+		}
+
+		faultyNodes[node] = true
+		o.log.Printf(
+			"Node %q has %s available, failed to migrate %s (reserved in %q storage class)",
```
laverya, line 522:
```suggestion
			"Node %q has %s available, which is less than the %s that would be migrated from the %q storage class)",
```

pkg/cluster/space/rook.go
```diff
@@ -0,0 +1,154 @@
...
+		return 0, fmt.Errorf("failed to read ceph status (nil)")
+	}
+
+	availint64 := int64(cluster.Status.CephStatus.Capacity.AvailableBytes)
+	replicasint64 := int64(pool.Spec.Replicated.Size)
+	return availint64 / replicasint64, nil
+}
+
+// reservedSpace returns the total size of all volumes using the source storage class (srcSC).
+func (r *RookChecker) reservedSpace(ctx context.Context) (int64, error) {
```
laverya, line 63: This feels like a function that shouldn't be in the rook checker specifically - should it go to k8sutil?

(there's probably one in openebs that does the same for a storageclass on a node too)

pkg/k8sutil/storage.go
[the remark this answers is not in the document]
laverya: Yeah I kinda went "look, if this comes up for real in the wild, I'll special case it. I don't think it's happening though" on that logic

pkg/cluster/space/openebs.go
emosbaugh: On the other hand it's pretty common to parse the output of linux command line with awk

kurl_util/cmd/pvmigrate/main.go
ricardomaraschini: fixed.

kurl_util/cmd/pvmigrate/main.go
ricardomaraschini: fixed.

kurl_util/cmd/pvmigrate/main.go
ricardomaraschini: fixed.

pkg/k8sutil/pod.go
ricardomaraschini: No problem, would you mind suggesting a name for it ?

pkg/k8sutil/pod_test.go
ricardomaraschini: fixed.

pkg/cluster/space/rook.go
ricardomaraschini: I don't like it. `clusterspace.RookDiskSpaceValidator` does look too repetitive.

pkg/cluster/space/openebs.go
ricardomaraschini: fixed.

pkg/cluster/space/rook.go
ricardomaraschini: There is a single function that is called by both: `k8sutil.PVSReservationPerNode()`. Rook simply sums the result while OpenEBS takes node names into account. Not sure if this change would bring us any improvements.

pkg/cluster/space/rook.go
ricardomaraschini: I don't like to use prefix `get` in these scenarios. This seems more like a java than a go thing. I haven't understood the `With` as well, can you elaborate ?

pkg/cluster/space/rook.go
ricardomaraschini: fixed

pkg/cluster/space/rook.go
camilamacedo86: Will this check check anything else?
What about only DiskSpaceValidator or DiskSpaceChecker?

My idea here were we have a more meaningful name about what it actually does.

pkg/k8sutil/pod.go
camilamacedo86: By following the convention any fun that returns a bool usually is named with HasSomething or IsSomething 
So, I'd recommend here HasPVC wdyt? Would that make sense?

kurl_util/cmd/pvmigrate/main.go
camilamacedo86: Have we not the kurl CLI? How this main will be called?
How/where it will be used?

ricardomaraschini: It is called by some of our add-ons like [here](https://github.com/replicatedhq/kURL/blob/aea79861716d66787f0b31670f1fc74a7ee16d1f/scripts/common/rook.sh#L144-L154). This is a wrap of the this [project](https://github.com/replicatedhq/pvmigrate). We just execute a couple of checks prior to calling the `pvmigrate`.

pkg/k8sutil/pod.go
ricardomaraschini: Yeah, I understand. I think solely `HasPVC` lacks a little bit of context like "what" has a PVC. Maybe `PodHasPVC` ? But then why not `Uses` instead of `Has` (it makes more sense to me).

pkg/cluster/space/rook.go
ricardomaraschini: I tend to prefer the idiomatic Go, like [here](https://go.dev/doc/effective_go#Getters). That being said, the function has been renamed as suggested.

pkg/cluster/space/rook.go
ricardomaraschini: fixed.

pkg/k8sutil/pod.go
ricardomaraschini: function renamed to `HasPVC`.

pkg/k8sutil/storage.go
ricardomaraschini: I don't see how this would bring us any advantage tbh. A map from name to value is adequate to what we are trying to achieve.

camilamacedo86: By looking here I thought that could be something like 
load data and store in a temp struct and then after we could check out 

However, that would not be a blocker and is more a matter of preference. So. I am closing this one.

kurl_util/cmd/pvmigrate/main.go
camilamacedo86: I see I think we never would like that someone called it directly.
So, I think we can close this one.

pkg/cluster/space/openebs.go
```diff
@@ -0,0 +1,599 @@
...
+	for _, pvc := range pvcs {
+		propagation := metav1.DeletePropagationForeground
+		delopts := metav1.DeleteOptions{PropagationPolicy: &propagation}
+		if err := o.kcli.CoreV1().PersistentVolumeClaims("default").Delete(
+			ctx, pvc.Name, delopts,
+		); err != nil {
+			if errors.IsNotFound(err) {
+				continue
+			}
+			return fmt.Errorf("failed to delete temp pvc %s: %w", pvc.Name, err)
```
emosbaugh, line 171: should failing to delete one pvc fail all? im pretty sensitive to tasks that dont do a good job of cleaning up after themselves

emosbaugh (changes requested):

pkg/cluster/space/rook.go
camilamacedo86: Hi @ricardomaraschini,

> I tend to prefer the idiomatic Go, like [here](https://go.dev/doc/effective_go#Getters). That being said, the function has been renamed as suggested. 

But in this case we have struct.Spec(). Yes, in this case we do need get/set. 
However, in the code here we have an func. So, ihmo both are not the same scenario.

pkg/k8sutil/job.go
```diff
@@ -0,0 +1,109 @@
...
+	lastContainerStatuses := map[string]corev1.ContainerState{}
+	for _, status := range jobPod.Status.ContainerStatuses {
+		lastContainerStatuses[status.Name] = status.State
+	}
+
+	logs := map[string][]byte{}
+	for _, container := range jobPod.Spec.Containers {
+		options := &corev1.PodLogOptions{Container: container.Name}
+		plogs, err := cli.CoreV1().Pods(jobPod.Namespace).GetLogs(jobPod.Name, options).Stream(ctx)
+		if err != nil {
```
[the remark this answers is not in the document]
ricardomaraschini: fixed.

pkg/cluster/space/openebs.go
```diff
@@ -0,0 +1,599 @@
...
+		// pos is now the position where the actual used space is.
+		pos = len(words) - 4
+		usedBytes, err := strconv.ParseInt(words[pos], 10, 64)
+		if err != nil {
+			return 0, 0, fmt.Errorf("failed to parse %q as used space: %w", words[pos], err)
+		}
+
+		return freeBytes, usedBytes, nil
+	}
+	return 0, 0, fmt.Errorf("failed to locate free space info in pod log: %s", string(output))
```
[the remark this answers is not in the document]
ricardomaraschini: fixed.

pkg/cluster/space/openebs.go
```diff
@@ -0,0 +1,599 @@
...
+		}
+
+		if _, ok := seen[words[1]]; ok {
+			continue
+		}
+
+		seen[words[1]] = true
+		mounts = append(mounts, words[1])
+	}
+	if len(mounts) == 0 {
```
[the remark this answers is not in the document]
ricardomaraschini: fixed.

pkg/cluster/space/openebs.go
```diff
@@ -0,0 +1,599 @@
...
+
+	basePath, err := o.basePath(ctx)
+	if err != nil {
+		return nil, fmt.Errorf("failed to read openebs base path: %w", err)
+	}
+
+	var tmpPVCs []*corev1.PersistentVolumeClaim
+	defer func() {
+		o.log.Printf("Deleting temporary pvcs")
+		if err := o.deleteTmpPVCs(ctx, tmpPVCs); err != nil {
```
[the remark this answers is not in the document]
ricardomaraschini: fixed

pkg/cluster/space/openebs.go
ricardomaraschini: fixed.

pkg/cluster/space/rook.go
camilamacedo86: Hi @ricardomaraschini,

> I tend to prefer the idiomatic Go, like [here](https://go.dev/doc/effective_go#Getters). That being said, the function has been renamed as suggested.

Thank you for the reply, and allow me to clarify my thoughts, understanding and motivations on this one.

The suggestion here does not go against ` idiomatic Go` concept. See in the doc(link shared): 
`and **it's often appropriate to do so**, but it's neither idiomatic nor necessary to put Get into the getter's name`. 

Therefore, we do not need to implement getter/setter for all as that usually is done in JAVA, as you point out, for example. _(that does not see fit well with the stack either)_. Then, when it is implicit that it is a get, it is not required to use Get. Therefore, that also does not mean never using GET as a prefix. It shows means use when it is required only for readability.  

So, in this case,  we do not have `validator.Validator()` such as in the doc example, where we know what it is returning. We have validator.freeSpace() then: is that returning the free space amount OR doing an operation to ensure a freeSpace? That was my motivation for the suggestion.

pkg/k8sutil/job.go
```diff
@@ -0,0 +1,113 @@
...
+			message := fmt.Sprintf("failed to get container %s logs: %s", container.Name, err)
+			logger.Print(message)
+			logs[container.Name] = []byte(message)
+			continue
```
emosbaugh, line 92: arent logs critical if the job worked? i was worried in the case that the job failed due to an error that made the pod not runnable. in this case logs would always fail. should you only continue if `!worked`?

ricardomaraschini: fixed.

emosbaugh (approved):

PR #3680, opened 2022-11-05
(Test PR -> Do not merge) WIP - test: ci/githubactions: enable debug mode to analyse issues

camilamacedo86:

The ci has been failing and it is test to check if we would be able to better understand the issues. 

Fixes #

```release-note

```

### DIFF
--- a/.github/workflows/pr-license-scan.yaml
+++ b/.github/workflows/pr-license-scan.yaml
@@ -8,10 +8,15 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v3
-      - name: "Run FOSSA Analyze" 
+      - name: "Run FOSSA Scan"
         uses: fossas/fossa-action@main
         with:
           api-key: 1d5511d63d5a4ccda6d3cef2a7ef08c1 # push-only token, safe to expose
+          debug: true
+      - uses: actions/upload-artifact@v3
+        with:
+          name: fossa.debug.json.gz
+          path: ./fossa.debug.json.gz
       - name: "Run FOSSA Test"
         uses: fossas/fossa-action@main
         with:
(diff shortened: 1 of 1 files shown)


emosbaugh:
Adding @e3b0c442 as a reviewer to this one as he maintains that action

.github/workflows/pr-license-scan.yaml
```diff
@@ -8,12 +8,18 @@ jobs:
...
+      - uses: actions/upload-artifact@v3
+        with:
+          name: fossa.debug.json.gz
+          path: ./fossa.debug.json.gz
       - name: "Run FOSSA Test"
         uses: fossas/fossa-action@main
         with:
           api-key: 1d5511d63d5a4ccda6d3cef2a7ef08c1 # push-only token, safe to expose
           run-tests: true
+          debug: true
```
camilamacedo86, line 25:
```suggestion
          debug: true
                - uses: actions/upload-artifact@v3
        with:
          name: fossa.debug.json.gz
          path: ./fossa.debug.json.gz
```

.github/workflows/pr-license-scan.yaml
```diff
@@ -8,12 +8,22 @@ jobs:
...
+                - uses: actions/upload-artifact@v3
```
camilamacedo86, line 26:
```suggestion
      - uses: actions/upload-artifact@v3
```

e3b0c442:
I am already working directly with the service provider and they have what they need at this point. 

There will be a PR Monday morning which will overhaul this and centralize the action so that we can get adjust these types of things globally. 

As far as existing failures, many are login issues. This is clear in the action log, and rebasing on main should address these, though where possible if we can wait until things are centralized on Monday we can avoid more PRs on this than needed.

camilamacedo86:
Hi @e3b0c442,

Thx, I just create this PR see if I could enable the debug so that I could check what is wrong and why it is not passing.
However, I understand that you are working in this need and how that works will change for all projects.  So, I will wait for all changes to see if after all I still be unable to not check why/what is the reason for the failure and etc. No rush at all.

Agan, thank you for your explanation and time.

PR #3681, opened 2022-11-05
Fix build-test-success is skipped on failure

emosbaugh:

Branch protections rules are not working because required check build-test-success is skipped on a failure of it's needs.

https://github.com/replicatedhq/kURL/actions/runs/2967656145/jobs/4750952929

### DIFF
--- a/.github/workflows/build-test.yaml
+++ b/.github/workflows/build-test.yaml
@@ -38,13 +38,6 @@ jobs:
 
       - run: make -C kurl_util deps test build
 
-  test-shell:
-    runs-on: ubuntu-20.04
-    steps:
-    - uses: actions/checkout@v3
-    - run: sudo apt-get install -y shunit2
-    - run: make test-shell shunit2
-
   build-bin-kurl:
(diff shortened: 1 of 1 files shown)


.github/workflows/build-test.yaml
```diff
@@ -62,12 +55,29 @@ jobs:
 
       - run: make deps test build/bin/kurl
 
+  test-shell:
```
camilamacedo86, line 58: So, the test-shell must occurs after the the build-bin-kurl?
If so, should we not add here something like `needs: build-bin-kurl`?

camilamacedo86 (approved):
Hi @emosbaugh,
 
I am not sure if I follow up in why we need to check the result of the tests.Theoretically, if we run `command A` then it fails the command would be exit with 1 which would make the ci/test fails automatically.

However, I see that we have the [build-test-success](https://github.com/replicatedhq/kURL/blob/main/.github/workflows/build-test.yaml#L65-L73) already. So, it seems required for how that has been made and the changes here shows fine.

/lgtm

PR #3693, opened 2022-11-09
scale down ekco before running rgw to minio migration

laverya:

this avoids a potential race condition where both the rgw-to-minio and minio-to-ha-minio migrations could run at the same time

Fixes #

```release-note
NONE
```

### DIFF
--- a/scripts/common/object_store.sh
+++ b/scripts/common/object_store.sh
@@ -69,6 +69,12 @@ function object_store_bucket_exists() {
 function migrate_rgw_to_minio() {
     report_addon_start "rook-ceph-to-minio" "v1"
 
+    if kubernetes_resource_exists kurl deployment ekc-operator; then
+        kubectl -n kurl scale deploy ekc-operator --replicas=0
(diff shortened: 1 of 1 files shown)


scripts/common/object_store.sh
```diff
@@ -117,6 +123,11 @@ EOF
     spinner_until 300 kubernetes_pod_completed sync-object-store default || true
     kubectl logs -f sync-object-store || true
 
+    # even if the migration failed, we should ensure ekco is running again
+    if kubernetes_resource_exists kurl deployment ekc-operator; then
+        kubectl -n kurl scale deploy ekc-operator --replicas=1
+    fi
+
```
camilamacedo86, line 130: Wdyt about check it is running again before move forward ?

laverya: there's no need - we don't use it during the minio migration process (or generally any of the kurl processes)

it's more for all the cluster management stuff that comes up during times that we're not running an installer, if that makes any sense

emosbaugh (approved):

PR #3702, opened 2022-11-11
weave-to-flannel ADR

laverya:

Fixes #

```release-note

```

### DIFF
--- a/docs/arch/adr-008-weave-to-flannel-migration.md
+++ b/docs/arch/adr-008-weave-to-flannel-migration.md
@@ -1,0 +1,39 @@
+# ADR 8: Weave to Flannel Migration
+
+## Context
+
+As proposed in ADR 6, the future CNI of kURL will be Flannel.
+However, almost all existing installations are running Weave.
+
+A migration path must be provided for those instances.
+
+There were two general proposals for a migration path.
+One was to use Multus-CNI to allow using both Flannel and Weave at the same time before stripping out Weave, allowing an upgrade while only taking down one node at once.
+The other was to run a migration that removed Weave from all instances simultaneously before installing Flannel.
+This would be far simpler, not require an additional IP range, and not require Multus to remain in the cluster going forwards.
+It would also require 'stopping the world' and cause downtime.
+
+## Decision
+
+Provide a stop-the-world migration path to move from Weave to Flannel.
+
+## Solution
+
+Create a script to do the following:
+
+1. synchronize execution on all cluster nodes to minimize downtime
+2. Uninstall weave, removing iptables rules and host files
+3. Edit kubeadm.conf on all primary nodes and run kubeadm init
+4. Install flannel
+5. Stop kubelet, flush iptables, restart containerd and start kubelet
+6. wait for nodes to be ready, and then recreate cluster-networking pods (kube-system, then CSI, then everything else)
+
+This script would be able to be called explicitly from `tasks.sh` or after opt-in from the user as part of the kURL upgrade process.
+
+## Status
+
+Proposed
+
+## Consequences
+
+Users will be able to move from Weave to Flannel, but downtime will need to be scheduled to do so.
--- a/docs/arch/adr-008-weave-to-flannel-migration.md
+++ b/docs/arch/adr-008-weave-to-flannel-migration.md
@@ -40,5 +40,6 @@ Proposed
 
 Users will be able to move from Weave to Flannel, but downtime will need to be scheduled to do so.
 
-If a migration failed, manual remediation would be required.
+A failed migration may result in a cluster down scenario.
+Manual remediation would be required.
 If failures were frequent (either in testing or after deployment) an automatic remediation/rollback script could be created.
--- a/docs/arch/adr-008-weave-to-flannel-migration.md
+++ b/docs/arch/adr-008-weave-to-flannel-migration.md
@@ -12,6 +12,7 @@ One was to use Multus-CNI to allow using both Flannel and Weave at the same time
 The other was to run a migration that removed Weave from all instances simultaneously before installing Flannel.
 This would be far simpler, not require an additional IP range, and not require Multus to remain in the cluster going forwards.
 It would also require 'stopping the world' and cause downtime.
+In testing, this downtime was under 5 minutes when running commands manuall on a 3 node cluster.
 
 ## Decision
 
@@ -21,12 +22,13 @@ Provide a stop-the-world migration path to move from Weave to Flannel.
 
 Create a script to do the following:
 
-1. synchronize execution on all cluster nodes to minimize downtime
+1. Synchronize execution on all cluster nodes to minimize downtime
 2. Uninstall weave, removing iptables rules and host files
-3. Edit kubeadm.conf on all primary nodes and run kubeadm init
+3. Edit kubeadm.conf on all primary nodes and run kubeadm init (or potentially edit config files directly)
 4. Install flannel
 5. Stop kubelet, flush iptables, restart containerd and start kubelet
-6. wait for nodes to be ready, and then recreate cluster-networking pods (kube-system, then CSI, then everything else)
+6. Wait for nodes to be ready, and then recreate cluster-networking pods (kube-system, then CSI, then everything else)
+7. Run an inter-container (and inter-node) networking check
 
 This script would be able to be called explicitly from `tasks.sh` or after opt-in from the user as part of the kURL upgrade process.
 
@@ -37,3 +39,6 @@ Proposed
 ## Consequences
 
 Users will be able to move from Weave to Flannel, but downtime will need to be scheduled to do so.
+
+If a migration failed, manual remediation would be required.
+If failures were frequent (either in testing or after deployment) an automatic remediation/rollback script could be created.


docs/arch/adr-008-weave-to-flannel-migration.md
```diff
@@ -0,0 +1,39 @@
...
+This would be far simpler, not require an additional IP range, and not require Multus to remain in the cluster going forwards.
+It would also require 'stopping the world' and cause downtime.
+
+## Decision
+
+Provide a stop-the-world migration path to move from Weave to Flannel.
+
+## Solution
+
+Create a script to do the following:
```
emosbaugh, line 22: If the script failed at any given point would it be able to roll back? If not perhaps that belongs in the consequences

docs/arch/adr-008-weave-to-flannel-migration.md
```diff
@@ -0,0 +1,39 @@
...
+
+1. synchronize execution on all cluster nodes to minimize downtime
+2. Uninstall weave, removing iptables rules and host files
+3. Edit kubeadm.conf on all primary nodes and run kubeadm init
```
emosbaugh, line 26: Rather than running kubeadm init would it be better to follow the steps in this guide? https://kubernetes.io/docs/tasks/administer-cluster/kubeadm/kubeadm-reconfigure/

docs/arch/adr-008-weave-to-flannel-migration.md
```diff
@@ -0,0 +1,39 @@
...
+4. Install flannel
+5. Stop kubelet, flush iptables, restart containerd and start kubelet
+6. wait for nodes to be ready, and then recreate cluster-networking pods (kube-system, then CSI, then everything else)
```
camilamacedo86, line 29: Hi @laverya,

It seems like you did a POC to came up with this steps. WDYT about add a new section here "technical details" with the steps/scripts and etc done to manually perform these steps?

docs/arch/adr-008-weave-to-flannel-migration.md
```diff
@@ -0,0 +1,39 @@
+# ADR 8: Weave to Flannel Migration
+
+## Context
+
+As proposed in ADR 6, the future CNI of kURL will be Flannel.
```
camilamacedo86, line 5: What about link this one.

docs/arch/adr-008-weave-to-flannel-migration.md
```diff
@@ -0,0 +1,39 @@
...
+However, almost all existing installations are running Weave.
+
+A migration path must be provided for those instances.
+
+There were two general proposals for a migration path.
+One was to use Multus-CNI to allow using both Flannel and Weave at the same time before stripping out Weave, allowing an upgrade while only taking down one node at once.
+The other was to run a migration that removed Weave from all instances simultaneously before installing Flannel.
+This would be far simpler, not require an additional IP range, and not require Multus to remain in the cluster going forwards.
+It would also require 'stopping the world' and cause downtime.
```
camilamacedo86, line 14: What about we have here:

## Proposal

You describe what do you think is the best option.

Then, we add another section

## Alternatives

Also, I think we could have on this one `Risks and Mitigations` where we describe the downtime and what are the mitigations ( script that does the work in parallel to reduce the time and communication )

docs/arch/adr-008-weave-to-flannel-migration.md
```diff
@@ -0,0 +1,39 @@
...
+5. Stop kubelet, flush iptables, restart containerd and start kubelet
+6. wait for nodes to be ready, and then recreate cluster-networking pods (kube-system, then CSI, then everything else)
+
+This script would be able to be called explicitly from `tasks.sh` or after opt-in from the user as part of the kURL upgrade process.
+
+## Status
+
+Proposed
+
+## Consequences
```
camilamacedo86, line 37: Also, wdyt about `Test Plan`?

How do we will test it out?

docs/arch/adr-008-weave-to-flannel-migration.md
```diff
@@ -0,0 +1,39 @@
...
+
+Users will be able to move from Weave to Flannel, but downtime will need to be scheduled to do so.
```
camilamacedo86, line 39: I think more one section here could be added , wdyt about `Drawbacks`? 

> The idea is to find the best form of an argument why this enhancement should not be implemented.
> 
> What trade-offs (technical/efficiency cost, user experience, flexibility, supportability, etc) must be made in order to implement this? What are the reasons we might not want to undertake this proposal, and how do we overcome them?
> 

> Does this proposal implement a behavior that's new/unique/novel? Is it poorly aligned with existing user expectations? Will it be a significant maintenance burden? Is it likely to be superceded by something else in the near future?

docs/arch/adr-008-weave-to-flannel-migration.md
```diff
@@ -0,0 +1,39 @@
...
+## Context
+
+As proposed in ADR 6, the future CNI of kURL will be Flannel.
+However, almost all existing installations are running Weave.
+
+A migration path must be provided for those instances.
+
+There were two general proposals for a migration path.
+One was to use Multus-CNI to allow using both Flannel and Weave at the same time before stripping out Weave, allowing an upgrade while only taking down one node at once.
+The other was to run a migration that removed Weave from all instances simultaneously before installing Flannel.
```
camilamacedo86, line 12: What are the pros/cons of each scenario?

docs/arch/adr-008-weave-to-flannel-migration.md
```diff
@@ -0,0 +1,39 @@
...
+
+Create a script to do the following:
+
+1. synchronize execution on all cluster nodes to minimize downtime
+2. Uninstall weave, removing iptables rules and host files
+3. Edit kubeadm.conf on all primary nodes and run kubeadm init
+4. Install flannel
+5. Stop kubelet, flush iptables, restart containerd and start kubelet
+6. wait for nodes to be ready, and then recreate cluster-networking pods (kube-system, then CSI, then everything else)
+
```
camilamacedo86, line 30: What we need to do to ensure that the migration was done successfully? 
Could we add some steps with basic checks?

docs/arch/adr-008-weave-to-flannel-migration.md
```diff
@@ -0,0 +1,39 @@
...
+## Solution
+
+Create a script to do the following:
+
+1. synchronize execution on all cluster nodes to minimize downtime
+2. Uninstall weave, removing iptables rules and host files
+3. Edit kubeadm.conf on all primary nodes and run kubeadm init
+4. Install flannel
+5. Stop kubelet, flush iptables, restart containerd and start kubelet
```
camilamacedo86, line 28: So, the migration path is from k8s 1.25,. Users in old versions will not be allowed to do it. Am I right?
In this case scenarios, would we suggest to them upgrade the cluster and then migrate? 

Maybe that is more one point to add.

docs/arch/adr-008-weave-to-flannel-migration.md
camilamacedo86, line 14: how long is the downtime? What is the size of the required window in evarage?

docs/arch/adr-008-weave-to-flannel-migration.md
laverya: Agreed, rollback seems plausible but it would likely take more engineering effort to setup (or manual intervention)

docs/arch/adr-008-weave-to-flannel-migration.md
laverya: Yeah, there's already code in weave to ensure that inter-pod networking works, should be able to leverage that

docs/arch/adr-008-weave-to-flannel-migration.md
laverya: I don't think there's a limitation on what k8s versions the migration can be done with?

Or at least I don't forsee any blockers for migrating on 1.23 (or earlier)

Just not from Docker, we'd have to migrate docker->containerd first (or handle docker too)

docs/arch/adr-008-weave-to-flannel-migration.md
laverya: In testing, <5m (when running commands manually and thus not perfectly synchronized)

docs/arch/adr-008-weave-to-flannel-migration.md
laverya: https://github.com/replicatedhq/kURL/blob/3876f5e7c24ab99d8680c45ec701a609e491f7c0/scripts/common/kubernetes.sh#L786

Incidentally I'm going to make a PR to make this have podantiaffinity, so that it'll run on different nodes when possible

docs/arch/adr-008-weave-to-flannel-migration.md
camilamacedo86: WDYT about we plan to just doc the steps for manual intervention/rollback if required as how debug it instead of try to implement it?

docs/arch/adr-008-weave-to-flannel-migration.md
```diff
@@ -0,0 +1,44 @@
+# ADR 8: Weave to Flannel Migration
+
```
camilamacedo86, line 2: Hi @laverya,

I think it is just missing add/doc the answer of the question and then it seems good enough to get merged.

docs/arch/adr-008-weave-to-flannel-migration.md
laverya: Partly because it depends on where in the process things go sideways
probably easy to write the doc anyways though

docs/arch/adr-008-weave-to-flannel-migration.md
laverya: What do you mean?

docs/arch/adr-008-weave-to-flannel-migration.md
```diff
@@ -37,3 +39,6 @@ Proposed
 ## Consequences
 
 Users will be able to move from Weave to Flannel, but downtime will need to be scheduled to do so.
+
+If a migration failed, manual remediation would be required.
```
emosbaugh, line 43: can you go into more detail what this means? i assume the cluster is down.

laverya: Yeah cluster would be very down
'why down' depends on where things broke, though

docs/arch/adr-008-weave-to-flannel-migration.md
```diff
@@ -0,0 +1,44 @@
...
+
+## Status
+
+Proposed
+
+## Consequences
+
+Users will be able to move from Weave to Flannel, but downtime will need to be scheduled to do so.
+
+If a migration failed, manual remediation would be required.
```
emosbaugh, line 43:
```suggestion
A failed migration may result in a cluster down scenario.
Manual remediation would be required.
```

emosbaugh (changes requested):

emosbaugh (approved):

PR #3743, opened 2022-11-16
chore: add a helper to test the project and make eaiser the process

camilamacedo86:

That makes faster/easier test things out since I can just update/comment/change the values and run all by once. Also, help new contributors first check if all is fine for them looking up for customizations and understand the details.
 

None
None

## Steps to reproduce
None

None

None


CONTRIBUTING.md
```diff
@@ -48,15 +51,12 @@ Testing can be accomplished on systems capable of hosting supported container ru
...
     ```bash
-    # Local workstation
-    make build/packages/kubernetes/1.19.3/ubuntu-18.04
-    make build/packages/kubernetes/1.19.3/images
-    make dist/containerd-1.6.8.tar.gz && tar xzvf dist/containerd-1.6.8.tar.gz
-    make dist/weave-2.8.1.tar.gz && tar xzvf dist/weave-2.8.1.tar.gz
-    make dist/openebs-3.3.0.tar.gz && tar xzvf dist/openebs-3.3.0.tar.gz
-    make dist/registry-2.8.1.tar.gz && tar xzvf dist/registry-2.8.1.tar.gz
+    # To build the sample under the [hack/testdata](./hack/testdata) directory targeting Ubuntu 22.04
+    make build/sample/ubuntu-22.04
```
laverya, line 59: I strongly disagree with changing this. The point of this set of commands is not to be something that the average developer runs. The point of this set is to expose, in clear form, exactly what commands need to be run for a build - and let people know how to change versions etc.

CONTRIBUTING.md
```diff
@@ -72,44 +72,12 @@ Testing can be accomplished on systems capable of hosting supported container ru
...
-      registry:
-        version: 2.8.1"
-    ```
-
-    After modifying the the Manifest, the `make watchrsync` command will automatically build the scripts and upload them to the remote server.
-    You must wait for the message `synced` to test out your changes on the server:
+   You must wait for the message `synced` to test out your changes on the server:
     
-    ![Screenshot 2022-11-06 at 20 06 35](https://user-images.githubusercontent.com/7708031/200198100-19219107-84dd-4631-a0e4-3200ad5feb99.png)
+   ![Screenshot 2022-11-06 at 20 06 35](https://user-images.githubusercontent.com/7708031/200198100-19219107-84dd-4631-a0e4-3200ad5feb99.png)
```
laverya, line 77: I would strongly prefer to keep the example manifest file here.

Is it too verbose?

CONTRIBUTING.md
camilamacedo86: We still able to change the versions and run all make bundles manually. It is an optional helper. 
Note that without it is required to:
- Copy and past from the latest.sh script the INSTALLER spec OR from a doc
- Run make per make which lead to does not know if you executed all that wish or not
- Looking every time what is that target to build the make command?  

Then note that with the helper you just need to update the sample and the makefile target to do the tests.
It makes the life easier (ihmo) for tests purposes.

CONTRIBUTING.md
camilamacedo86: That is not about to be verbose. The problem is not have the example here, the problem is need to make all copy and pastes, run make per make and etc. Keep a sample under a testdata make the life easier. (ihmo)

Regards the explanation it helps described the steps afterwords in more detail because usually first we only want to ensure that we are able to test things out and see the install running and a follow step we will be looking for to know how to test specific things/versions/specs.

CONTRIBUTING.md
laverya: If you want to have both the 'make all' and this, fine. But 'go edit this makefile' is a different level of discoverability than 'here's the command to build this addon version'.

camilamacedo86: > If you want to have both the 'make all the versions for one spec' command and this, great! 
Hun, are you suggesting a script that read the scripts/Manifest and create all commands? 
If so, I think that it is too much effort for a dev helper and I would still needing to copy and pasting the script/manifest spec content, ensure indentation and etc. 

> But 'go edit this makefile' is a different level of discoverability than 'here's the command to build this addon version'
The dev does not need to edit. However, it is much more sample just update the versions on samples and make to do another test then all steps. it is just a dev helper, if someone prefer can copy and paste the commands and run in the line as well. The idea here is just create a full example to make easier test out/understand how it works.

CONTRIBUTING.md
camilamacedo86: @laverya I add its purpose and etc in the target.
Also, see that the new target is under the Development section when we run make --help.

camilamacedo86:
It was not accepted so I am closing this one.

PR #3744, opened 2022-11-16
add ci test to not allow merge changes in the script/Manifest by mistake

camilamacedo86:

Does not allow we get merged prs that update the script/Manifest (not intentionally, just because it was used on the tests)
From now one the build test would fail with: (in this case)

Screenshot 2022-11-16 at 18 52 20

NONE

NONE

## Steps to reproduce
NONE

NONE

NONE


.gitignore
```diff
@@ -17,7 +17,7 @@ packages/*/*/assets
 packages/*/*/images
 packages/*/*/rhel-*
 packages/*/*/ubuntu-*
-scripts/Manifest
```
camilamacedo86, line 20: @laverya Thinking better that is not the best solution. Why?
If we need to change the layout we will need to push 2 prs ( one removing it and with the change then another returning back, if we forget the second one we might face the error again )

laverya: you can still commit things that are in gitignore, you just need to use the right commands

laverya: hmmm apparently gitignore does not work like I remembered it did

laverya: there's update-index --assume-unchanged but that's just for local

.github/workflows/build-test.yaml
```diff
@@ -21,6 +21,19 @@ jobs:
...
+    runs-on: ubuntu-20.04
+    steps:
+      - uses: actions/setup-go@v2
+        with:
+          go-version: 1.19
+        id: go
+      - uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
+      - run: make test-sanity
```
laverya, line 35: we should probably call this something more descriptive, like 'dont-change-manifest-file' or similar

camilamacedo86: I thought on this one be like a sanity which can do more than on test check
In this case it is calling go vet (I notice that we are not checking this), we could for example ensure that all header has a license and etc. However, if we try to commit a change on the scripts/Manifest we will get the error message:

```
$ make test-sanity           
...
### WARNING: You cannot merge changes on the script/Manifest! ###
make: *** [test-sanity] Error 1
```

camilamacedo86: But I will change that :-)

emosbaugh (changes requested):
can you document in contributing.md that if a change needs to be made to the manifest then it must be made in both files?

emosbaugh (approved):

PR #3749, opened 2022-11-17
Document object storage architecture and deployment and releases

emosbaugh:

Update the ARCHITECTURE.md document with added details on the object storage bucket directory structure and the automation that builds and deploys to the bucket.

### DIFF
--- a/ARCHITECTURE.md
+++ b/ARCHITECTURE.md
@@ -53,16 +53,20 @@ The browser UI for creating and viewing kURL installers.
 
 #### Typescript API
 
+https://github.com/replicatedhq/kURL-api/tree/main/src
+
 The API for creating kURL URLs and rendering installers.
 The API accepts a kURL Installer spec and returns a URL including a deterministic hash of the spec for installing a cluster based on that spec, for example https://k8s.kurl.sh/5e61e80.
 The API interacts stores these hashes in its relational database for retrieval.
-Additionally, the API is responsible for resolving the add-on version ("latest" and ".x") and rendering the spec when a script is requested, for example https://k8s.kurl.sh/5e61e80/install.sh or shorthand https://k8s.kurl.sh/5e61e80.
-The API reads the current kURL version from the object storage bucket (https://kurl-sh.s3.amazonaws.com/dist/VERSION) and uses that version to lookup add-on version information (e.g. https://kurl-sh.s3.amazonaws.com/dist/v2022.04.19-0/supported-versions-gen.json) for spec resolution.
+Additionally, the API is responsible for [resolving](https://github.com/replicatedhq/kURL-api/blob/f43d0fb1ed770b938dea97de64f39dacdd495c68/src/installers/index.ts#L955-L977) the add-on version ("latest" and ".x") and [rendering](https://github.com/replicatedhq/kURL-api/blob/f43d0fb1ed770b938dea97de64f39dacdd495c68/src/util/services/templates.ts) the spec when a script is requested, for example https://k8s.kurl.sh/5e61e80/install.sh or shorthand https://k8s.kurl.sh/5e61e80.
+The API [reads](https://github.com/replicatedhq/kURL-api/blob/f43d0fb1ed770b938dea97de64f39dacdd495c68/src/util/package/kurl-version.ts) the current kURL version from the object storage bucket (https://kurl-sh.s3.amazonaws.com/dist/VERSION) and uses that version to lookup add-on version information (e.g. https://kurl-sh.s3.amazonaws.com/dist/v2022.04.19-0/supported-versions-gen.json) for spec resolution.
 
 #### Go API Proxy
 
-This API proxies most requests directly to the Typescript API.
-Additionally, this API is responsible for assembling and streaming air-gap bundles back to the end-user.
+https://github.com/replicatedhq/kURL-api/tree/main/cmd/server
+
+This API [proxies](https://github.com/replicatedhq/kURL-api/blob/f43d0fb1ed770b938dea97de64f39dacdd495c68/cmd/server/main.go#L66-L69) most requests directly to the Typescript API.
+Additionally, this API is responsible for [assembling and streaming](https://github.com/replicatedhq/kURL-api/blob/f43d0fb1ed770b938dea97de64f39dacdd495c68/cmd/server/main.go#L212-L396) air-gap bundles back to the end-user.
 Air-gap packages are assembled from individual add-on package archives (e.g. https://kurl-sh.s3.amazonaws.com/dist/v2022.04.19-0/rook-1.0.4.tar.gz) and streamed as a single archive back to the end user when requested (e.g. https://k8s.kurl.sh/bundle/5e61e80.tar.gz).
 
 #### Relational Database
@@ -81,9 +85,9 @@ Stores built kURL assets including add-on package archives.
 
 | Object | Description |
 | ------ | ----------- |
-| s3://[bucket]/[(dist\|staging)]/VERSION                                    | The current kURL version, used by the API to determine what version subdirectory to serve packages from |
-| s3://[bucket]/[(dist\|staging)]/[kURL version]/supported-versions-gen.json | Add-on and version information for a given release of kURL, for resolving "latest" and "dot x" versions as well as validating installers |
-| s3://[bucket]/[(dist\|staging)]/[kURL version]/[entrypoint].tmpl           | Script templates rendered by the API, install.sh, join.sh, upgrade.sh and tasks.sh |
+| s3://[bucket]/[(dist\|staging)]/VERSION                                    | The current kURL version, used by the Typescript API to determine what version subdirectory to serve packages from |
+| s3://[bucket]/[(dist\|staging)]/[kURL version]/supported-versions-gen.json | Add-on and version information for a given release of kURL, used by the Typescript API for resolving "latest" and "dot x" versions as well as validating installers |
+| s3://[bucket]/[(dist\|staging)]/[kURL version]/[entrypoint].tmpl           | Script templates rendered by the Typescript API, install.sh, join.sh, upgrade.sh and tasks.sh |
 | s3://[bucket]/[(dist\|staging)]/[kURL version]/[addon-version].tar.gz      | Add-on package archives |
 
 ## kURL Installer
@@ -167,7 +171,7 @@ The kURL API merges the external add-on registry with its internal list of add-o
 
 Upon releasing kURL, scripts and add-on package archives are built and uploaded to to the kURL object storage bucket along with metadata including the Git sha from which they were generated.
 Once complete, the VERSION file is updated to point to the current version of kURL.
-The API makes use of this VERSION file to resolve the scripts and add-on packages.
+The Typescript API makes use of this VERSION file to resolve the scripts and add-on packages.
 
 ### Production Workflow
 
--- a/ARCHITECTURE.md
+++ b/ARCHITECTURE.md
@@ -56,7 +56,7 @@ The browser UI for creating and viewing kURL installers.
 The API for creating kURL URLs and rendering installers.
 The API accepts a kURL Installer spec and returns a URL including a deterministic hash of the spec for installing a cluster based on that spec, for example https://k8s.kurl.sh/5e61e80.
 The API interacts stores these hashes in its relational database for retrieval.
-Additionally, the API is responsible for resolving the add-on version ("latetst" and ".x") and rendering the spec when a script is requested, for example https://k8s.kurl.sh/5e61e80/install.sh or shorthand https://k8s.kurl.sh/5e61e80.
+Additionally, the API is responsible for resolving the add-on version ("latest" and ".x") and rendering the spec when a script is requested, for example https://k8s.kurl.sh/5e61e80/install.sh or shorthand https://k8s.kurl.sh/5e61e80.
 The API reads the current kURL version from the object storage bucket (https://kurl-sh.s3.amazonaws.com/dist/VERSION) and uses that version to lookup add-on version information (e.g. https://kurl-sh.s3.amazonaws.com/dist/v2022.04.19-0/supported-versions-gen.json) for spec resolution.
 
 #### Go API Proxy
@@ -171,7 +171,7 @@ The API makes use of this VERSION file to resolve the scripts and add-on package
 
 ### Production Workflow
 
-Production release are triggerd by running the command `make tag-and-release` or pushing a tag in the format "v*.*.*".
+Production release are triggered by running the command `make tag-and-release` or pushing a tag in the format "v*.*.*".
 Production releases are uploaded to the object storage bucket at prefix `dist`, for example https://kurl-sh.s3.amazonaws.com/dist/v2022.04.19-0/.
 Before building add-on packages, the workflow will first check if there were any changes made to the add-on since the previous production release based on metadata included with the add-on package.
 If no changes were made, the package will be copied from the previous production release to optimize for build times.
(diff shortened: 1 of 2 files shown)


ARCHITECTURE.md
```diff
@@ -51,34 +51,40 @@ https://github.com/replicatedhq/kurl-api
...
-#### Go API Proxy
+#### Typescript API
 
-This API proxies most requests directly to the Typescript API.
-Additionally, this API interacts with the object store and is responsible for assembling and streaming air-gap bundles back to the end-user.
+The API for creating kURL URLs and rendering installers.
+The API accepts a kURL Installer spec and returns a URL including a deterministic hash of the spec for installing a cluster based on that spec, for example https://k8s.kurl.sh/5e61e80.
+The API interacts stores these hashes in its relational database for retrieval.
+Additionally, the API is responsible for resolving the add-on version ("latest" and ".x") and rendering the spec when a script is requested, for example https://k8s.kurl.sh/5e61e80/install.sh or shorthand https://k8s.kurl.sh/5e61e80.
+The API reads the current kURL version from the object storage bucket (https://kurl-sh.s3.amazonaws.com/dist/VERSION) and uses that version to lookup add-on version information (e.g. https://kurl-sh.s3.amazonaws.com/dist/v2022.04.19-0/supported-versions-gen.json) for spec resolution.
```
camilamacedo86, line 61: Could we point put the repo + code ?
I think this one is https://github.com/replicatedhq/kURL-api/tree/main/src/controllers. Am I right?

ARCHITECTURE.md
```diff
@@ -51,34 +51,40 @@ https://github.com/replicatedhq/kurl-api
...
 
-#### Typescript API
+#### Go API Proxy
 
-The API for creating and consuming installers. Interacts with the relational database.
+This API proxies most requests directly to the Typescript API.
```
camilamacedo86, line 64: Where? Could we point out the code?

ARCHITECTURE.md
```diff
@@ -51,34 +51,40 @@ https://github.com/replicatedhq/kurl-api
...
+Air-gap packages are assembled from individual add-on package archives (e.g. https://kurl-sh.s3.amazonaws.com/dist/v2022.04.19-0/rook-1.0.4.tar.gz) and streamed as a single archive back to the end user when requested (e.g. https://k8s.kurl.sh/bundle/5e61e80.tar.gz).
 
 #### Relational Database
 
 Stores installer hashes.
 
 #### Object Store
 
-Stores built kURL assets including add-on bundles.
+Stores built kURL assets including add-on package archives.
```
camilamacedo86, line 74: It still not clear for me. Why? 
Ok. I build those to do the tests. However, it would be very nice to know how they are stored?
What are the formats? Could we add an example ?

ARCHITECTURE.md
```diff
@@ -51,34 +51,40 @@ https://github.com/replicatedhq/kurl-api
...
+![kURL.sh Architecture](https://user-images.githubusercontent.com/371319/166266866-267833a7-b21f-4665-b657-a55d6271b48b.png)
 
 ### Object Storage
 
 | Object | Description |
 | ------ | ----------- |
-| s3://[bucket]/[kURL version]/[entrypoint].tmpl           | Script templates rendered by the API. install.sh, join.sh, upgrade.sh and tasks.sh |
-| s3://[bucket]/[kURL version]/[addon-version].tar.gz      | Add-on bundles |
-| s3://[bucket]/[kURL version]/supported-versions-gen.json | Add-on and version information for a given release of kURL. For resolving "latest" and "dot x" versions as well as validating installers. |
+| s3://[bucket]/[(dist\|staging)]/VERSION                                    | The current kURL version, used by the API to determine what version subdirectory to serve packages from |
```
camilamacedo86, line 84: What API? We have more than one?

ARCHITECTURE.md
```diff
@@ -51,34 +51,40 @@ https://github.com/replicatedhq/kurl-api
...
+| s3://[bucket]/[(dist\|staging)]/[kURL version]/supported-versions-gen.json | Add-on and version information for a given release of kURL, for resolving "latest" and "dot x" versions as well as validating installers |
```
camilamacedo86, line 85: Who uses the supported-versions-gen.json? Where/how it is built? What is the project and code?

ARCHITECTURE.md
```diff
@@ -188,3 +194,34 @@ kURL automation, more specifically the `import-external-addons` GitHub action, p
 New versions are published to the external add-on registry and packages are copied from the source and stored in the kURL S3 bucket.
 
 The kURL API merges the external add-on registry with its internal list of add-on versions, making them available to the end-user.
+
+## Deployment and Releases
+
+Upon releasing kURL, scripts and add-on package archives are built and uploaded to to the kURL object storage bucket along with metadata including the Git sha from which they were generated.
+Once complete, the VERSION file is updated to point to the current version of kURL.
+The API makes use of this VERSION file to resolve the scripts and add-on packages.
```
camilamacedo86, line 202: Could we point out what API and the Version File?

ARCHITECTURE.md
```diff
@@ -188,3 +194,34 @@ kURL automation, more specifically the `import-external-addons` GitHub action, p
...
+
+Production release are triggered by running the command `make tag-and-release` or pushing a tag in the format "v*.*.*".
+Production releases are uploaded to the object storage bucket at prefix `dist`, for example https://kurl-sh.s3.amazonaws.com/dist/v2022.04.19-0/.
+Before building add-on packages, the workflow will first check if there were any changes made to the add-on since the previous production release based on metadata included with the add-on package.
+If no changes were made, the package will be copied from the previous production release to optimize for build times.
+Next, the workflow will check if there were any changes made since the previous staging release and copy from staging if no changes were made.
+The is to account for a scenario where a commit were being tagged other than what is at the HEAD of main.
+If changes were made to both production and staging, the package will be built from source and uploaded.
+Finally, VERSION file is updated at https://kurl-sh.s3.amazonaws.com/dist/VERSION to point to the new production version.
+Historical production releases are never removed from the object storage bucket.
```
camilamacedo86, line 214: So, we have like 4GB for each release and we are keeping them for each version releases. Is that?

camilamacedo86: Have we not specific jobs that are trigged in this scenario?
What tests are executed in this scenario that are not executed when we push a PR?

ARCHITECTURE.md
```diff
@@ -188,3 +194,34 @@ kURL automation, more specifically the `import-external-addons` GitHub action, p
...
+Staging releases are triggered on merge to main.
+Staging release versions use the most current release version tag and append prerelease `-dirty`, for example v2022.04.19-0-dirty.
+Due to this versioning scheme, staging releases will overwrite the previous staging release if no production release occurred prior.
+This is intentional to optimize for storage costs.
+Staging releases are uploaded to the object storage bucket at prefix `staging`, for example https://kurl-sh.s3.amazonaws.com/staging/v2022.04.19-0-dirty/.
+Before building add-on packages, the workflow will first check if there were any changes made to the add-on since the previous staging release based on metadata included with the add-on package.
+If no changes were made, the package will be copied from the previous staging release to optimize for build times.
+If changes were made, the package will be built from source and uploaded.
+Finally, VERSION file is updated at https://kurl-sh.s3.amazonaws.com/staging/VERSION to point to the new staging version.
+Historical staging releases are never removed from the object storage bucket.
```
camilamacedo86, line 227: Have we a reason to keep all artefacts for the staging ones? Have we any scenario where would be required to use the old ones? If not, could we not remove to reduce the cost to keep all?

ARCHITECTURE.md
```diff
@@ -51,34 +51,40 @@ https://github.com/replicatedhq/kurl-api
...
+#### Go API Proxy
 
-The API for creating and consuming installers. Interacts with the relational database.
+This API proxies most requests directly to the Typescript API.
+Additionally, this API is responsible for assembling and streaming air-gap bundles back to the end-user.
+Air-gap packages are assembled from individual add-on package archives (e.g. https://kurl-sh.s3.amazonaws.com/dist/v2022.04.19-0/rook-1.0.4.tar.gz) and streamed as a single archive back to the end user when requested (e.g. https://k8s.kurl.sh/bundle/5e61e80.tar.gz).
 
 #### Relational Database
 
 Stores installer hashes.
```
camilamacedo86, line 70: What about add here:

How the hashes are generated? What is the source?
Where we can find the schemas? What is the motivation for ? 
How uses this database? vendor, kurl-API ? What API inside of the kurl-API needs to sql the database?
What are the projects that uses these database? What is stored here? Have this database relation with other projects that are not kURL? 

Something like:

> Stores installer hashes built from .... (code/repo) .... . The database schemas are ... . The installer scripts are stored because .... . 

On top of that, I did some questions about that and @laverya shared that the schemas are into : https://github.com/replicatedhq/vandoor/tree/main/migrations/kustomize/schema

Then, in this case I understand that we would have : https://github.com/replicatedhq/vandoor/blob/main/migrations/kustomize/schema/kurl-installer.yaml
But that seems to create the relation that 1 team have 1 installer schema ID. But that still not clear for me.

Should not each release be associated to an installer release?
Or have we not this association?
What happens if I want to upgrade and I have an replicated/kots app/release installed with an old kURL script?
Should not an app/vendor release be associated to script and not k8s supported versions?

ARCHITECTURE.md
```diff
@@ -51,34 +51,40 @@ https://github.com/replicatedhq/kurl-api
...
+Additionally, the API is responsible for resolving the add-on version ("latest" and ".x") and rendering the spec when a script is requested, for example https://k8s.kurl.sh/5e61e80/install.sh or shorthand https://k8s.kurl.sh/5e61e80.
+The API reads the current kURL version from the object storage bucket (https://kurl-sh.s3.amazonaws.com/dist/VERSION) and uses that version to lookup add-on version information (e.g. https://kurl-sh.s3.amazonaws.com/dist/v2022.04.19-0/supported-versions-gen.json) for spec resolution.
 
-#### Typescript API
+#### Go API Proxy
 
-The API for creating and consuming installers. Interacts with the relational database.
+This API proxies most requests directly to the Typescript API.
+Additionally, this API is responsible for assembling and streaming air-gap bundles back to the end-user.
+Air-gap packages are assembled from individual add-on package archives (e.g. https://kurl-sh.s3.amazonaws.com/dist/v2022.04.19-0/rook-1.0.4.tar.gz) and streamed as a single archive back to the end user when requested (e.g. https://k8s.kurl.sh/bundle/5e61e80.tar.gz).
```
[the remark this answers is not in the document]
laverya: this is also the file that handles proxying requests to typescript

ARCHITECTURE.md
laverya: vandoor is not a public repo, so we can't really link to it here

ARCHITECTURE.md
laverya: Yes, people intentionally run previous kURL versions - it's a way for vendors to do their own QA for a release before inflicting it upon their users

ARCHITECTURE.md
camilamacedo86: That is fine. But could we clarify how the data is stored/when and etc? Also, I added the questions here to try take advantage and see if I can learn how the projects are connected.

ARCHITECTURE.md
camilamacedo86: Answered in the call

ARCHITECTURE.md
camilamacedo86: mainly answered .. 
I have just one point here that I will ask offiline

ARCHITECTURE.md
camilamacedo86: Mainly answered as well closing this one.

ARCHITECTURE.md
emosbaugh: No. I propose we change this to delete staging releases after some measure

ARCHITECTURE.md
emosbaugh: https://github.com/replicatedhq/kURL/pull/3749/files#diff-8f6366fd8e7a5fa30f7f05879999195b7cf5fa1e3d157822eb37f0e91d226cfdR60

emosbaugh:
@camilamacedo86 i added more links. can you re-review? thanks

ARCHITECTURE.md
```diff
@@ -51,34 +51,44 @@ https://github.com/replicatedhq/kurl-api
...
+#### Typescript API
 
-This API proxies most requests directly to the Typescript API.
-Additionally, this API interacts with the object store and is responsible for assembling and streaming air-gap bundles back to the end-user.
+https://github.com/replicatedhq/kURL-api/tree/main/src
 
-#### Typescript API
+The API for creating kURL URLs and rendering installers.
+The API accepts a kURL Installer spec and returns a URL including a deterministic hash of the spec for installing a cluster based on that spec, for example https://k8s.kurl.sh/5e61e80.
+The API interacts stores these hashes in its relational database for retrieval.
```
camilamacedo86, line 60: That still not clear for me. We have all possible combinations for each release stored. Am I right?
Then, how/where the SHA is generated? is possible to link the code / place?
Also, what is the schema of this table? 
Is this scripts also stored in the same database that we have the replicated users, portal data or have we an database only for kURL? I  understand that have all together. 

But I think we need to have a place that describe how it is working as pointing out. If we be unable to put here then, wdyt about we add in the wiki? Then we link the wiki page here ( valid only for replicated team members something like that)

ARCHITECTURE.md
```diff
@@ -51,34 +51,44 @@ https://github.com/replicatedhq/kurl-api
...
+
+https://github.com/replicatedhq/kURL-api/tree/main/cmd/server
+
+This API [proxies](https://github.com/replicatedhq/kURL-api/blob/f43d0fb1ed770b938dea97de64f39dacdd495c68/cmd/server/main.go#L66-L69) most requests directly to the Typescript API.
+Additionally, this API is responsible for [assembling and streaming](https://github.com/replicatedhq/kURL-api/blob/f43d0fb1ed770b938dea97de64f39dacdd495c68/cmd/server/main.go#L212-L396) air-gap bundles back to the end-user.
+Air-gap packages are assembled from individual add-on package archives (e.g. https://kurl-sh.s3.amazonaws.com/dist/v2022.04.19-0/rook-1.0.4.tar.gz) and streamed as a single archive back to the end user when requested (e.g. https://k8s.kurl.sh/bundle/5e61e80.tar.gz).
 
 #### Relational Database
 
 Stores installer hashes.
```
camilamacedo86, line 74: How we build the hashes ? Could we describe it as well?
Is hashes made from what?

ARCHITECTURE.md
camilamacedo86: Understand what we store, when and how as where it is used help us to understand how the project is working, the requirements, how the projects are connected and etc. I still a little strangle to get this part.

ARCHITECTURE.md
```diff
@@ -188,3 +198,34 @@ kURL automation, more specifically the `import-external-addons` GitHub action, p
 New versions are published to the external add-on registry and packages are copied from the source and stored in the kURL S3 bucket.
 
 The kURL API merges the external add-on registry with its internal list of add-on versions, making them available to the end-user.
+
+## Deployment and Releases
+
+Upon releasing kURL, scripts and add-on package archives are built and uploaded to to the kURL object storage bucket along with metadata including the Git sha from which they were generated.
+Once complete, the VERSION file is updated to point to the current version of kURL.
```
camilamacedo86, line 205: To understand how it works would be great know what is the format of this VERSION file?
What has in this Version file? Is possible to provide an example?

emosbaugh: I've linked to the file itself

ARCHITECTURE.md
emosbaugh: I've tried to address all these questions in the commit https://github.com/replicatedhq/kURL/pull/3749/commits/3548f35b19cf194a5415a232e21b5b26b8a9c5f7

camilamacedo86 (approved):
it is a quit improvement. 🥇 
I am approving this one.

See that I raised some questions and etc. 
But we can always address them in a follow up as well :-)

camilamacedo86 (approved):

PR #3751, opened 2022-11-18
Use OpenEBS app version as add-on version

emosbaugh:

We are using the openebs chart version as the add-on version. This does not track the actual version from the project. Instead use the app version for the add-on version.


camilamacedo86 (dismissed):

rrpolanco (approved):
lgtm

PR #3415, opened 2022-09-03
Automated OpenEBS version update 3.3.1

replicated-ci-kurl:
Automated changes by the [cron-openebs-update](https://github.com/replicatedhq/kURL/blob/main/.github/workflows/update-openebs.yaml) GitHub action

```release-note
Adds [OpenEBS add-on](https://kurl.sh/docs/add-ons/openebs) version 3.3.1.
```

PR #3752, opened 2022-11-18
Enable OpenEBS add-on auto-merge

emosbaugh:

Enables auto-merge for the OpenEBS add-on

### DIFF
--- a/.github/workflows/update-openebs.yaml
+++ b/.github/workflows/update-openebs.yaml
@@ -80,6 +80,7 @@ jobs:
           openebs
           type::chore
           kurl::type::feature
+          auto-merge
         draft: false
         base: "main"
         body: |


camilamacedo86 (approved):

.github/workflows/update-openebs.yaml
```diff
@@ -80,6 +80,7 @@ jobs:
           openebs
           type::chore
           kurl::type::feature
+          auto-merge
```
camilamacedo86, line 83: Are we sure that all new versions will work fine?

camilamacedo86 (dismissed):
.

camilamacedo86 (changes requested):
Could we discuss this one first? Note that: 
- we have a test that has been failing: https://github.com/replicatedhq/kURL/pull/3415
- we cannot ensure that new versions will work well and has no breaking changes
- I could not find any other case scenario that we automatically get new add-ons version merged.
- if the new version pass on the tests then, we can manually approve and let it be merged, So, it does not seems a huge effort. Why would we want to configure for this PRs get merged automatically? 

In this way, I do not think that we should move within.

camilamacedo86 (approved):
My concern was answered and addressed by @rrpolanco explanation:

> We have a github action that runs as part of the PR when our cronjob detects a new app version update. When auto-merge is enabled, there's a validation check that runs which determines if testgrid passed or failed. If testgrid did not pass, the auto-merge will not happen.

/approved

PR #3760, opened 2022-11-21
Add Rook version 1.10.6

emosbaugh:

Updates rook to the latest 1.10 release

Fixes NONE

```release-note
Adds [Rook add-on](https://kurl.sh/docs/add-ons/rook) version 1.10.6.
```

NONE


camilamacedo86 (approved):

addons/rook/template/generate.sh
```diff
@@ -55,16 +55,6 @@ function generate() {
...
-    # Upgrading passed v16.2.6 based on this note in the Rook docs
-    # https://www.rook.io/docs/rook/v1.6/ceph-upgrade.html#ceph-version-upgrades
-    # WARNING: There is a notice from Ceph for users upgrading to Ceph Pacific v16.2.6 or lower from an earlier major version of Ceph. If you are upgrading to Ceph Pacific (v16), please upgrade to v16.2.7 or higher if possible.
-    if echo "$ceph_image" | grep -q ":v16\.2" ; then
-        local ceph_version_patch="$(echo "$ceph_image" | grep -o "v[0-9]*\.[0-9]*\.[0-9]*" | sed "s/v[0-9]*\.[0-9]*\.\([0-9]*\).*/\1/")"
-        if [ "$ceph_version_patch" -lt "7" ]; then
-            ceph_image="$(echo "$ceph_image" | sed "s/v[0-9]*\.[0-9]*\.[0-9]*/v16.2.7/")"
-        fi
-    fi
-
```
camilamacedo86, line 67: @emosbaugh should we really remove this one?

emosbaugh: this script is to generate new versions. the latest version of rook uses ceph quincy 17. so i dont think we will ever ship ceph pacific 16 again.

addons/rook/template/base/install.sh
```diff
@@ -160,12 +160,6 @@ function rook_operator_deploy() {
     # upgrade first before applying auth_allow_insecure_global_id_reclaim policy
     rook_maybe_auth_allow_insecure_global_id_reclaim
 
-    # disable bluefs_buffered_io for rook 1.8.x and 1.9.x
-    # See:
-    #   - https://github.com/rook/rook/issues/10160#issuecomment-1168303067
-    #   - https://tracker.ceph.com/issues/54019
-    rook_maybe_bluefs_buffered_io
```
camilamacedo86, line 167: What happens if we have installed an old version and we upgrade to this pinned ones with the latest kURL release? Also, from the oldest to the latest should we need to still disabling it?

addons/rook/template/base/install.sh
```diff
@@ -520,16 +514,6 @@ function rook_should_fail_install() {
...
-function rook_maybe_bluefs_buffered_io() {
-    local dst="${DIR}/kustomize/rook/operator"
-
-    semverParse "$ROOK_VERSION"
-    local rook_major_minor_version="${major}.${minor}"
-    if [ "$rook_major_minor_version" = "1.8" ] || [ "$rook_major_minor_version" = "1.9" ]; then
-        sed -i "/\[global\].*/a\    bluefs_buffered_io = false" "$dst/configmap-rook-config-override.yaml"
-    fi
-}
-
```
rrpolanco, line 532: Why are we removing this? Wouldn't we still want this if upgrading _to_ Rook 1.8 or 1.9 from 1.4?

camilamacedo86: Same here: https://github.com/replicatedhq/kURL/pull/3760/files#r1036420648 :-P

addons/rook/template/base/install.sh
emosbaugh: this is in the template used to generate new versions. it is [still](https://github.com/replicatedhq/kURL/blob/6737ccb509692ba58d74fe1dc45482aee579242c/addons/rook/1.9.12/install.sh#L172) in the rook add-on versions affected by the issue. if a user were to upgrade to a later version they would no longer be on a version of ceph affected by the issue (17 quincy).

addons/rook/template/base/install.sh
emosbaugh: this is in the template used to generate new versions. it is [still](https://github.com/replicatedhq/kURL/blob/6737ccb509692ba58d74fe1dc45482aee579242c/addons/rook/1.9.12/install.sh#L172) in the rook add-on versions affected by the issue. if a user were to upgrade to a later version they would no longer be on a version of ceph affected by the issue (17 quincy).

rrpolanco: ah, that's right! The affected 1.8 and 1.9 versions will still have this fix but not this version and later. Got it.

PR #3763, opened 2022-11-22
Merged Patch weave images for vulnerabilities 

camilamacedo86:

Address CVEs

## Steps to reproduce
Scan the images: https://github.com/replicatedhq/kURL/actions/runs/3521359488/jobs/5903133316

```release-note
Adds [Weave add-on](https://kurl.sh/docs/add-ons/weave) versions 2.8.1-20221122 to address the critical CVEs
 (CVE-2022-42915, CVE-2022-42915) and the high CVEs (CVE-2022-42916, CVE-2022-42916)
```

### DIFF
--- a/web/src/installers/versions.js
+++ b/web/src/installers/versions.js
@@ -97,7 +97,6 @@ module.exports.InstallerVersions = {
   ],
   weave: [
     // cron-weave-update-265
-    "2.8.1-20221122",
     "2.6.5-20221025",
     "2.6.5-20221006",
     "2.6.5-20220825",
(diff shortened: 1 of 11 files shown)


web/src/installers/versions.js
```diff
@@ -97,6 +97,7 @@ module.exports.InstallerVersions = {
   ],
   weave: [
     // cron-weave-update-265
+    "2.8.1-20221122",
```
laverya, line 100: This looks wrong

camilamacedo86: Why? That was added by the script automatically ./generate.sh
It follows the same approach of : https://github.com/replicatedhq/kURL/pull/3628/files

laverya: Because it's 281 in the 265 section, and it's in here twice (shall on line 110)

laverya (approved):

PR #2998, opened 2022-06-08
use runc 1.1.2 with containerd

divolgin:
#### Which issue(s) this PR fixes:

Fixes #

#### Does this PR introduce a user-facing change?

```release-note

```

PR #3766, opened 2022-11-22
fix: ensure that runc version added is the same used/supported by

camilamacedo86:

To ensure that runc version added is the same used/supported so that we might fix the currently issues reported within.

Fixes # https://github.com/replicatedhq/kURL/pull/2998

The documentation and the reason for the specific versions pinned are described via comments.

```
fix: ensure that containerd addons versions will be shipped with the respective supported runc version. Containerd addon versions >= 1.6.4 will be built with runc version `v1.1.4` instead of `v1.0.0-rc95`
```

### DIFF
--- a/addons/containerd/template/script.sh
+++ b/addons/containerd/template/script.sh
@@ -24,8 +24,28 @@ function init_manifest_file() {
 
     cat <<EOT >> $file
 yum libzstd
-asset runc https://github.com/opencontainers/runc/releases/download/v1.0.0-rc95/runc.amd64
 EOT
+    # Note that containerd requires runc and each release officially uses one specific version in their
+    # tests. Therefore, that is the version of runc which is supported and should be used by each
+    # respective containerd release. More info: https://github.com/containerd/containerd/blob/main/docs/RUNC.md
+    if echo "$version" | grep -qF "1.2."; then
+        # See the runc version used to test the releases 1.2:https://github.com/containerd/containerd/blob/release/1.2/vendor.conf#L23
+        echo "asset runc https://github.com/opencontainers/runc/releases/download/v1.0.0-rc10/runc.amd64" >> $file
+    elif echo "$version" | grep -qF "1.3."; then
+        # See the runc version used to test the releases 1.3:https://github.com/containerd/containerd/blob/release/1.3/vendor.conf#L33
+        echo "asset runc https://github.com/opencontainers/runc/releases/download/v1.0.0-rc10/runc.amd64" >> $file
+    elif echo "$version" | grep -qF "1.4."; then
+        # See the runc version used to test the releases 1.4:https://github.com/containerd/containerd/blob/release/1.4/script/setup/runc-version
+        echo "asset runc https://github.com/opencontainers/runc/releases/download/v1.0.3/runc.amd64" >> $file
+    else
+        # detect runc version from upstream project used to test the containerd release
+        local runc_version="$(curl -sSL https://raw.githubusercontent.com/containerd/containerd/release/"$(echo "$version" | grep -Eo '[[:digit:]]\.[[:digit:]]+')"/script/setup/runc-version)"
+        if [ -z "$runc_version" ]; then
+            echo "Failed to detect runc version"
+            exit 1
+        fi
+        echo "asset runc https://github.com/opencontainers/runc/releases/download/$runc_version/runc.amd64" >> $file
+    fi
 }
 
 function add_supported_os_to_manifest_file() {
--- a/addons/containerd/1.5.10/Manifest
+++ b/addons/containerd/1.5.10/Manifest
@@ -1,5 +1,5 @@
 yum libzstd
-asset runc https://github.com/opencontainers/runc/releases/download/v1.0.0-rc95/runc.amd64
+asset runc https://github.com/opencontainers/runc/releases/download/v1.1.4/runc.amd64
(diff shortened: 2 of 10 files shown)


addons/containerd/template/script.sh
```diff
@@ -24,8 +24,23 @@ function init_manifest_file() {
 
     cat <<EOT >> $file
 yum libzstd
-asset runc https://github.com/opencontainers/runc/releases/download/v1.0.0-rc95/runc.amd64
 EOT
+    # Note that containerd requires runc and each release officially uses one specific version in their
+    # tests. Therefore, that is the version of runc which is supported and should be used by each
+    # respective containerd release. More info: https://github.com/containerd/containerd/blob/main/docs/RUNC.md
+    if echo "$version" | grep -qE "1\.2\."; then
```
emosbaugh, line 31:
```suggestion
    if echo "$version" | grep -qF "1.2."; then
```

addons/containerd/template/script.sh
```diff
@@ -24,8 +24,23 @@ function init_manifest_file() {
...
+        # See the runc version used to test the releases 1.2:https://github.com/containerd/containerd/blob/release/1.2/vendor.conf#L23
+        echo "asset runc https://github.com/opencontainers/runc/releases/download/v1.0.0-rc10/runc.amd64" >> $file
+    elif echo "$version" | grep -qE "1\.3\."; then
```
emosbaugh, line 34:
```suggestion
    if echo "$version" | grep -qF "1.3."; then
```

addons/containerd/template/script.sh
```diff
@@ -24,8 +24,23 @@ function init_manifest_file() {
...
+        # See the runc version used to test the releases 1.3:https://github.com/containerd/containerd/blob/release/1.3/vendor.conf#L33
+        echo "asset runc https://github.com/opencontainers/runc/releases/download/v1.0.0-rc10/runc.amd64" >> $file
+    elif echo "$version" | grep -qE "1\.4\."; then
```
emosbaugh, line 37:
```suggestion
    if echo "$version" | grep -qF "1.4."; then
```

emosbaugh (changes requested):

addons/containerd/template/script.sh
```diff
@@ -24,8 +24,23 @@ function init_manifest_file() {
...
+        # See the runc version used to test the releases 1.4:https://github.com/containerd/containerd/blob/release/1.4/script/setup/runc-version
+        echo "asset runc https://github.com/opencontainers/runc/releases/download/v1.0.3/runc.amd64" >> $file
+    else
+        # See the runc version used to test the releases 1.5:https://github.com/containerd/containerd/blob/release/1.5/script/setup/runc-version
+        echo "asset runc https://github.com/opencontainers/runc/releases/download/v1.1.4/runc.amd64" >> $file
+    fi
```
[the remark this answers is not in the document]
camilamacedo86: Good thought :-)

addons/containerd/template/script.sh
```diff
@@ -24,8 +24,24 @@ function init_manifest_file() {
...
+        echo "asset runc https://github.com/opencontainers/runc/releases/download/v1.0.0-rc10/runc.amd64" >> $file
+    elif echo "$version" | grep -qF "1.3."; then
+        # See the runc version used to test the releases 1.3:https://github.com/containerd/containerd/blob/release/1.3/vendor.conf#L33
+        echo "asset runc https://github.com/opencontainers/runc/releases/download/v1.0.0-rc10/runc.amd64" >> $file
+    elif echo "$version" | grep -qF "1.4."; then
+        # See the runc version used to test the releases 1.4:https://github.com/containerd/containerd/blob/release/1.4/script/setup/runc-version
+        echo "asset runc https://github.com/opencontainers/runc/releases/download/v1.0.3/runc.amd64" >> $file
+    else
+        # detect runc version from upstream project used to test the containerd release
+        local runc_version="$(curl -sSL https://raw.githubusercontent.com/containerd/containerd/release/"$(echo "$version" | grep -Eo '[[:digit:]]\.[[:digit:]]+')"/script/setup/runc-version)"
```
emosbaugh, line 42: can we bail here if we failed to get the version?

```suggestion
        local runc_version="$(curl -sSL https://raw.githubusercontent.com/containerd/containerd/release/"$(echo "$version" | grep -Eo '[[:digit:]]\.[[:digit:]]+')"/script/setup/runc-version)"
        if [ -z "$runc_version" ]; then
            echo "Failed to detect runc version"
            exit 1
        fi
```

emosbaugh (changes requested):

addons/containerd/template/script.sh
camilamacedo86: Good point !

emosbaugh (approved):

camilamacedo86:
Analyse of errors faced
----
**[test-addons (containerd, 1.6.4, pr-3766-772f45f) / test-addon](https://github.com/replicatedhq/kURL/actions/runs/3565416119/jobs/5990552742#logs)**

In this case the failure is correct because the version is incompatible.
Internal task open for we check if we can improve the tests [62721]

emosbaugh:
It looks like most errors are failures to downgrade or failures when checking the version changed on an upgrade although its upgrading to the same version.

The airgap unpacking errors I believe are due to a deployment resetting the connection but I am not sure.

emosbaugh (approved):

PR #3767, opened 2022-11-22
Run Node Disk and Volume Validation Checks when Upgrading from Rook to OpenEBS

rrpolanco:

Apply node disk validation introduced in https://github.com/replicatedhq/kURL/pull/3674 and volume access modes validation introduced in https://github.com/replicatedhq/pvmigrate/pull/139 when upgrading from Rook to OpenEBS.

Fixes #
Internal tracker ref: [sc-61144]

Unfortunately, the volume access modes validation cannot run at pre-init time because it requires the `openebs` storage class to be installed.

```release-note
Disk and Volume validation checks will be run prior to migrating from Rook to OpenEBS. A failed validation check will abort the upgrade.
```

### DIFF
--- a/addons/openebs/3.3.0/install.sh
+++ b/addons/openebs/3.3.0/install.sh
@@ -60,6 +60,8 @@ function openebs_maybe_migrate_from_rook() {
 function openebs_maybe_rook_migration_checks() {
 
     # get the list of StorageClasses that use rook-ceph
+    local rook_scs
+    local rook_default_sc
     rook_scs=$(kubectl get storageclass | grep rook | grep -v '(default)' | awk '{ print $1}') # any non-default rook StorageClasses
     rook_default_sc=$(kubectl get storageclass | grep rook | grep '(default)' | awk '{ print $1}') # any default rook StorageClasses
 
@@ -74,6 +76,9 @@ function openebs_maybe_rook_migration_checks() {
     do
         # run validation checks for non default Rook storage classes
         rook_scs_pvmigrate_dryrun_output=$($BIN_PVMIGRATE --source-sc "$rook_sc" --dest-sc "$OPENEBS_LOCALPV_STORAGE_CLASS" --rsync-image "$KURL_UTIL_IMAGE" --preflight-validation-only || true)
+        if [ $? -ne 0 ]; then
+            break
+        fi
     done
 
     if [ -n "$rook_default_sc" ] ; then
--- a/addons/openebs/3.3.0/install.sh
+++ b/addons/openebs/3.3.0/install.sh
@@ -15,7 +15,7 @@ function openebs_pre_init() {
     export PREVIOUS_OPENEBS_VERSION="$(openebs_get_running_version)"
 
     openebs_bail_unsupported_upgrade
-    prompt_migrate_from_rook
+    openebs_prompt_migrate_from_rook
 }
 
 function openebs() {
@@ -49,58 +49,47 @@ function openebs_maybe_migrate_from_rook() {
     if [ -z "$ROOK_VERSION" ]; then
         if kubectl get ns | grep -q rook-ceph; then
             # show validation errors from pvmigrate
-            # if there are errors, maybe_prompt_migrate_from_rook_dryrun_errors() will bail
-            maybe_prompt_migrate_from_rook_dryrun_errors
-            rook_ceph_to_sc_migration "$OPENEBS_LOCALPV_STORAGE_CLASS"
+            # if there are errors, maybe_rook_to_openebs_migration_checks() will bail
+            maybe_rook_to_openebs_migration_checks
+            rook_ceph_to_sc_migration "$OPENEBS_LOCALPV_STORAGE_CLASS" "1"
             DID_MIGRATE_ROOK_PVCS=1 # used to automatically delete rook-ceph if object store data was also migrated
         fi
     fi
 }
 
-function maybe_prompt_migrate_from_rook_dryrun_errors() {
-    echo "Running Rook to OpenEBS Migration Checks ..."
+function maybe_rook_to_openebs_migration_checks() {
 
     # get the list of StorageClasses that use rook-ceph
     rook_scs=$(kubectl get storageclass | grep rook | grep -v '(default)' | awk '{ print $1}') # any non-default rook StorageClasses
     rook_default_sc=$(kubectl get storageclass | grep rook | grep '(default)' | awk '{ print $1}') # any default rook StorageClasses
 
+    # Ensure openebs-localpv-provisioner deployment is ready
+    printf "awaiting openebs-localpv-provisioner deployment\n"
+    spinner_until 120 deployment_fully_updated openebs openebs-localpv-provisioner
+
+    echo "Running Rook to OpenEBS Migration Checks ..."
     local rook_scs_pvmigrate_dryrun_output
     local rook_default_sc_pvmigrate_dryrun_output
     for rook_sc in $rook_scs
     do
         # run validation checks for non default Rook storage classes
-        rook_scs_pvmigrate_dryrun_output=$($BIN_PVMIGRATE --source-sc "$rook_sc" --dest-sc "$OPENEBS_LOCALPV_STORAGE_CLASS" --dry-run 2>/dev/null)
+        rook_scs_pvmigrate_dryrun_output=$($BIN_PVMIGRATE --source-sc "$rook_sc" --dest-sc "$OPENEBS_LOCALPV_STORAGE_CLASS" --rsync-image "$KURL_UTIL_IMAGE" --preflight-validation-only 2>/dev/null || true)
     done
 
     if [ -n "$rook_default_sc" ] ; then
         # run validation checks for Rook default storage class
-        rook_default_sc_pvmigrate_dryrun_output=$($BIN_PVMIGRATE --source-sc "$rook_default_sc" --dest-sc "$OPENEBS_LOCALPV_STORAGE_CLASS" --dry-run 2>/dev/null)
+        rook_default_sc_pvmigrate_dryrun_output=$($BIN_PVMIGRATE --source-sc "$rook_default_sc" --dest-sc "$OPENEBS_LOCALPV_STORAGE_CLASS" --rsync-image "$KURL_UTIL_IMAGE" --preflight-validation-only 2>/dev/null || true)
     fi
 
-    if { [ -n "$rook_scs_pvmigrate_dryrun_output" ] && [[ "$rook_scs_pvmigrate_dryrun_output" != *"ProvisioningFailed"* ]]; } || { [ -n "$rook_default_sc_pvmigrate_dryrun_output" ] && [[ "$rook_default_sc_pvmigrate_dryrun_output" != *"ProvisioningFailed"* ]]; }; then
-        printf "${YELLOW}"
-        printf "\n"
-        printf "$rook_scs_pvmigrate_dryrun_output"
-        printf "\n"
-        printf "$rook_default_sc_pvmigrate_dryrun_output"
-        printf "\n"
-        printf "${NC}"
-        printf "\n"
-        return
-    else
-        printf "${RED}"
-        printf "\n"
-        printf "$rook_scs_pvmigrate_dryrun_output"
-        printf "\n"
-        printf "$rook_default_sc_pvmigrate_dryrun_output"
-        printf "\n"
-        printf "${NC}"
-        printf "\n"
-    fi
+    if [ -n "$rook_scs_pvmigrate_dryrun_output" ] || [ -n "$rook_default_sc_pvmigrate_dryrun_output" ] ; then
+        logFail "$rook_scs_pvmigrate_dryrun_output"
+        logFail "$rook_default_sc_pvmigrate_dryrun_output"
+        
         bail "Cannot upgrade from Rook to OpenEBS due to previous errors."
+    fi
 }
 
-function prompt_migrate_from_rook() {
+function openebs_prompt_migrate_from_rook() {
     local ceph_disk_usage_total
     local rook_ceph_exec_deploy=rook-ceph-operator
 
--- a/addons/openebs/3.3.0/install.sh
+++ b/addons/openebs/3.3.0/install.sh
@@ -75,8 +75,7 @@ function openebs_maybe_rook_migration_checks() {
     for rook_sc in $rook_scs
     do
         # run validation checks for non default Rook storage classes
-        rook_scs_pvmigrate_dryrun_output=$($BIN_PVMIGRATE --source-sc "$rook_sc" --dest-sc "$OPENEBS_LOCALPV_STORAGE_CLASS" --rsync-image "$KURL_UTIL_IMAGE" --preflight-validation-only || true)
-        if [ $? -ne 0 ]; then
+        if ! rook_scs_pvmigrate_dryrun_output=$($BIN_PVMIGRATE --source-sc "$rook_sc" --dest-sc "$OPENEBS_LOCALPV_STORAGE_CLASS" --rsync-image "$KURL_UTIL_IMAGE" --preflight-validation-only) ; then
             break
         fi
     done
--- a/addons/openebs/3.3.0/install.sh
+++ b/addons/openebs/3.3.0/install.sh
@@ -49,44 +49,47 @@ function openebs_maybe_migrate_from_rook() {
     if [ -z "$ROOK_VERSION" ]; then
         if kubectl get ns | grep -q rook-ceph; then
             # show validation errors from pvmigrate
-            # if there are errors, maybe_rook_to_openebs_migration_checks() will bail
-            maybe_rook_to_openebs_migration_checks
+            # if there are errors, openebs_maybe_rook_migration_checks() will bail
+            openebs_maybe_rook_migration_checks
             rook_ceph_to_sc_migration "$OPENEBS_LOCALPV_STORAGE_CLASS" "1"
             DID_MIGRATE_ROOK_PVCS=1 # used to automatically delete rook-ceph if object store data was also migrated
         fi
     fi
 }
 
-function maybe_rook_to_openebs_migration_checks() {
+function openebs_maybe_rook_migration_checks() {
 
     # get the list of StorageClasses that use rook-ceph
     rook_scs=$(kubectl get storageclass | grep rook | grep -v '(default)' | awk '{ print $1}') # any non-default rook StorageClasses
     rook_default_sc=$(kubectl get storageclass | grep rook | grep '(default)' | awk '{ print $1}') # any default rook StorageClasses
 
     # Ensure openebs-localpv-provisioner deployment is ready
-    printf "awaiting openebs-localpv-provisioner deployment\n"
+    echo "awaiting openebs-localpv-provisioner deployment"
     spinner_until 120 deployment_fully_updated openebs openebs-localpv-provisioner
 
-    echo "Running Rook to OpenEBS Migration Checks ..."
+    echo "running Rook to OpenEBS migration checks ..."
     local rook_scs_pvmigrate_dryrun_output
     local rook_default_sc_pvmigrate_dryrun_output
     for rook_sc in $rook_scs
     do
         # run validation checks for non default Rook storage classes
-        rook_scs_pvmigrate_dryrun_output=$($BIN_PVMIGRATE --source-sc "$rook_sc" --dest-sc "$OPENEBS_LOCALPV_STORAGE_CLASS" --rsync-image "$KURL_UTIL_IMAGE" --preflight-validation-only 2>/dev/null || true)
+        rook_scs_pvmigrate_dryrun_output=$($BIN_PVMIGRATE --source-sc "$rook_sc" --dest-sc "$OPENEBS_LOCALPV_STORAGE_CLASS" --rsync-image "$KURL_UTIL_IMAGE" --preflight-validation-only || true)
     done
 
     if [ -n "$rook_default_sc" ] ; then
         # run validation checks for Rook default storage class
-        rook_default_sc_pvmigrate_dryrun_output=$($BIN_PVMIGRATE --source-sc "$rook_default_sc" --dest-sc "$OPENEBS_LOCALPV_STORAGE_CLASS" --rsync-image "$KURL_UTIL_IMAGE" --preflight-validation-only 2>/dev/null || true)
+        rook_default_sc_pvmigrate_dryrun_output=$($BIN_PVMIGRATE --source-sc "$rook_default_sc" --dest-sc "$OPENEBS_LOCALPV_STORAGE_CLASS" --rsync-image "$KURL_UTIL_IMAGE" --preflight-validation-only || true)
     fi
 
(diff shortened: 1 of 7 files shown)


addons/openebs/3.3.0/install.sh
```diff
@@ -48,12 +48,58 @@ function openebs() {
...
+        rook_scs_pvmigrate_dryrun_output=$($BIN_PVMIGRATE --source-sc "$rook_sc" --dest-sc "$OPENEBS_LOCALPV_STORAGE_CLASS" --dry-run 2>/dev/null)
+    done
+
+    if [ -n "$rook_default_sc" ] ; then
+        # run validation checks for Rook default storage class
+        rook_default_sc_pvmigrate_dryrun_output=$($BIN_PVMIGRATE --source-sc "$rook_default_sc" --dest-sc "$OPENEBS_LOCALPV_STORAGE_CLASS" --dry-run 2>/dev/null)
+    fi
+
+    if { [ -n "$rook_scs_pvmigrate_dryrun_output" ] && [[ "$rook_scs_pvmigrate_dryrun_output" != *"ProvisioningFailed"* ]]; } || { [ -n "$rook_default_sc_pvmigrate_dryrun_output" ] && [[ "$rook_default_sc_pvmigrate_dryrun_output" != *"ProvisioningFailed"* ]]; }; then
+        printf "${YELLOW}"
```
camilamacedo86, line 81: Could we create a lib/add to an utils/common file functions such as the following example? So that, every time that we want print a waring that will be yellow, an error will be red and an info blue for example. Also, that can avoid we per mistake forget to reset the color. 

See: 

**Lib example**

```shell
# Turn colors in this script off by setting the NO_COLOR variable in your
# environment to any value:
NO_COLOR=${NO_COLOR:-""}
if [ -z "$NO_COLOR" ]; then
  warning_color=$'\e[1;33m' # required check what would be yellow here
  error_color=$'\e[0;31m'
  reset_color=$'\e[0m'
else
  warning_color=''
  error_color=''
  reset_color=''
fi

function log() { printf '%s\n' "$*"; }
function error() { error_text "ERROR:" $* >&2; }
function fatal() { error "$@"; exit 1; }

function warning_text {
  echo "$warning_color$*$reset_color"
}

function error_text {
  echo "$error_color$*$reset_color"
}

```

**Usage example**

```
 echo -e "my test for the error :\n${value}"

OR 

warning_text "My warning text will be outputted in yellow"
```

addons/openebs/3.3.0/install.sh
```diff
@@ -48,12 +48,58 @@ function openebs() {
...
+            # show validation errors from pvmigrate
+            # if there are errors, maybe_prompt_migrate_from_rook_dryrun_errors() will bail
+            maybe_prompt_migrate_from_rook_dryrun_errors
             rook_ceph_to_sc_migration "$OPENEBS_LOCALPV_STORAGE_CLASS"
             DID_MIGRATE_ROOK_PVCS=1 # used to automatically delete rook-ceph if object store data was also migrated
         fi
     fi
 }
 
+function maybe_prompt_migrate_from_rook_dryrun_errors() {
```
camilamacedo86, line 60: WDYT about something like rook_to_openebs_migration_checks only?

addons/openebs/template/base/install.sh
```diff
@@ -48,12 +48,61 @@ function openebs() {
...
+            # show validation errors from pvmigrate
+            # if there are errors, maybe_prompt_migrate_from_rook_dryrun_errors() will bail
+            maybe_prompt_migrate_from_rook_dryrun_errors
             rook_ceph_to_sc_migration "$OPENEBS_LOCALPV_STORAGE_CLASS"
             DID_MIGRATE_ROOK_PVCS=1 # used to automatically delete rook-ceph if object store data was also migrated
         fi
     fi
 }
 
+function maybe_prompt_migrate_from_rook_dryrun_errors() {
```
camilamacedo86, line 60: This code is duplicated. Could we not centralize it in a common.sh or util.sh under the dir?

kurl_util/cmd/pvmigrate/main.go
```diff
@@ -37,8 +38,19 @@ func main() {
...
+	flag.BoolVar(&printVersion, "version", false, "Print the version of the client")
+	flag.IntVar(&opts.PodReadyTimeout, "pod-ready-timeout", 90, "length of time to wait (in seconds) for volume validation pod(s) to go into Ready phase")
+
 	flag.Parse()
 
+	if printVersion {
+		fmt.Printf("Running pvmigrate build:\n")
+		version.Print()
+		os.Exit(0)
+	}
```
camilamacedo86, line 52: It seems that we need another subcommand instead of it.

kurl_util/cmd/pvmigrate/main.go
```diff
@@ -52,17 +64,51 @@ func main() {
...
 	}
 
-	if err = migrate.Migrate(ctx, logger, cli, opts); err != nil {
-		fmt.Printf("%s\n", err.Error())
-		os.Exit(1)
+	if !skipPvAccessModeCheck || dryRun {
+		unsupportedPVCs, err := validatePVAccessMode(ctx, logger, cfg, cli, opts)
+		if err != nil {
+			logger.Printf("failed to validate access modes for destination storage provider %s: %s", opts.DestSCName, err)
+			os.Exit(1)
```
camilamacedo86, line 78: same (return the error ideally or use logger.Fatalf() if required`: https://github.com/replicatedhq/kURL/pull/3767/files#r1030680619

kurl_util/cmd/pvmigrate/main.go
```diff
@@ -52,17 +64,51 @@ func main() {
...
+		}
+
+		if len(unsupportedPVCs) != 0 {
+			logger.Printf("PVC Access Mode Validation Failed: there are PVCs for storage class %s that cannot be mounted using the destination storage class %s.", opts.SourceSCName, opts.DestSCName)
+			migrate.PrintPVAccessModeErrors(unsupportedPVCs)
+			os.Exit(0)
```
camilamacedo86, line 84: same (return the error ideally or use logger.Fatalf() if required`: https://github.com/replicatedhq/kURL/pull/3767/files#r1030680619

Also, we can use q+ to print what is returned into `migrate.PrintPVAccessModeErrors(unsupportedPVCs)` formatted.

kurl_util/cmd/pvmigrate/main.go
```diff
@@ -52,17 +64,51 @@ func main() {
...
+		}
+	}
+
+	if !dryRun {
+		if err = migrate.Migrate(ctx, logger, cli, opts); err != nil {
+			fmt.Printf("%s\n", err.Error())
+			os.Exit(1)
```
camilamacedo86, line 91: same (return the error ideally or use logger.Fatalf() if required`: https://github.com/replicatedhq/kURL/pull/3767/files#r1030680619

camilamacedo86: Also, why the \n?

addons/openebs/3.3.0/install.sh
camilamacedo86: @rrpolanco after look it better see that the kURL has an standard/utils to print the warnings already (i.e. https://github.com/replicatedhq/kURL/blob/d8de81118f3a113036d3125733ce806d3421e8c1/addons/rook/1.5.11/install.sh#L43 )

Look:

Screenshot 2022-11-28 at 18 55 42

WDYT about we try to keep here all using the same approaches? So that if we want to improve or change that can be done horizontally.

emosbaugh (changes requested):

addons/openebs/template/base/install.sh
emosbaugh: the add-on version is generated from the template directory. it is duplicated in each add-on version. it is our pattern not to use shared code as we try not to change existing add-on versions to avoid breakage. this is likely a larger conversation if we want to change things at this level.

addons/openebs/3.3.0/install.sh
```diff
@@ -48,12 +48,58 @@ function openebs() {
...
+    echo "Running Rook to OpenEBS Migration Checks ..."
+
+    # get the list of StorageClasses that use rook-ceph
+    rook_scs=$(kubectl get storageclass | grep rook | grep -v '(default)' | awk '{ print $1}') # any non-default rook StorageClasses
+    rook_default_sc=$(kubectl get storageclass | grep rook | grep '(default)' | awk '{ print $1}') # any default rook StorageClasses
```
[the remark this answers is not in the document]
rrpolanco: There are no labels on the storage class so I can't use that approach.

addons/openebs/3.3.0/install.sh
rrpolanco: Good suggestion. I'll use `logFail()` to output the validation failures from the command.

addons/openebs/3.3.0/install.sh
rrpolanco: The `maybe_` prefix for naming functions is quite prevalent for functions which are contingent up performing some action in this project. I do agree this func needs to a better name, however.

kurl_util/cmd/pvmigrate/main.go
```diff
@@ -52,17 +64,51 @@ func main() {
 		os.Exit(1)
 	}
 
-	if !skipFreeSpaceCheck {
+	if !skipFreeSpaceCheck || dryRun {
 		if err := checkFreeSpace(ctx, logger, cfg, cli, opts); err != nil {
-			logger.Printf("Failed to check cluster free space: %s", err)
+			logger.Printf("failed to check cluster free space: %s", err)
 			os.Exit(1)
```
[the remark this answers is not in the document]
rrpolanco: good suggestion. We want to indicate that when any of the validation checks fail then that is considered a failure by this command and thus we return `os.Exit(10`.

kurl_util/cmd/pvmigrate/main.go
rrpolanco: I've update this.

kurl_util/cmd/pvmigrate/main.go
rrpolanco: Thanks. I changed this to use `logger.Fatalf()`

addons/openebs/3.3.0/install.sh
```diff
@@ -49,58 +49,47 @@ function openebs_maybe_migrate_from_rook() {
...
+            maybe_rook_to_openebs_migration_checks
+            rook_ceph_to_sc_migration "$OPENEBS_LOCALPV_STORAGE_CLASS" "1"
             DID_MIGRATE_ROOK_PVCS=1 # used to automatically delete rook-ceph if object store data was also migrated
         fi
     fi
 }
 
-function maybe_prompt_migrate_from_rook_dryrun_errors() {
-    echo "Running Rook to OpenEBS Migration Checks ..."
+function maybe_rook_to_openebs_migration_checks() {
```
emosbaugh, line 60: https://github.com/replicatedhq/kURL/blob/main/addons/README.md#requirements

> All shell functions defined in the add-on's install.sh script should be prefixed with the name of the add-on to avoid naming collisions, for example function rook_dashboard_ready_spinner().

```suggestion
function openebs_maybe_rook_to_openebs_migration_checks() {
```

emosbaugh (changes requested):

kurl_util/cmd/pvmigrate/main.go
laverya: is there a reason NOT to print the version?

We want to be able to figure out the version from the logs a customer sends us showing an error

addons/openebs/template/base/install.sh
```diff
@@ -48,13 +48,51 @@ function openebs() {
...
 
-function prompt_migrate_from_rook() {
+function openebs_maybe_rook_migration_checks() {
+
+    # get the list of StorageClasses that use rook-ceph
+    rook_scs=$(kubectl get storageclass | grep rook | grep -v '(default)' | awk '{ print $1}') # any non-default rook StorageClasses
+    rook_default_sc=$(kubectl get storageclass | grep rook | grep '(default)' | awk '{ print $1}') # any default rook StorageClasses
+
+    # Ensure openebs-localpv-provisioner deployment is ready
+    echo "awaiting openebs-localpv-provisioner deployment"
```
camilamacedo86, line 67: Not a blocker 
But could we not use as we have in kURL logStep logWarn logFail as well?

camilamacedo86 (commented):
Looking the code all shows fine for me.  🥇  

Just a nit https://github.com/replicatedhq/kURL/pull/3767/files#r1049440977 (not a blocker at all)

I think that would be nice get the lgtm and approval from @ricardomaraschini  before get merged because is who done the implementations of the commands used here

addons/openebs/template/base/install.sh
rrpolanco: We could. However, mapping messages to different colors is a lot more involved than I expected. I rather we have a separate PR for making the output more CLI UX friendly.

rrpolanco:
> Testgrid Run(s) Executing @ https://testgrid.kurl.sh/run/pr-3767-4df5b37-openebs-3.3.0-k8s-docker-localpv-2022-12-15T14:56:09Z

`localpv migrate from rook` openebs add-on test if failing because it doesn't take into account the installer changes in this PR. I have tested the rook to openebs migration locally and it works.

kurl_util/cmd/pvmigrate/main.go
rrpolanco: @laverya I added this to make the console messages more readable when the installer is doing an upgrade from rook to openebs. Basically, I wanted to seamlessly embed `pvmigrate` in the installer and abstract its usage.

Unfortunately, kURL console logs are rarely provided by the customer.

rrpolanco: @laverya I will push a change to print the `pvmigrate` change.

laverya (dismissed):

addons/openebs/3.3.0/install.sh
rrpolanco: I've addressed this.

addons/openebs/template/base/install.sh
```diff
@@ -48,13 +48,51 @@ function openebs() {
...
+    spinner_until 120 deployment_fully_updated openebs openebs-localpv-provisioner
+
+    echo "running Rook to OpenEBS migration checks ..."
+    local rook_scs_pvmigrate_dryrun_output
+    local rook_default_sc_pvmigrate_dryrun_output
+    for rook_sc in $rook_scs
+    do
+        # run validation checks for non default Rook storage classes
+        rook_scs_pvmigrate_dryrun_output=$($BIN_PVMIGRATE --source-sc "$rook_sc" --dest-sc "$OPENEBS_LOCALPV_STORAGE_CLASS" --rsync-image "$KURL_UTIL_IMAGE" --preflight-validation-only || true)
```
emosbaugh, line 76: this may overwrite a failed validation with a successful one since you are running it in a loop

addons/openebs/template/base/install.sh
```diff
@@ -48,13 +48,51 @@ function openebs() {
...
         fi
     fi
 }
 
-function prompt_migrate_from_rook() {
+function openebs_maybe_rook_migration_checks() {
+
+    # get the list of StorageClasses that use rook-ceph
+    rook_scs=$(kubectl get storageclass | grep rook | grep -v '(default)' | awk '{ print $1}') # any non-default rook StorageClasses
+    rook_default_sc=$(kubectl get storageclass | grep rook | grep '(default)' | awk '{ print $1}') # any default rook StorageClasses
```
emosbaugh, line 64:
```suggestion
    local rook_scs
    local rook_default_sc
    rook_scs=$(kubectl get storageclass | grep rook | grep -v '(default)' | awk '{ print $1}') # any non-default rook StorageClasses
    rook_default_sc=$(kubectl get storageclass | grep rook | grep '(default)' | awk '{ print $1}') # any default rook StorageClasses
```

kurl_util/cmd/pvmigrate/main.go
```diff
@@ -37,31 +42,60 @@ func main() {
...
+	// update migrate options with flag values
+	opts.PodReadyTimeout = time.Duration(podReadyTimeout) * time.Second
+	opts.DeletePVTimeout = time.Duration(deletePVTimeout) * time.Second
+
+	logger := log.New(os.Stderr, "", 0)
 	cfg, err := config.GetConfig()
 	if err != nil {
-		logger.Printf("failed to get config: %s\n", err.Error())
-		os.Exit(1)
+		logger.Fatalf("failed to get config: %s\n", err.Error())
```
emosbaugh, line 68:
```suggestion
		logger.Fatalf("failed to get config: %s", err)
```

kurl_util/cmd/pvmigrate/main.go
```diff
@@ -37,31 +42,60 @@ func main() {
...
 	}
 
 	cli, err := kubernetes.NewForConfig(cfg)
 	if err != nil {
-		logger.Printf("failed to create kubernetes clientset: %s\n", err.Error())
-		os.Exit(1)
+		logger.Fatalf("failed to create kubernetes clientset: %s\n", err.Error())
```
emosbaugh, line 73:
```suggestion
		logger.Fatalf("failed to create kubernetes clientset: %s", err)
```

kurl_util/cmd/pvmigrate/main.go
```diff
@@ -37,31 +42,60 @@ func main() {
...
 			os.Exit(1)
 		}
 	}
 
-	if err = migrate.Migrate(ctx, logger, cli, opts); err != nil {
-		fmt.Printf("%s\n", err.Error())
-		os.Exit(1)
+	if !preflightValidationOnly {
+		if err = migrate.Migrate(ctx, logger, cli, opts); err != nil {
+			logger.Fatalf("%s\n", err.Error())
```
emosbaugh, line 97:
```suggestion
			logger.Fatal(err)
```

kurl_util/cmd/pvmigrate/main.go
```diff
@@ -37,31 +42,60 @@ func main() {
...
+	flag.BoolVar(&printVersion, "version", false, "Print the version of the client")
+	flag.IntVar(&podReadyTimeout, "pod-ready-timeout", 60, "length of time to wait (in seconds) for volume validation pod(s) to go into Ready phase")
+	flag.IntVar(&deletePVTimeout, "delete-pv-timeout", 300, "length of time to wait (in seconds) for backing PV to be removed when temporary PVC is deleted")
+	flag.BoolVar(&skipPreflightValidation, "skip-preflight-validation", false, "skips pre-migration validation")
+	flag.BoolVar(&preflightValidationOnly, "preflight-validation-only", false, "skip the migration and run preflight validation only")
+
 	flag.Parse()
 
-	logger := log.New(os.Stdout, "", 0)
+	fmt.Printf("Running pvmigrate build:\n")
```
emosbaugh, line 53: This will get output as part of [these](https://github.com/replicatedhq/kURL/pull/3767/files#diff-a639b8e65bf2b4327bfb8a263798cad88f2c558c76e1dda96169edb3b7bcf4eeR85-R86) log lines. This should go to stderr using the logger.

emosbaugh (changes requested):

addons/openebs/template/base/install.sh
rrpolanco: Good point. I'll break out of the loop if the command fails and continue.

addons/openebs/template/base/install.sh
rrpolanco: curious, why is this better? I'd rather have two lines of code vs four.

kurl_util/cmd/pvmigrate/main.go
rrpolanco: why do you think the version info belongs in stdout?

kurl_util/cmd/pvmigrate/main.go
rrpolanco: Fixed.

kurl_util/cmd/pvmigrate/main.go
rrpolanco: Fixed.

kurl_util/cmd/pvmigrate/main.go
rrpolanco: Fixed.

kurl_util/cmd/pvmigrate/main.go
emosbaugh: i think it belongs in stderr but you have it going to stdout. in fact i would say it belongs in stdout if printVersion == true otherwise stderr

addons/openebs/template/base/install.sh
emosbaugh: trying not to pollute the global namespace

rrpolanco: ah, good point.

kurl_util/cmd/pvmigrate/main.go
```diff
@@ -37,31 +42,60 @@ func main() {
...
+	version.Print()
```
emosbaugh, line 54: this should go to the same as the line above

addons/openebs/template/base/install.sh
```diff
@@ -48,13 +48,56 @@ function openebs() {
...
+    spinner_until 120 deployment_fully_updated openebs openebs-localpv-provisioner
+
+    echo "running Rook to OpenEBS migration checks ..."
+    local rook_scs_pvmigrate_dryrun_output
+    local rook_default_sc_pvmigrate_dryrun_output
+    for rook_sc in $rook_scs
+    do
+        # run validation checks for non default Rook storage classes
+        rook_scs_pvmigrate_dryrun_output=$($BIN_PVMIGRATE --source-sc "$rook_sc" --dest-sc "$OPENEBS_LOCALPV_STORAGE_CLASS" --rsync-image "$KURL_UTIL_IMAGE" --preflight-validation-only || true)
+        if [ $? -ne 0 ]; then
```
emosbaugh, line 79: This will not work. let me see if i can find an example in the code where we have similar logic

```
$ bash test.sh
test.sh: line 5: adsfadsf: command not found
0
$ cat test.sh
#!/bin/bash

set -e

rook_scs_pvmigrate_dryrun_output=$(adsfadsf || true)
echo $?
```

addons/openebs/template/base/install.sh
```diff
@@ -48,13 +48,56 @@ function openebs() {
...
+            break
+        fi
```
emosbaugh, line 81: i think this will work but please test it

```suggestion
        if ! rook_scs_pvmigrate_dryrun_output=$($BIN_PVMIGRATE --source-sc "$rook_sc" --dest-sc "$OPENEBS_LOCALPV_STORAGE_CLASS" --rsync-image "$KURL_UTIL_IMAGE" --preflight-validation-only) ; then
            break
        fi
```

emosbaugh (changes requested):

addons/openebs/template/base/install.sh
rrpolanco: I see. It's the `|| true` that's coercing a return code of `0`.

addons/openebs/template/base/install.sh
rrpolanco: This works!

emosbaugh (dismissed):

emosbaugh (approved):

PR #3769, opened 2022-11-22
feat: default val for flags from env

ricardomaraschini:

this commits makes the object store sync sub command to start using environment variables values as default for flags.

with this it will become easier to start a container passing variables directly from secrets.


emosbaugh (commented):
Should use use viper with AutomaticEnv?

https://github.com/spf13/viper#working-with-environment-variables

PR #3784, opened 2022-11-29
feat: add kurl command line for disk space check

ricardomaraschini:

this commits adds a new kurl command to check the amount of available space, by storage class, in the cluster.

usage example:
```
kurl cluster free-disk-space [--storageclass openebs] [--openebs-node node] [--openebs-image eeacms/rsync:2.3]
```

output examples:

```
$ kubectl get storageclasses
NAME                PROVISIONER                     RECLAIMPOLICY   VOLUMEBINDINGMODE      ALLOWVOLUMEEXPANSION   AGE
default (default)   rook-ceph.rbd.csi.ceph.com      Delete          Immediate              true                   69m
openebs             openebs.io/local                Delete          WaitForFirstConsumer   false                  64m
rook-cephfs         rook-ceph.cephfs.csi.ceph.com   Delete          Immediate              true                   69m

$ /tmp/kurl cluster free-disk-space --storageclass openebs
62421426176     ip-172-31-60-177
62894407680     ip-172-31-80-103

$ # if no --storageclass flag is provided it looks for the default one.
$ /tmp/kurl cluster free-disk-space
85880889344
```

```release-note
NONE
```

NONE

### DIFF
--- a/pkg/cli/cluster_free_disk_space.go
+++ b/pkg/cli/cluster_free_disk_space.go
@@ -148,9 +148,8 @@ func NewClusterCheckFreeDiskSpaceCmd(cli CLI) *cobra.Command {
 	var selectedClass *storagev1.StorageClass
 
 	cmd := &cobra.Command{
-		Use:          "check-free-disk-space",
-		Short:        "List and analyse the available disk space for a given Storage Class.",
-		SilenceUsage: true,
+		Use:   "check-free-disk-space",
+		Short: "List and analyse the available disk space for a given Storage Class.",
 		Example: fmt.Sprintf(""+
 			"In the following examples 'openebs' is the name of a storage class backed by the %s storage provisioner while 'rook' is the name of a storage\n"+
 			"class backed by the %s or %s storage provisioners.\n\n"+
@@ -176,11 +175,21 @@ func NewClusterCheckFreeDiskSpaceCmd(cli CLI) *cobra.Command {
 			openEBSLocalProvisioner, rookRBDProvisioner, rookCephFSProvisioner,
 		),
 		PreRunE: func(cmd *cobra.Command, args []string) error {
-			k8sConfig := config.GetConfigOrDie()
-			clientSet = kubernetes.NewForConfigOrDie(k8sConfig)
-			rookClientSet = rookcli.NewForConfigOrDie(k8sConfig)
+			k8sConfig, err := config.GetConfig()
+			if err != nil {
+				return fmt.Errorf("failed to read kubernetes configuration: %w", err)
+			}
 
-			var err error
+			clientSet, err = kubernetes.NewForConfig(k8sConfig)
+			if err != nil {
+				return fmt.Errorf("failed to create kubernetes client: %w", err)
+			}
+
+			rookClientSet, err = rookcli.NewForConfig(k8sConfig)
+			if err != nil {
+				return fmt.Errorf("failed to create rook client: %w", err)
+			}
+
 			if selectedClass, err = getStorageClassByName(cmd.Context(), clientSet, forStorageClass); err != nil {
 				return err
 			}
--- a/pkg/cli/cluster_free_disk_space.go
+++ b/pkg/cli/cluster_free_disk_space.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"io"
 	"log"
+	"os"
 	"os/signal"
 	"syscall"
 
@@ -76,8 +77,12 @@ func hasEnoughSpace(node string, free, requested int64) (string, bool) {
 // evaluateOpenEBSFreeSpace checks how much space is available in a storage class backed by openEBSLocalProvisioner. biggerThan is
 // used to check if there is enough room in one node (if onNode != "") or in all nodes (onNode == ""). onNode is the node name, image
 // is the image to be used by the openebs disk free checker pod while the biggerThan is expressed in bytes.
-func evaluateOpenEBSFreeSpace(ctx context.Context, kubeCli kubernetes.Interface, image, scname, onNode string, biggerThan int64) error {
+func evaluateOpenEBSFreeSpace(ctx context.Context, kubeCli kubernetes.Interface, image, scname, onNode string, biggerThan int64, debug bool) error {
 	logger := log.New(io.Discard, "", 0)
+	if debug {
+		logger = log.New(os.Stderr, "", 0)
+	}
+
 	freeSpaceGetter, err := clusterspace.NewOpenEBSFreeDiskSpaceGetter(kubeCli, logger, image, scname)
 	if err != nil {
 		return fmt.Errorf("failed to start openebs free space getter: %w", err)
@@ -135,7 +140,12 @@ func evaluateRookFreeSpace(ctx context.Context, kubeCli kubernetes.Interface, ro
 	if free < requested {
 		return fmt.Errorf("not enough space on rook (requested %s, available %s)", requestedString, freeString)
 	}
-	fmt.Printf("Available disk space found in rook: %s\n", freeString)
+
+	message := fmt.Sprintf("Available disk space found in rook: %s", freeString)
+	if requested > 0 {
+		message = fmt.Sprintf("%s (requested %s)", message, requestedString)
+	}
+	fmt.Println(message)
 	return nil
 }
 
@@ -146,10 +156,12 @@ func NewClusterCheckFreeDiskSpaceCmd(cli CLI) *cobra.Command {
 	var clientSet kubernetes.Interface
 	var rookClientSet rookcli.Interface
 	var selectedClass *storagev1.StorageClass
+	var debug bool
 
 	cmd := &cobra.Command{
-		Use:   "check-free-disk-space",
-		Short: "List and analyse the available disk space for a given Storage Class.",
+		Use:          "check-free-disk-space",
+		Short:        "List and analyse the available disk space for a given Storage Class.",
+		SilenceUsage: true,
 		Example: fmt.Sprintf(""+
 			"In the following examples 'openebs' is the name of a storage class backed by the %s storage provisioner while 'rook' is the name of a storage\n"+
 			"class backed by the %s or %s storage provisioners.\n\n"+
@@ -190,6 +202,11 @@ func NewClusterCheckFreeDiskSpaceCmd(cli CLI) *cobra.Command {
 				return fmt.Errorf("failed to create rook client: %w", err)
 			}
 
+			debug, err = cmd.Flags().GetBool("debug")
+			if err != nil {
+				return fmt.Errorf("failed to read persistent debug flag: %w", err)
+			}
+
 			if selectedClass, err = getStorageClassByName(cmd.Context(), clientSet, forStorageClass); err != nil {
 				return err
 			}
@@ -212,7 +229,7 @@ func NewClusterCheckFreeDiskSpaceCmd(cli CLI) *cobra.Command {
 
 			switch selectedClass.Provisioner {
 			case openEBSLocalProvisioner:
-				return evaluateOpenEBSFreeSpace(ctx, clientSet, openEBSImage, selectedClass.Name, openEBSNode, biggerThanBytes)
+				return evaluateOpenEBSFreeSpace(ctx, clientSet, openEBSImage, selectedClass.Name, openEBSNode, biggerThanBytes, debug)
 
 			case rookCephFSProvisioner, rookRBDProvisioner:
 				return evaluateRookFreeSpace(ctx, clientSet, rookClientSet, selectedClass.Name, biggerThanBytes)
--- a/pkg/cli/cluster_free_disk_space.go
+++ b/pkg/cli/cluster_free_disk_space.go
@@ -1,0 +1,133 @@
+package cli
+
+import (
+	"bytes"
+	"fmt"
+	"log"
+
+	"github.com/spf13/cobra"
+	v1 "k8s.io/api/storage/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes"
+	"sigs.k8s.io/controller-runtime/pkg/client/config"
+
+	rookcli "github.com/rook/rook/pkg/client/clientset/versioned"
+
+	clusterspace "github.com/replicatedhq/kurl/pkg/cluster/space"
+)
+
+const (
+	// defaultOpenEBSPodImage is the image used during disk free check for openebs storage.
+	// this is the same image used by the pvmigrate project, it may be any image containing
+	// 'df' and 'cat' commands.
+	defaultOpenEBSPodImage          = "eeacms/rsync:2.3"
+	isDefaultStorageClassAnnotation = "storageclass.kubernetes.io/is-default-class"
+	openEBSLocalProvisioner         = "openebs.io/local"
+	rookRBDProvisioner              = "rook-ceph.rbd.csi.ceph.com"
+	rookCephFSProvisioner           = "rook-ceph.cephfs.csi.ceph.com"
+)
+
+// NewClusterCheckFreeDiskSpaceCmd returns a command that is capable of reporting back the amount of free space in the cluster
+// for a provided storage class.
+func NewClusterCheckFreeDiskSpaceCmd(cli CLI) *cobra.Command {
+	var forStorageClass, openEBSImage, openEBSNode string
+	logbuf := bytes.NewBuffer(nil)
+	logger := log.New(logbuf, "", 0)
+	k8sConfig := config.GetConfigOrDie()
+	clientSet := kubernetes.NewForConfigOrDie(k8sConfig)
+	rookClientSet := rookcli.NewForConfigOrDie(k8sConfig)
+
+	cmd := &cobra.Command{
+		Use:          "check-free-disk-space",
+		Long:         fmt.Sprintf("Returns the amount of free disk space (in bytes) for a given storage class\nSupported storage provisioners: %s, %s, %s", openEBSLocalProvisioner, rookRBDProvisioner, rookCephFSProvisioner),
+		Example:      "kurl cluster check-free-disk-space --storageclass openebs --openebs-node-name node0 --openebs-image ubuntu:latest",
+		SilenceUsage: true,
+		RunE: func(cmd *cobra.Command, args []string) (err error) {
+			classes, err := clientSet.StorageV1().StorageClasses().List(cmd.Context(), metav1.ListOptions{})
+			if err != nil {
+				return fmt.Errorf("failed to list cluster storage classes: %w", err)
+			}
+
+			var selectedClass *v1.StorageClass
+			for _, class := range classes.Items {
+				val, ok := class.Annotations[isDefaultStorageClassAnnotation]
+				if ok && val == "true" && forStorageClass == "" {
+					selectedClass = &class
+					logger.Printf("Selected default storage class %q", selectedClass.Name)
+					break
+				}
+
+				if class.Name == forStorageClass {
+					selectedClass = &class
+					logger.Printf("Selected storage class %q", selectedClass.Name)
+					break
+				}
+			}
+
+			if selectedClass == nil {
+				if forStorageClass == "" {
+					return fmt.Errorf("failed to find default storage class")
+				}
+				return fmt.Errorf("failed to find storage class %q", forStorageClass)
+			}
+
+			switch selectedClass.Provisioner {
+			case openEBSLocalProvisioner:
+				freeSpaceGetter, err := clusterspace.NewOpenEBSFreeDiskSpaceGetter(clientSet, logger, openEBSImage, selectedClass.Name)
+				if err != nil {
+					fmt.Print(logbuf.String())
+					return fmt.Errorf("failed to start openebs free space getter: %w", err)
+				}
+
+				volumes, err := freeSpaceGetter.OpenEBSVolumes(cmd.Context())
+				if err != nil {
+					fmt.Print(logbuf.String())
+					return fmt.Errorf("failed to get openebs free space: %w", err)
+				}
+
+				for node, volume := range volumes {
+					if openEBSNode == "" {
+						fmt.Printf("%d\t%s\n", volume.Free, node)
+						continue
+					}
+
+					if node != openEBSNode {
+						logger.Printf("Skipping node %q", node)
+						continue
+					}
+
+					fmt.Printf("%d\n", volume.Free)
+					return nil
+				}
+
+				if openEBSNode != "" {
+					fmt.Print(logbuf.String())
+					return fmt.Errorf("failed to collect openebs free space: node %q not found", openEBSNode)
+				}
+				return nil
+
+			case rookCephFSProvisioner, rookRBDProvisioner:
+				freeSpaceGetter, err := clusterspace.NewRookFreeDiskSpaceGetter(clientSet, rookClientSet, selectedClass.Name)
+				if err != nil {
+					return fmt.Errorf("failed to start rook free space getter: %w", err)
+				}
+
+				space, err := freeSpaceGetter.GetFreeSpace(cmd.Context())
+				if err != nil {
+					return fmt.Errorf("failed to get rook free space: %w", err)
+				}
+
+				fmt.Println(space)
+				return nil
+
+			default:
+				return fmt.Errorf("provisioner %q is not supported", selectedClass.Provisioner)
+			}
+		},
+	}
+
+	cmd.Flags().StringVar(&forStorageClass, "storageclass", "", "Inform the storage class name for which to check the free disk space. If not informed the default storage will be used.")
+	cmd.Flags().StringVar(&openEBSImage, "openebs-image", defaultOpenEBSPodImage, fmt.Sprintf("The image used by OpenEBS disk free evaluation pod. If not informed the default image used is %s", defaultOpenEBSPodImage))
+	cmd.Flags().StringVar(&openEBSNode, "openebs-node-name", "", "Show OpenEBS free disk space only for provided node.")
+	return cmd
+}
--- a/pkg/cli/cluster_free_disk_space.go
+++ b/pkg/cli/cluster_free_disk_space.go
@@ -2,24 +2,29 @@ package cli
 
 import (
 	"bytes"
+	"context"
 	"fmt"
+	"io"
 	"log"
+	"os/signal"
+	"syscall"
 
-	"github.com/spf13/cobra"
-	v1 "k8s.io/api/storage/v1"
+	storagev1 "k8s.io/api/storage/v1"
+	"k8s.io/apimachinery/pkg/api/resource"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/client-go/kubernetes"
 	"sigs.k8s.io/controller-runtime/pkg/client/config"
 
+	"code.cloudfoundry.org/bytefmt"
 	rookcli "github.com/rook/rook/pkg/client/clientset/versioned"
+	"github.com/spf13/cobra"
 
 	clusterspace "github.com/replicatedhq/kurl/pkg/cluster/space"
 )
 
 const (
-	// defaultOpenEBSPodImage is the image used during disk free check for openebs storage.
-	// this is the same image used by the pvmigrate project, it may be any image containing
-	// 'df' and 'cat' commands.
+	// defaultOpenEBSPodImage is the image used during disk free check for openebs storage. this is the same image used by the
+	// pvmigrate project, it may be any image containing 'df' and 'cat' commands.
 	defaultOpenEBSPodImage          = "eeacms/rsync:2.3"
 	isDefaultStorageClassAnnotation = "storageclass.kubernetes.io/is-default-class"
 	openEBSLocalProvisioner         = "openebs.io/local"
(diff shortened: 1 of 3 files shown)


laverya (dismissed):

pkg/cli/cluster_free_disk_space.go
```diff
@@ -0,0 +1,125 @@
...
+)
+
+const (
+	IsDefaultStorageClassAnnotation = "storageclass.kubernetes.io/is-default-class"
+	OpenEBSLocalProvisioner         = "openebs.io/local"
+	RookRBDProvisioner              = "rook-ceph.rbd.csi.ceph.com"
+	RookCephFSProvisioner           = "rook-ceph.cephfs.csi.ceph.com"
+)
+
+func NewClusterFreeDiskSpaceCmd(cli CLI) *cobra.Command {
```
camilamacedo86, line 26: Would be possible doc the `NewClusterFreeDiskSpaceCmd ` func?

pkg/cli/cluster_free_disk_space.go
```diff
@@ -0,0 +1,125 @@
...
+	k8sConfig := config.GetConfigOrDie()
+	clientSet := kubernetes.NewForConfigOrDie(k8sConfig)
+	rookClientSet := rookcli.NewForConfigOrDie(k8sConfig)
+
+	cmd := &cobra.Command{
+		Use:          "free-disk-space",
+		Short:        "Returns the amount of free disk space (in bytes) for a given storage class",
+		SilenceUsage: true,
+		RunE: func(cmd *cobra.Command, args []string) (err error) {
+			classes, err := clientSet.StorageV1().StorageClasses().List(cmd.Context(), metav1.ListOptions{})
```
camilamacedo86, line 39: Could we properly encapsulate the methods into validate() and then run()?

pkg/cli/commands.go
```diff
@@ -21,6 +21,7 @@ func AddCommands(cmd *cobra.Command, cli CLI) {
 
 	clusterCmd := NewClusterCmd(cli)
 	clusterCmd.AddCommand(NewClusterNodesMissingImageCmd(cli))
+	clusterCmd.AddCommand(NewClusterFreeDiskSpaceCmd(cli))
```
camilamacedo86, line 24: The name/UX here does not seems intuitive for me:

> kurl cluster free-disk-space [--storageclass openebs] [--openebs-node node] [--openebs-image eeacms/rsync:2.3]

Will it ensure that we have a free disk space or is it checking validating that we have a free space?

Therefore, I understand that we have many checks in commands such this one.
So, we could looking for in the future how we could better organize and etc the subcommands.

However, for now could we call this one check-free-disk-space? Is that OK? WDYT?

pkg/cli/cluster_free_disk_space.go
```diff
@@ -0,0 +1,125 @@
...
+	logbuf := bytes.NewBuffer(nil)
+	logger := log.New(logbuf, "", 0)
+	k8sConfig := config.GetConfigOrDie()
+	clientSet := kubernetes.NewForConfigOrDie(k8sConfig)
+	rookClientSet := rookcli.NewForConfigOrDie(k8sConfig)
+
+	cmd := &cobra.Command{
+		Use:          "free-disk-space",
+		Short:        "Returns the amount of free disk space (in bytes) for a given storage class",
+		SilenceUsage: true,
```
camilamacedo86, line 37: Why are using `SilenceUsage `?

pkg/cli/cluster_free_disk_space.go
```diff
@@ -0,0 +1,125 @@
...
+		RunE: func(cmd *cobra.Command, args []string) (err error) {
```
camilamacedo86, line 38: Could we add an Example desc with example over how to call it? Also, could we add an Long desc describing when/why/it is required useful?  

See an example
https://github.com/kubernetes-sigs/kubebuilder/blob/master/pkg/plugins/golang/v3/init.go#L64-L76

pkg/cli/cluster_free_disk_space.go
```diff
@@ -0,0 +1,125 @@
...
+				return nil
+
+			default:
+				return fmt.Errorf("provisioner %q is not supported", selectedClass.Provisioner)
+			}
+		},
+	}
+
+	cmd.Flags().StringVar(&forStorageClass, "storageclass", "", "The storage class for which to check the free disk space.")
+	cmd.Flags().StringVar(&openEBSImage, "openebs-image", "eeacms/rsync:2.3", "The image used by OpenEBS disk free evaluation pod.")
```
camilamacedo86, line 122: Why the value is `eeacms/rsync:2.3`?
What happens if we inform an invalid option?
Should we not validate the inputs? 

Also, can we run the command without inform this flags? If not, then we need mark then as required, see `MarkFlagRequired("name")`

camilamacedo86 (changes requested):
Could we try before merge: 

- ensure that required flags are marked as required
- invalid flag inputs will return an error
- we have an example desc so that is possible to know looking in the code how it is used
- we have a long desc describing when/why/where it is useful / required?
- we have any exported const or func properly comment so that new code would not fail in lint checks when we be able to enable them.

pkg/cli/cluster_free_disk_space.go
ricardomaraschini: fixed.

pkg/cli/cluster_free_disk_space.go
ricardomaraschini: Could you please point me to an example of this ?

pkg/cli/commands.go
ricardomaraschini: fixed.

pkg/cli/cluster_free_disk_space.go
ricardomaraschini: Copy and Pasta. What are you worried about ?

pkg/cli/cluster_free_disk_space.go
ricardomaraschini: `eeacms/rsync:2.3` is the default image used by `pvmigrate`, that is why it is here. if an invalid input is provided the command will fail and an error will be reported back. all flags are optional.

pkg/cli/cluster_free_disk_space.go
```diff
@@ -0,0 +1,125 @@
...
+	cmd.Flags().StringVar(&openEBSNode, "openebs-node", "", "Show OpenEBS free disk space only for provided node.")
```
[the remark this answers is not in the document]
ricardomaraschini: Added in the `Example` property.

pkg/cli/cluster_free_disk_space.go
ricardomaraschini: I have added and example to the `Example` property. Is that enough ?

pkg/cli/cluster_free_disk_space.go
```diff
@@ -0,0 +1,128 @@
...
+					return fmt.Errorf("failed to find default storage class")
+				}
+				return fmt.Errorf("failed to find storage class %q", forStorageClass)
+			}
+
+			switch selectedClass.Provisioner {
+			case openEBSLocalProvisioner:
+				freeSpaceGetter, err := clusterspace.NewOpenEBSFreeDiskSpaceGetter(clientSet, logger, openEBSImage, selectedClass.Name)
+				if err != nil {
+					fmt.Print(logbuf.String())
```
camilamacedo86, line 74: If that fails why what we output here?

pkg/cli/cluster_free_disk_space.go
```diff
@@ -0,0 +1,128 @@
...
+					return fmt.Errorf("failed to start openebs free space getter: %w", err)
+				}
+
+				volumes, err := freeSpaceGetter.OpenEBSVolumes(cmd.Context())
+				if err != nil {
+					fmt.Print(logbuf.String())
```
camilamacedo86, line 80: If that fails why what we output here?

ricardomaraschini: this is what have been logged by the "freeSpaceChecker".

pkg/cli/cluster_free_disk_space.go
ricardomaraschini: this is what have been logged by the "freeSpaceChecker".

pkg/cli/cluster_free_disk_space.go
```diff
@@ -0,0 +1,128 @@
...
+					return fmt.Errorf("failed to get openebs free space: %w", err)
+				}
+
+				for node, volume := range volumes {
+					if openEBSNode == "" {
+						fmt.Printf("%d\t%s\n", volume.Free, node)
```
camilamacedo86, line 86: Question how is the expected result? 

value
value
value 
error? 

Should we not show the total? Could you please add an screenshot of your test for we see how that would to be?

ricardomaraschini: No, we must show per node (for openebs). There is an example in the description of this PR.

pkg/cli/cluster_free_disk_space.go
camilamacedo86: I thought in something like: 


```go
// NewRookFreeDiskSpaceGetter returns a command that is capable of reporting back the amount of free space in the cluster
// for a provided storage class.
func NewClusterFreeDiskSpaceCmd(cli CLI) *cobra.Command {
	var forStorageClass, openEBSImage, openEBSNode string
	logger = log.New(logbuf, "", 0)

	cmd := &cobra.Command{
		Use:          "free-disk-space",
		Short:        "Returns the amount of free disk space (in bytes) for a given storage class",
		SilenceUsage: true,
		PreRunE: func(cmd *cobra.Command, args []string) (err error) {
			k8sConfig = config.GetConfigOrDie()
			clientSet = kubernetes.NewForConfigOrDie(k8sConfig)
			rookClientSet = rookcli.NewForConfigOrDie(k8sConfig)
		},
		RunE: func(cmd *cobra.Command, args []string) (err error) {
			// See that here we can do what is requited to be done
			defaultClassProvisioner,err:= findDefaultClass()
			if err != nil {
				return err
			}
			output,err:= runCheck(defaultClassProvisioner);
			if err != nil {
				return err
			}

			// OUTPUT the result
			fmt.Sprintf("the availble free disk space with thr provisioner %s is: (%s)",
				output,defaultClassProvisioner.Name)
			return nil
		},
	}

	cmd.Flags().StringVar(&forStorageClass, "storageclass", "", "The storage class for which to check the free disk space.")
	cmd.Flags().StringVar(&openEBSImage, "openebs-image", "eeacms/rsync:2.3", "The image used by OpenEBS disk free evaluation pod.")
	cmd.Flags().StringVar(&openEBSNode, "openebs-node", "", "Show OpenEBS free disk space only for provided node.")
	return cmd
}

func runCheck(*v1.StorageClass) (string, err){
	switch selectedClass.Provisioner {
	case openEBSLocalProvisioner:
		....

	default:
		return fmt.Errorf("provisioner %q is not supported", selectedClass.Provisioner)
	}
}

func findDefaultClass() (*v1.StorageClass, error) {
	var selectedClass *v1.StorageClass
	classes, err := clientSet.StorageV1().StorageClasses().List(cmd.Context(), metav1.ListOptions{})
	if err != nil {
		return nil, fmt.Errorf("failed to list cluster storage classes: %w", err)
	}

	for _, class := range classes.Items {
		val, ok := class.Annotations[isDefaultStorageClassAnnotation]
		if ok && val == "true" && forStorageClass == "" {
			selectedClass = &class
			logger.Printf("Selected default storage class %q", selectedClass.Name)
			break
		}

		if class.Name == forStorageClass {
			selectedClass = &class
			logger.Printf("Selected storage class %q", selectedClass.Name)
			break
		}
	}

	if selectedClass == nil {
		if forStorageClass == "" {
			return fmt.Errorf("failed to find default storage class")
		}
		return fmt.Errorf("failed to find storage class %q", forStorageClass)
	}
	return selectedClass,
}
```

WDYT? Is that make sense?

pkg/cli/cluster_free_disk_space.go
camilamacedo86: If we change the version would still be that the default value?
Could we create a const on top and set the default value with a comment on top for help in the future if we need to understand it?

pkg/cli/cluster_free_disk_space.go
camilamacedo86: @ricardomaraschini see that in this implementation we will

print value 
print value
error found and returned / stop

Example:

$ /tmp/kurl cluster free-disk-space --storageclass openebs
62421426176     ip-172-31-60-177
62894407680     ip-172-31-80-103
Error 

Therefore, I think we should change that for return the error only after check all values OR just return the output when errors are not found. WDYT is that make sense?

ricardomaraschini: This output won't happen. Or we print the output or the error. Why do you think it is possible ?

camilamacedo86: HI @ricardomaraschini, 

> This output won't happen. Or we print the output or the error. Why do you think it is possible ?

We have a loop and for each time we print the value see:
> fmt.Printf("%d\t%s\n", volume.Free, node) 
Then, if an error occurs in the loop we return an error and we stop to check.

So, how the scenario above will not happens?
To do what you are saying here `Or we print the output or the error` we can not use fmt.Print in the loop. We need to store the values in array and just print them after all process.`

pkg/cli/cluster_free_disk_space.go
camilamacedo86: fmt.Print will print before that fails right?

ricardomaraschini: Yes, this is the log of the freeSpaceGetter and it makes sense to print it if something goes wrong.

camilamacedo86: OK . You told that the idea is print the value that we have + the errors so I think we can close this one.
Thank you for the clarifications.

pkg/cli/cluster_free_disk_space.go
camilamacedo86: I see here we will continue. Sorry, I did not look this one properly.

pkg/cli/cluster_free_disk_space.go
camilamacedo86: If we want to not show the usage when we did not use the flags accordingly or an error occurs that is fine 👍

pkg/cli/cluster_free_disk_space.go
```diff
@@ -0,0 +1,133 @@
...
+	rookcli "github.com/rook/rook/pkg/client/clientset/versioned"
+
+	clusterspace "github.com/replicatedhq/kurl/pkg/cluster/space"
+)
+
+const (
+	// defaultOpenEBSPodImage is the image used during disk free check for openebs storage.
+	// this is the same image used by the pvmigrate project, it may be any image containing
+	// 'df' and 'cat' commands.
+	defaultOpenEBSPodImage          = "eeacms/rsync:2.3"
```
camilamacedo86, line 23: How we will keep it maintained?
Will we need to update the default value when pvmigrate change it there?
Can we just use it from there or at least add a TODO for we export this const there and get here?
Is that something that make sense?

ricardomaraschini: We don't need to keep them both in sync, no. There is a flag that can be used to replace the image as well. What else to you think is necessary ? If you prefer I can simply remove the default and let then customise in the installer.

pkg/cli/cluster_free_disk_space.go
ricardomaraschini: That is fine, it is in use in other commands of the same binary.

pkg/cli/cluster_free_disk_space.go
camilamacedo86: I think the default value is great and it probably is the latest release one.
Yes, i see that we can use the flag but would be great if we do not need update that manually when a new release for example happens. However, that is not a blocker for we get this one merged.

pkg/cli/cluster_free_disk_space.go
camilamacedo86: The suggestion look for cohesion so we can grow in the future. 
This method has more than 30 lines which would not pass in a lint check for complexity for example. 

Anyway, I will not block the PR because of this.

camilamacedo86 (dismissed):

pkg/cli/cluster_free_disk_space.go
```diff
@@ -0,0 +1,133 @@
...
+
+			default:
+				return fmt.Errorf("provisioner %q is not supported", selectedClass.Provisioner)
+			}
+		},
+	}
+
+	cmd.Flags().StringVar(&forStorageClass, "storageclass", "", "The storage class for which to check the free disk space.")
+	cmd.Flags().StringVar(&openEBSImage, "openebs-image", defaultOpenEBSPodImage, "The image used by OpenEBS disk free evaluation pod.")
+	cmd.Flags().StringVar(&openEBSNode, "openebs-node", "", "Show OpenEBS free disk space only for provided node.")
```
camilamacedo86, line 131: @ricardomaraschini here, how can we inform the node? Is an IP, is an label? Can we add an example?

pkg/cli/cluster_free_disk_space.go
```diff
@@ -0,0 +1,133 @@
...
+	logbuf := bytes.NewBuffer(nil)
+	logger := log.New(logbuf, "", 0)
+	k8sConfig := config.GetConfigOrDie()
+	clientSet := kubernetes.NewForConfigOrDie(k8sConfig)
+	rookClientSet := rookcli.NewForConfigOrDie(k8sConfig)
+
+	cmd := &cobra.Command{
+		Use:          "check-free-disk-space",
+		Short:        "Returns the amount of free disk space (in bytes) for a given storage class",
+		Example:      "kurl cluster check-free-disk-space --storageclass openebs",
```
camilamacedo86, line 43:
```suggestion
		Example:      "kurl cluster check-free-disk-space --storageclass openebs",
```

Can we improve this example by adding all possible flags with values? So that we know how to use it?

pkg/cli/cluster_free_disk_space.go
```diff
@@ -0,0 +1,133 @@
...
+				fmt.Println(space)
+				return nil
+
+			default:
+				return fmt.Errorf("provisioner %q is not supported", selectedClass.Provisioner)
+			}
+		},
+	}
+
+	cmd.Flags().StringVar(&forStorageClass, "storageclass", "", "The storage class for which to check the free disk space.")
```
camilamacedo86, line 129:
```suggestion
	cmd.Flags().StringVar(&forStorageClass, "storageclass", "", fmt.Sprintf("Inform the storage class for which to check the free disk space. [Options: %s, %s, %s]", openEBSLocalProvisioner, rookRBDProvisioner, rookCephFSProvisioner ))
```

	isDefaultStorageClassAnnotation = "storageclass.kubernetes.io/is-default-class"
	openEBSLocalProvisioner         = "openebs.io/local"
	rookRBDProvisioner              = "rook-ceph.rbd.csi.ceph.com"
	rookCephFSProvisioner           = "rook-ceph.cephfs.csi.ceph.com"

pkg/cli/cluster_free_disk_space.go
```diff
@@ -0,0 +1,133 @@
...
+	cmd.Flags().StringVar(&openEBSImage, "openebs-image", defaultOpenEBSPodImage, "The image used by OpenEBS disk free evaluation pod.")
```
camilamacedo86, line 130:
```suggestion
	cmd.Flags().StringVar(&openEBSImage, "openebs-image", defaultOpenEBSPodImage, fmt.Sprintf("The image used by OpenEBS disk free evaluation pod. If not informed the default image used is %s", defaultOpenEBSPodImage))
```

camilamacedo86 (commented):
I added just a few nits for we improve the flags desc and examples.
Otherwise, 
/lgtm

camilamacedo86 (approved):
Approved 🥇 

But see if you agree to address the nits for we improve the examples and flags description.
Please, feel free to ping me out to approve again as well.

pkg/cli/cluster_free_disk_space.go
ricardomaraschini: Made the node flag clear (it is by node's name). Added to the sample. Described the supported provisioners.

pkg/cli/cluster_free_disk_space.go
ricardomaraschini: Done.

pkg/cli/cluster_free_disk_space.go
ricardomaraschini: I agree with adding complexity when it is necessary and I don't consider that this is the case. Moreover 30 lines for go (where possibly half would be error treatment) is not a long function.

pkg/cli/cluster_free_disk_space.go
ricardomaraschini: As this is not a blocker I am going to resolve it.

pkg/cli/cluster_free_disk_space.go
ricardomaraschini: As this is not a blocker I am resolving it.

camilamacedo86 (dismissed):

pkg/cli/cluster_free_disk_space.go
```diff
@@ -0,0 +1,133 @@
...
+	rookRBDProvisioner              = "rook-ceph.rbd.csi.ceph.com"
+	rookCephFSProvisioner           = "rook-ceph.cephfs.csi.ceph.com"
+)
+
+// NewClusterCheckFreeDiskSpaceCmd returns a command that is capable of reporting back the amount of free space in the cluster
+// for a provided storage class.
+func NewClusterCheckFreeDiskSpaceCmd(cli CLI) *cobra.Command {
+	var forStorageClass, openEBSImage, openEBSNode string
+	logbuf := bytes.NewBuffer(nil)
+	logger := log.New(logbuf, "", 0)
```
emosbaugh, line 35: what do you think about using cmd.ErrOrStderr() here?

pkg/cli/cluster_free_disk_space.go
```diff
@@ -0,0 +1,133 @@
...
+	k8sConfig := config.GetConfigOrDie()
+	clientSet := kubernetes.NewForConfigOrDie(k8sConfig)
+	rookClientSet := rookcli.NewForConfigOrDie(k8sConfig)
```
emosbaugh, line 38: can you instantiate these in the run() function?

emosbaugh (changes requested):

pkg/cli/cluster_free_disk_space.go
camilamacedo86: IHMO it should to be PreRunE since is a pre-requirement as suggested https://github.com/replicatedhq/kURL/pull/3784#discussion_r1034843469

I know that the methods here will exit(1) if fails but I also think that should be inside of the command so that if errors occurs we cobra would get that. ( i.e it is within silenceUsage is true but if want to remove that, cobra would output the usage/help for the command as well )

emosbaugh: im fine with prerune. i just dont like that to get the help text you have to be able to instantiate the kube client

pkg/cli/cluster_free_disk_space.go
```diff
@@ -0,0 +1,223 @@
...
+
+			default:
+				fmt.Printf("Provisioner %q is not supported, unable to determine free space.\n", selectedClass.Provisioner)
+				return nil
+			}
+		},
+	}
+
+	cmd.Flags().StringVar(&forStorageClass, "storageclass", "", "Inform the storage class name for which to check the free disk space. If not informed the default storage will be used.")
+	cmd.Flags().StringVar(&biggerThanString, "bigger-than", "", "Compares if the cluster free disk space is bigger than the provided value. Accepts the same format as used when defining storage requests in Kubernetes (e.g. 10G, 5Gi, 500M).")
```
camilamacedo86, line 219: Why do we need this one?
Do we really a command that checks if the value is > than an a value ? 
Can not this check be done where this info is useful?

pkg/cli/cluster_free_disk_space.go
```diff
@@ -0,0 +1,223 @@
...
+
+			if biggerThanString == "" {
+				return nil
+			}
+
+			parsed, err := resource.ParseQuantity(biggerThanString)
+			if err != nil {
+				return fmt.Errorf("failed to parse %s as a quantity: %w", biggerThanString, err)
+			}
+			biggerThanBytes = parsed.Value()
```
camilamacedo86, line 205: I do not think that we need this implementation here.
If we will use the check in a place where is required to know if the amount is or not bigger than X then this code can be there. I have a command that check the free space why this command also will check if the free space is bigger than a value informed? It seems out of the scope. Could we please remove this one?

pkg/cli/cluster_free_disk_space.go
```diff
@@ -0,0 +1,223 @@
...
+			openEBSLocalProvisioner, rookRBDProvisioner, rookCephFSProvisioner,
+		),
+		Long: fmt.Sprintf(""+
+			"This program returns the amount of free disk space (in bytes) for a given storage class or compares if there is enough space to hold a certain\n"+
+			"amount of data (--bigger-than). For OpenEBS, when no --bigger-than flag is provided, this program returns a list of nodes and their respective\n"+
+			"free space, while for Rook storage only the available space is returned. When --bigger-than flag is used this program sets the exit code to zero\n"+
+			"if the space available is bigger than the quantity provided. For OpenEBS, if no node has been provided through --openebs-node-name the exit code\n"+
+			"will be zero only if all nodes free space are bigger than the provided quantity (see Examples section for more details).\n\n"+
+			"Supports the following storage provisioners: %s, %s, %s",
+			openEBSLocalProvisioner, rookRBDProvisioner, rookCephFSProvisioner,
```
camilamacedo86, line 175: Same of above can we use a const for the text here?

pkg/cli/cluster_free_disk_space.go
```diff
@@ -0,0 +1,232 @@
...
+
+			if biggerThanString == "" {
+				return nil
+			}
+
+			parsed, err := resource.ParseQuantity(biggerThanString)
+			if err != nil {
+				return fmt.Errorf("failed to parse %s as a quantity: %w", biggerThanString, err)
+			}
+			biggerThanBytes = parsed.Value()
```
camilamacedo86, line 205:
```suggestion
```
I do not think that we need this implementation here.
If we will use the check in a place where is required to know if the amount is or not bigger than X then this code can be there. I have a command that check the free space why this command also will check if the free space is bigger than a value informed? It seems out of the scope. Could we please remove this one?

pkg/cli/cluster_free_disk_space.go
```diff
@@ -0,0 +1,232 @@
...
+
+			default:
+				fmt.Printf("Provisioner %q is not supported, unable to determine free space.\n", selectedClass.Provisioner)
+				return nil
+			}
+		},
+	}
+
+	cmd.Flags().StringVar(&forStorageClass, "storageclass", "", "Inform the storage class name for which to check the free disk space. If not informed the default storage will be used.")
+	cmd.Flags().StringVar(&biggerThanString, "bigger-than", "", "Compares if the cluster free disk space is bigger than the provided value. Accepts the same format as used when defining storage requests in Kubernetes (e.g. 10G, 5Gi, 500M).")
```
camilamacedo86, line 228:
```suggestion
```
Why do we need this one?
Do we really a command that checks if the value is > than an a value ? 
Can not this check be done where this info is useful?

pkg/cli/cluster_free_disk_space.go
```diff
@@ -0,0 +1,232 @@
...
+			if err != nil {
+				return fmt.Errorf("failed to create kubernetes client: %w", err)
+			}
+
+			rookClientSet, err = rookcli.NewForConfig(k8sConfig)
+			if err != nil {
+				return fmt.Errorf("failed to create rook client: %w", err)
+			}
+
+			if selectedClass, err = getStorageClassByName(cmd.Context(), clientSet, forStorageClass); err != nil {
```
camilamacedo86, line 193: We are just checking if the value is or not supported above.
Ideally we do like:
a) First check all inputs ( return an error if something fails )
b) After we check the inputs we run the command. In this case we can return an error here before to check if the values are or nor supported that seems a little strange.

pkg/cli/cluster_free_disk_space.go
```diff
@@ -0,0 +1,232 @@
...
+	freeString := bytefmt.ByteSize(uint64(free))
+	if free >= requested {
+		message := fmt.Sprintf("Node %s has %s available", node, freeString)
+		if requested > 0 {
+			message = fmt.Sprintf("%s (requested %s)", message, requestedString)
+		}
+		return message, true
+	}
+	return fmt.Sprintf("Not enough space on node %s (requested %s, available %s)", node, requestedString, freeString), false
+}
```
camilamacedo86, line 74:
```suggestion
```

Why we would add that in the check? 
What is the motivation?

pkg/cli/cluster_free_disk_space.go
ricardomaraschini: @emosbaugh can you take a look at the new version and see if this still applies?

pkg/cli/cluster_free_disk_space.go
```diff
@@ -0,0 +1,223 @@
...
+			"# reports the available space in rook\n"+
+			"kurl cluster check-free-disk-space --storageclass rook\n\n"+
+			"# reports the available space in all nodes\n"+
+			"kurl cluster check-free-disk-space --storageclass openebs\n\n"+
+			"# checks if there is 10G available on node node0\n"+
+			"kurl cluster check-free-disk-space --storageclass openebs --openebs-node-name node0 --openebs-image ubuntu:latest --bigger-than 10G\n\n"+
+			"# checks if there is 10G available on all nodes in the cluster\n"+
+			"kurl cluster check-free-disk-space --storageclass openebs --bigger-than 10G\n"+
+			"# checks if there is 20G available in the cluster on the default storage class\n"+
+			"kurl cluster check-free-disk-space --bigger-than 20G\n",
```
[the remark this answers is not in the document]
ricardomaraschini: No. Feel free to open a PR to address this once this is merged.

pkg/cli/cluster_free_disk_space.go
ricardomaraschini: Same answer.

pkg/cli/cluster_free_disk_space.go
ricardomaraschini: See my previous answers

emosbaugh (dismissed):

camilamacedo86 (approved):

pkg/cli/cluster_free_disk_space.go
ricardomaraschini: @emosbaugh I have used the `debug` flag to enable the output to stderr. Can you take a look ?

camilamacedo86 (approved):

PR #3798, opened 2022-12-02
WIP: fix - remove docker.io package to install containerd

camilamacedo86:

To fix issue to upgrade or install when docker package versions are installed. (ubuntu)
(docker.io does not have been purge if installed)

Fixes # [60835]

Local tests done: (see that now we are removing the packages when/if the docker is installed)

Screenshot 2022-12-02 at 21 43 25

## Steps to reproduce

```
$ sudo apt-get update
```

```release-note
fix issue to install or upgrade to containerd.io when host has docker.io package installed. (ubuntu) 
```

NONE


emosbaugh (commented):
are the packages in this doc relevant to the story?

https://docs.docker.com/engine/install/ubuntu/#uninstall-old-versions

```
 sudo apt-get remove docker docker-engine docker.io containerd runc
```

the error message states the conflicting package is "containerd"

```
 containerd.io conflicts with containerd
```

camilamacedo86:
Hi @emosbaugh,

Thank you for your attention and time. Following the comments inline

> are the packages in this doc relevant to the story?
> https://docs.docker.com/engine/install/ubuntu/#uninstall-old-versions
>  sudo apt-get remove docker docker-engine [docker.io](http://docker.io/) containerd runc

We are using dpkg to install containerd and to remove the docker packages already see:

https://github.com/replicatedhq/kURL/blob/3b512e30381fcb3934cd0f3b455e4f027f05274e/scripts/common/docker.sh#L108-L113

So, in order to keep the consistence I fixed the issue using the same manager pkg tool instead of try to achieve it with apt-get. Therefore, regards the error:

> the error message states the conflicting package is "containerd"
> [containerd.io](http://containerd.io/) conflicts with containerd

It shows to be a bug where the docker packages are not removed automatically, see: 

- https://github.com/moby/moby/issues/38185#issuecomment-638596621
- https://bugs.launchpad.net/ubuntu/+source/docker.io/+bug/1939140
- https://bugs.launchpad.net/ubuntu/+source/docker.io/+bug/1940920
- https://unix.stackexchange.com/questions/611228/getting-series-of-file-conflicts-like-runc-and-containerd-when-trying-to-install
- https://askubuntu.com/questions/1273024/docker-io-depends-containerd-1-2-6-0ubuntu1

So, the less complex solutions seems to be only remove the packages before installing containerd to avoid the conflict.
I supplemented the changes by adding a comment to clarify it as well. 

I hope that the above info answer your questions. Please, let me know if you agree with or if you would like to propose any other solution to workaround the issue and achieve the same goal.

scripts/common/host-packages.sh
```diff
@@ -114,6 +114,13 @@ function _dpkg_install_host_packages() {
         return 0
     fi
 
+    # Remove docker packages, if installed to install containerd
```
camilamacedo86, line 117:
```suggestion
    # Remove docker packages, if installed to install containerd
    # It is required because of an bug where the package manager is unable to sort it out and 
    # the installation/upgrade will fails with `dpkg: no, cannot proceed with removal of containerd ... docker.io     
    # depends on containerd (>= 1.2.6-0ubuntu1~)  containerd is to be removed.` 
    # More info: https://bugs.launchpad.net/ubuntu/+source/docker.io/+bug/1940920
    # https://bugs.launchpad.net/ubuntu/+source/docker.io/+bug/1939140       
```

emosbaugh (approved):

PR #3803, opened 2022-12-05
test getting+filtering old items in the kurl bucket without deleting them

laverya:

Upon each merge, after starting testgrid, list the files from old staging kurl releases (>31d). Then, print the list.

In a future PR, delete items in the list.

Fixes #

```release-note

```

### DIFF
--- a/.github/workflows/deploy-staging.yaml
+++ b/.github/workflows/deploy-staging.yaml
@@ -202,3 +202,21 @@ jobs:
         MSG="Testgrid Run(s) Executing @ https://testgrid.kurl.sh/run/${REF}"
         echo "::set-output name=msg::${MSG}"
         echo "::notice ::${MSG}"
+
+  cleanup-s3:
+    runs-on: ubuntu-20.04
+    needs:
+      - testgrid-run
+    steps:
+      - name: Checkout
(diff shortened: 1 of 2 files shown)


ricardomaraschini (approved):

.github/workflows/deploy-staging.yaml
```diff
@@ -202,3 +202,21 @@ jobs:
...
+      - name: Checkout
+        uses: actions/checkout@v3
+
+
+      - name: cleanup versions in staging S3 more than 30 days old
+        run: |
+          bin/cleanup-staging.sh
+        env:
+          S3_BUCKET: kurl-sh
+          AWS_ACCESS_KEY_ID: ${{ secrets.AWS_PROD_ACCESS_KEY_ID }}
```
camilamacedo86, line 220: the problem with the secrets is that we cannot allow pull request targets from a fork.
So we need to add a conditions such as : https://github.com/replicatedhq/kURL/blob/9daf09ec243416ee936d496bbfd5bfe056f7d911/.github/workflows/test-addon-pr.yaml#L19

laverya: This only runs after merge

PR #3809, opened 2022-12-05
Add missing code to rook 1.10.6 add-on

emosbaugh:

Updates were made to the rook template function after the pr was made for rook 1.10.6. This brings in missing code.

### DIFF
--- a/addons/rook/1.10.6/install.sh
+++ b/addons/rook/1.10.6/install.sh
@@ -65,7 +65,6 @@ function rook() {
     # Disable EKCO updates
     # Disallow the EKCO operator from updating Rook custom resources during a Rook upgrade
     rook_disable_ekco_operator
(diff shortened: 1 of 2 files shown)


addons/rook/1.10.6/install.sh
```diff
@@ -64,8 +64,8 @@ function rook() {
 
     # Disable EKCO updates
     # Disallow the EKCO operator from updating Rook custom resources during a Rook upgrade
-    # EKCO will be enabled (i.e. deployment scaled up) again when the EKCO add-on is applied
-    disable_ekco_operator
+    rook_disable_ekco_operator
+    ROOK_DID_DISABLE_EKCO_OPERATOR=1
```
laverya, line 68: Should this env var be set inside the function?

laverya: instead of unconditionally

laverya (approved):

PR #3813, opened 2022-12-06
bug: missing kubectl during add-on pre init

ricardomaraschini:

When fresh installing the kubectl binary is not yet available during the add-on pre init execution. Avoid to fail there otherwise the install process can not move forward.


ricardomaraschini:
I will be reopening this in a moment. This is pretty much imcomplete.

PR #3825, opened 2022-12-08
chore: adding more test cases for minio migration

ricardomaraschini:

this commits adds some test scenarios and introduce some small changes to make debug easier:

- migration from 2020-01-25T02-50-51Z using openebs.
- migration from 2020-01-25T02-50-51Z while increasing claim size.
- showing up longhorn pod logs for easier debug in case of failure.

### DIFF
--- a/addons/minio/template/testgrid/k8s-docker.yaml
+++ b/addons/minio/template/testgrid/k8s-docker.yaml
@@ -166,3 +166,84 @@
     minio_object_store_info
     validate_testfile rwtest testfile.txt
     validate_read_write_object_store postupgrade upgradefile.txt
+    dump_longhorn_logs
+
+# upgrade from 2020-01-25T02-50-51Z using openebs
+- name: upgrade minio from 2020-01-25T02-50-51Z (openebs)
+  flags: "yes"
+  installerSpec:
+    kubernetes:
+      version: "latest"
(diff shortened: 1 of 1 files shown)


addons/minio/template/testgrid/k8s-docker.yaml
```diff
@@ -166,3 +166,84 @@
...
+    source /opt/kurl-testgrid/testhelpers.sh
+    minio_object_store_info
+    validate_testfile rwtest testfile.txt
+    validate_read_write_object_store postupgrade upgradefile.txt
+
+- name: upgrade minio from 2020-01-25T02-50-51Z while increasing PVC claim size (openebs)
+  flags: "yes"
+  installerSpec:
+    kubernetes:
+      version: "latest"
```
camilamacedo86, line 215: Should we not install 1.23.x or 1.24.x here to do the check?

camilamacedo86 (approved):
Just a question. Otherwise shows fine.

addons/minio/template/testgrid/k8s-docker.yaml
ricardomaraschini: I don't think this is very relevant because what we are testing here is the Minio upgrade.

PR #3834, opened 2022-12-09
feat: warning when docker be installed that it is deprecated for k8s versions >= 1.20

camilamacedo86:

If users opt-in to use docker that ought to know that it is no longer recommended. 

## Steps to reproduce

Select docker to be installed.  

Screenshot 2022-12-09 at 07 18 10

```release-note
feat: warning users about k8s docker deprecation in 1.20 and recommend containerd instead.  
```

NONE


scripts/common/docker.sh
```diff
@@ -37,6 +37,8 @@ function install_docker() {
         if [ -z "$DOCKER_VERSION" ]; then
             printf "${RED}The installer did not specify a version of Docker to include, but is required by all kURL installation scripts currently. The latest supported version of Docker will be installed.${NC}\n"
             DOCKER_VERSION="19.03.4"
+        else
+            printf "${YELLOW}Note that Kubernetes deprecate docker in its release version 1.20 and it is recommended to use containerd for Kubernetes installs >= 1.20${NC}\n"
```
laverya, line 41:
```suggestion
            printf "${YELLOW}Kubernetes has deprecated Docker as of 1.20, and it is recommended to use containerd instead.${NC}\n"
```

emosbaugh (changes requested):

emosbaugh (approved):

PR #3836, opened 2022-12-09
chore: temporarily removing longhorn logs

ricardomaraschini:

testgrid started to fail after this funcion call was added. it is still not clear why it is failing but i am removing it so i can have a fresh run and analyse if it makes a difference.


camilamacedo86 (approved):
Could you please add the link for the testgrid where you are checking the error?

ricardomaraschini:
> Could you please add the link for the testgrid where you are checking the error?

https://testgrid.kurl.sh/run/pr-3612-67a9b59-minio-2022-12-07T00-56-37Z-k8s-docker-2022-12-08T20:30:27Z

PR #3837, opened 2022-12-09
bug: take only "default" namespace into account (OpenEBS check)

ricardomaraschini:

Whenever checking the amount of free disk space for the OpenEBS provisioner the Kurl binary creates Pods and PVCs in the "default" namespace, it never creates or needs to access anything outside of that namespace. 

The current logic does not take that into account and analyses PVs that point to PVCs in any namespace, this can lead to an unexpected timeout when, by bad luck, the cluster has another PVC with the same random name in a different namespace.

Fixes #63415

### DIFF
--- a/pkg/cluster/space/openebs_free_disk_space_getter.go
+++ b/pkg/cluster/space/openebs_free_disk_space_getter.go
@@ -326,7 +326,7 @@ func (o *OpenEBSFreeDiskSpaceGetter) deleteTmpPVCs(pvcs []*corev1.PersistentVolu
 
 	pvsByPVCName := map[string]corev1.PersistentVolume{}
 	for _, pv := range pvs.Items {
-		if pv.Spec.ClaimRef == nil {
+		if pv.Spec.ClaimRef == nil || pv.Spec.ClaimRef.Namespace != "default" {
 			continue
 		}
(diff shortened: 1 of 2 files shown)


rrpolanco (approved):

rrpolanco:
@ricardomaraschini I'm curious, if the `default` ns doesn't exist, do we create it? Fail?

ricardomaraschini:
> @ricardomaraschini I'm curious, if the `default` ns doesn't exist, do we create it? Fail?

I have never thought about "default" namespace not existing (never seen it but I guess this is perfectly fine). Do you think we should use a different namespace (or even create a temporary one for that matter) ?

rrpolanco:
> > @ricardomaraschini I'm curious, if the `default` ns doesn't exist, do we create it? Fail?
> 
> @rrpolanco I have never thought about "default" namespace not existing (never seen it but I guess this is perfectly fine). Do you think we should use a different namespace (or even create a temporary one for that matter) ?

@ricardomaraschini For the PVMigrate validation [PR](https://github.com/replicatedhq/pvmigrate/pull/139) I made the same initial assumption and initially restricted the validation resources within the `default` ns but I was told to not presume the `default` will be there. Not a big deal IMO since creating a namespace is cheap. Just something to keep in mind.

PR #3839, opened 2022-12-09
fix: error to install containerd.io that conflicts with containerd because an containerd pkg installed within docker (only for new installs)

camilamacedo86:

fix: error to install containerd.io that conflicts with containerd because an containerd pkg installed within docker.io. (Only for new installs by checking if kubectl is or not present on the host) 

Example

``` 
⚙  Installing host packages containerd.io
Selecting previously unselected package containerd.io.
dpkg: considering removing containerd in favour of containerd.io ...
dpkg: no, cannot proceed with removal of containerd (--auto-deconfigure will help):
 docker.io depends on containerd (>= 1.2.6-0ubuntu1~)
  containerd is to be removed.

dpkg: regarding .../containerd.io_1.6.8-1_amd64.deb containing containerd.io:
 containerd.io conflicts with containerd
  containerd (version 1.5.9-0ubuntu1~20.04.4) is present and installed.

dpkg: error processing archive ./addons/containerd/1.6.8/ubuntu-20.04/containerd.io_1.6.8-1_amd64.deb (--install):
 conflicting packages - not installing containerd.io
Errors were encountered while processing:
 ./addons/containerd/1.6.8/ubuntu-20.04/containerd.io_1.6.8-1_amd64.deb
An error occurred on line 2017
```

Fixes # [sc-60835]

Now, if the docker be installed but the k8s not (new install) so it seems safe we remove the docker before install containerd.io.

Examples:

When we uninstall docker:

Screenshot 2022-12-09 at 12 44 06

## Steps to reproduce

```
$ sudo apt-get update
```

```release-note
fix: error to install containerd.io that conflicts with containerd because an containerd pkg is installed within docker
```

### DIFF
--- a/scripts/common/docker.sh
+++ b/scripts/common/docker.sh
@@ -207,3 +207,60 @@ function canonical_image_name() {
     fi
     echo "$image"
 }
+
+# It will only uninstall docker if is a new installation
+# and the installer has containerd set to workaround the bug issue:
+# `dpkg: no, cannot proceed with removal of containerd ... docker.io
+# depends on containerd (>= 1.2.6-0ubuntu1~)  containerd is to be removed.`
+# More info: https://bugs.launchpad.net/ubuntu/+source/docker.io/+bug/1940920
+# https://bugs.launchpad.net/ubuntu/+source/docker.io/+bug/1939140
+function uninstall_docker_new_installs_with_containerd() {
+
+     # If docker is not installed OR if conatinerd is not in the spec
+     # then, the docker should not be uninstalled
+     if ! commandExists docker || [ -z "$CONTAINERD_VERSION" ]; then
+          return
+     fi
+
+     # if k8s is installed already then, the docker should not be uninstalled
+     # so that it can be properly migrated to containerd
+     if commandExists kubectl ; then
+          return
+     fi
+
+     logWarn "Docker already exists on this machine and it will be uninstalled to avoid conflicts with containerd.\n"
+
+     if [ "$(docker ps -aq | wc -l)" != "0" ] ; then
+         docker ps -aq | xargs docker rm -f || true
+     fi
+     # The rm -rf /var/lib/docker command below may fail with device busy error, so remove as much
+     # data as possible now
+     docker system prune --all --volumes --force || true
+     systemctl disable docker.service --now || true
+
+     # Note that the docker.io can only be removed because it is prior install containerd and
+     # it is a new install. Otherwise, this dep is required.
+     # Important: The conflict is only removed when we uninstall docker.io
+     case "$LSB_DIST" in
+         ubuntu)
+             export DEBIAN_FRONTEND=noninteractive
+             dpkg --purge docker.io docker-ce docker-ce-cli
+             ;;
+
+         centos|rhel|amzn|ol)
+             local dockerPackages=("docker.io" "docker-ce" "docker-ce-cli")
+             if rpm -qa | grep -q 'docker-ce-rootless-extras'; then
+                 dockerPackages+=("docker-ce-rootless-extras")
+             fi
+             if rpm -qa | grep -q 'docker-scan-plugin'; then
+                 dockerPackages+=("docker-scan-plugin")
+             fi
+             rpm --erase ${dockerPackages[@]}
+             ;;
+     esac
+
+     rm -rf /var/lib/docker /var/lib/dockershim || true
+     rm -f /var/run/dockershim.sock || true
+     rm -f /var/run/docker.sock || true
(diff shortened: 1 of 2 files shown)


scripts/common/host-packages.sh
```diff
@@ -114,9 +114,19 @@ function _dpkg_install_host_packages() {
...
-    logSuccess "Host packages ${packages[*]} installed"
+    if DEBIAN_FRONTEND=noninteractive dpkg --install --force-depends-version --force-confold "${fullpath}"/*.deb; then
+        logSuccess "Host packages ${packages[*]} installed"
+    else
+        logFail "Unable to install the packages ${packages[*]}" >&2
+        if  commandExists docker && [[ "${packages[*]}" == *"containerd.io"* ]]; then
+            # If the error was faced because of conflicts it still possible to move forward because we will migrate docker to containerd and then, uninstall docker
+            # Usually, in this case containerd deps are installed already. Therefore, we just warning but we will not stop the install.
+            logWarn "Unable to install the containerd.io package probably because an compatible version of docker package is installed.\n
+            The script will continue and try to migrate the docker to containerd to perform this install. However, if it fails try to uninstall docker and re-run the script again.\n"
```
emosbaugh, line 125: Is this the best advice? If they uninstall docker in an airgapped environment how will they recover this images? I fear this will result in a support escalation.

scripts/common/docker.sh
```diff
@@ -207,3 +207,60 @@ function canonical_image_name() {
...
     echo "$image"
 }
+
+# It will only uninstall docker if is a new installation
+# and the installer has containerd set to workaround the bug issue:
+# `dpkg: no, cannot proceed with removal of containerd ... docker.io
+# depends on containerd (>= 1.2.6-0ubuntu1~)  containerd is to be removed.`
+# More info: https://bugs.launchpad.net/ubuntu/+source/docker.io/+bug/1940920
+# https://bugs.launchpad.net/ubuntu/+source/docker.io/+bug/1939140
+function uninstall_docker_new_installs_with_containerd() {
```
emosbaugh, line 217: this almost entirely duplicates the other function. is it possible to reuse code?

emosbaugh (changes requested):

scripts/common/docker.sh
camilamacedo86: It s not exactly the same, see that we cannot remove the same packages.

scripts/common/host-packages.sh
camilamacedo86, line 125:
```suggestion
            However, it will continue and try to migrate the docker installed to containerd to perform this install.\n"
```

scripts/common/host-packages.sh
```diff
@@ -114,9 +114,19 @@ function _dpkg_install_host_packages() {
...
-    logSuccess "Host packages ${packages[*]} installed"
+    if DEBIAN_FRONTEND=noninteractive dpkg --install --force-depends-version --force-confold "${fullpath}"/*.deb; then
+        logSuccess "Host packages ${packages[*]} installed"
+    else
+        logFail "Unable to install the packages ${packages[*]}" >&2
+        if  commandExists docker && [[ "${packages[*]}" == *"containerd.io"* ]]; then
+            # If the error was faced because of conflicts it still possible to move forward because we will migrate docker to containerd and then, uninstall docker
+            # Usually, in this case containerd deps are installed already. Therefore, we just warning but we will not stop the install.
+            logWarn "Unable to install the containerd.io package probably because an compatible version of docker package is installed.\n
+            However, it will continue and try to migrate the docker installed to containerd to perform this install.\n"
```
camilamacedo86, line 125: @emosbaugh I changed here to address your concern.
Also, I think to address the issue (common scenario) we can leave just the impl that remove docker only when we know that k8s is not installed. It is an extra think so, I think we could remove this one. WDYT?

emosbaugh: If we proceed here the script will never install the expected version of containerd, correct?

im not sure about this. i do like your idea (b) but i think it is required to install the expected containerd version after migration. if that is not possible there will be no upgrade path which is a non starter.

perhaps for now we can remove this code entirely and merge to get new installations working and follow up with another pr for existing installs.

emosbaugh (changes requested):

scripts/common/host-packages.sh
camilamacedo86: Hi @emosbaugh,

I removed the changes in the host file to have only the uninstall for new installs (when we can check that kubectl is not installed, docker.io is installed)

Let me know wdyt? Could we try get this one merged to check it with testgrid?

emosbaugh (approved):

PR #3844, opened 2022-12-12
feat: (for ubuntu) check status and try to fix broken dependency automatically

camilamacedo86:

kURL is distributing and installing the packages. Therefore, in some scenarios it might fail. 
Note that KURL uses dpkg to do the install of the packages ( [more info](https://github.com/replicatedhq/kURL/blob/main/scripts/common/host-packages.sh#L117) ) which cannot, for example sorted out unmet dependencies. From the [Package Management ubuntu docs](https://ubuntu.com/server/docs/package-management):

> “dpkg is a package manager for Debian-based systems. It can install, remove, and build packages, but unlike other package management systems, it cannot automatically download and install packages or their dependencies. Apt and Aptitude are newer, and layer additional features on top of dpkg. This section covers using dpkg to manage locally installed packages”

**Why broken packages occurs?** 

Bronken packages means that the package manager could not satisfy the criteria to properly install one or more packages. Note that each package has the info about what are the other versioned packages required to be installed within. Therefore the package manager has the purpose to ensure that the criteria for all packages installed in a server are properly respected. A lot of scenarios can result in this state. For further info: https://askubuntu.com/questions/140246/how-do-i-resolve-unmet-dependencies-after-adding-a-ppa#:~:text=One%20of%20the%20most%20common,Ubuntu%20repositories)%20or%20remove%20PPA.
 
*The goal of this PR is not sorted out all possible scenarios but instead:**
 
- a) begin to check if kURL broke or not the packages (today KURL does not do this check and finishes successfully)
- b) warning and call the recommended step to try fix the dependencies (which usually is suggested when this errors are faced) See, from https://linux.die.net/man/8/apt-get

> -f, --fix-broken
Fix. Attempt to correct a system with broken dependencies in place. This option, when used with install/remove, can omit any packages to permit APT to deduce a likely solution. Any package(s) that are specified must completely correct the problem. This option is sometimes necessary when running APT for the first time; APT itself does not allow broken package dependencies to exist on a system. It is possible that a system's dependency structure can be so corrupt as to require manual intervention. Use of this option together with -m may produce an error in some situations.
Configuration Item: APT::Get::Fix-Broken.
- c) use the --auto-deconfigure to allow remove the package when required 
- d) begin to fail if/when is not possible to sort it out to let users be aware of instead of move silent and finish successful . 

**Example Scenario that can be sorted out with:**

- Check the steps to reproduce this issue (where kURL is replacing a host dependency within a previous version causing unmet dependency)
- Then, see the error faced then run the commands to check the the apt-get status and follow the recommendation to try to fix it.

For further information check the references:
- [Difference Between APT and DPKG in Ubuntu](https://www.geeksforgeeks.org/difference-between-apt-and-dpkg-in-ubuntu/#:~:text=DPKG%20also%20supports%20local%20package%20installation.&text=APT%20doesn%27t%20terminate%20if,package%20with%20it%27s%20dependencies%20missing.)
- [What’s the Difference Between APT and dpkg in Ubuntu?](https://www.makeuseof.com/apt-vs-dpkg/)

Screenshot 2022-12-13 at 12 06 27

Related # [sc-62451] 
Fix # [sc-61869]
Fix # [sc-60835]

See:

Screenshot 2022-12-15 at 11 54 21

Also, see the test done with 60835:

Screenshot 2022-12-19 at 11 50 52

## Steps to reproduce

- Create a VM: Ubuntu 20.04
- Run: curl https://kurl.sh/version/v2022.11.10-1/3d6adc9 | sudo bash
- Then, check that the package manager is broken sudo apt-get check status

OR

- $ curl https://kurl-sh.s3.amazonaws.com/dist/v2022.11.10-1/kubernetes-1.25.3.tar.gz > kubernetes-1.25.3.tar.gz
- $ tar zxvf kubernetes-1.25.3.tar.gz 
- $ cd packages/kubernetes/1.25.3/ubuntu-20.04/
- $ sudo dpkg --install --force-depends-version --force-confold perl-modules-5.30_5.30.0-9ubuntu0.2_all.deb
- $ sudo apt-get check status

##### To test the changes made in this PR you can:

- Ensure that you have the changes done https://github.com/replicatedhq/kURL/pull/3870 to say no to remove docker
- Use any scenario where containerd should be installed such as k8s 1.25.3 and containerd 1.6.9
- Run `sudo apt-get update` and `sudo apt-get install docker.io`
- Then, run the  install  (sub bash ./install.sh) 

Also, you can run the latest without add any conflict and check that all will work fine.

```release-note
feat: (for ubuntu) after install required dependencies with dpkg, check apt-get status and try to fix broken dependency automatically to sorted out unmet dependencies with `sudo apt-get install -f -y`
```

NONE

### DIFF
--- a/scripts/common/host-packages.sh
+++ b/scripts/common/host-packages.sh
@@ -103,7 +103,7 @@ function _dpkg_apt_get_status_and_maybe_fix_broken_pkgs() {
     fi
 
     logWarn "Attempting to correct broken packages by running sudo apt-get install --fix-broken --yes"
-    sudo apt-get install --fix-broken --yes
+    apt-get install --fix-broken --yes
     if sudo apt-get check status ; then
         logSuccess "Broken packages fixed successfully"
         return
--- a/scripts/common/host-packages.sh
+++ b/scripts/common/host-packages.sh
@@ -102,8 +102,8 @@ function _dpkg_apt_get_status_and_maybe_fix_broken_pkgs() {
         return
     fi
 
-    logWarn "Attempting to correct broken packages by running apt-get install -f -y"
-    sudo apt-get install -f -y
+    logWarn "Attempting to correct broken packages by running sudo apt-get install --fix-broken --yes"
+    sudo apt-get install --fix-broken --yes
     if sudo apt-get check status ; then
         logSuccess "Broken packages fixed successfully"
         return
--- a/scripts/common/host-packages.sh
+++ b/scripts/common/host-packages.sh
@@ -108,7 +108,7 @@ function _dpkg_apt_get_status_and_maybe_fix_broken_pkgs() {
         logSuccess "Broken packages fixed successfully"
         return
     fi
-    bail "Unable to fix broken packages. It is required manual intervention. Run the command '$sudo apt-get check status' to get further information."
+    bail "Unable to fix broken packages. It is required manual intervention. Run the command '$ apt-get check status' to get further information."
 }
 
 function _dpkg_install_host_packages() {
--- a/scripts/common/host-packages.sh
+++ b/scripts/common/host-packages.sh
@@ -97,7 +97,7 @@ function dpkg_install_host_packages() {
 
 function _dpkg_apt_get_status_and_maybe_fix_broken_pkgs() {
     logStep "Checking package manager status"
-    if  sudo apt-get check status ; then
+    if apt-get check status ; then
         logSuccess "Status checked successfully. No broken packages were found."
         return
     fi
--- a/scripts/common/host-packages.sh
+++ b/scripts/common/host-packages.sh
@@ -95,6 +95,22 @@ function dpkg_install_host_packages() {
     _dpkg_install_host_packages "$dir" "$dir_prefix" "${packages[@]}"
 }
 
+function _dpkg_apt_get_status_and_maybe_fix_broken_pkgs() {
+    logStep "Checking package status"
+    if  sudo apt-get check status ; then
+        logSuccess "Status checked successfully. No broken packages were found."
+        return
+    fi
+
+    logWarn "Attempting to correct broken packages by running apt-get install -f -y"
+    sudo apt-get install -f -y
+    if sudo apt-get check status ; then
+        logSuccess "Broken packages fixed successfully"
+        return
+    fi
+    bail "Unable to fix broken packages. It is required manual intervention. Run the command '$sudo apt-get check status' to get further information."
+}
+
 function _dpkg_install_host_packages() {
     if [ "${SKIP_SYSTEM_PACKAGE_INSTALL}" == "1" ]; then
         logStep "Skipping installation of host packages: ${packages[*]}"
@@ -114,9 +130,11 @@ function _dpkg_install_host_packages() {
         return 0
     fi
 
-    DEBIAN_FRONTEND=noninteractive dpkg --install --force-depends-version --force-confold "${fullpath}"/*.deb
+    DEBIAN_FRONTEND=noninteractive dpkg --install --force-depends-version --force-confold --auto-deconfigure "${fullpath}"/*.deb
 
     logSuccess "Host packages ${packages[*]} installed"
+
+    _dpkg_apt_get_status_and_maybe_fix_broken_pkgs
 }
 
 function yum_install_host_archives() {
--- a/scripts/common/host-packages.sh
+++ b/scripts/common/host-packages.sh
@@ -96,7 +96,7 @@ function dpkg_install_host_packages() {
 }
 
 function _dpkg_apt_get_status_and_maybe_fix_broken_pkgs() {
-    logStep "Checking package status"
+    logStep "Checking package manager status"
     if  sudo apt-get check status ; then
         logSuccess "Status checked successfully. No broken packages were found."
         return
--- a/scripts/common/host-packages.sh
+++ b/scripts/common/host-packages.sh
@@ -104,7 +104,7 @@ function _dpkg_apt_get_status_and_maybe_fix_broken_pkgs() {
 
     logWarn "Attempting to correct broken packages by running sudo apt-get install --fix-broken --yes"
     apt-get install --fix-broken --yes
-    if sudo apt-get check status ; then
+    if apt-get check status ; then
         logSuccess "Broken packages fixed successfully"
         return
     fi


scripts/common/host-packages.sh
```diff
@@ -93,6 +93,26 @@ function dpkg_install_host_packages() {
...
+    _dpkg_apt_get_status_and_fix_broken_deps
+}
+
+function _dpkg_apt_get_status_and_fix_broken_deps() {
+    _out=$(sudo apt-get check status 2>/dev/null)
+    _status=$?
+    if [ "$_status" -eq "0" ]; then
+        logSuccess "Apt cache status checked successfully"
+    else
+        logStep "Attempting to correct broken dependencies in place"
```
camilamacedo86, line 105:
```suggestion
        logWarn "Attempting to correct broken dependencies in place"
```

scripts/common/host-packages.sh
```diff
@@ -93,6 +93,27 @@ function dpkg_install_host_packages() {
...
+        logWarn "Attempting to correct broken dependencies in place"
+        sudo apt-get install -f -y
+        _out=$(sudo apt-get check status 2>/dev/null)
+        _status=$?
+        if [ "$_status" -eq "0" ]; then
+            logSuccess "Broken dependencies fixed successfully"
+        else
+            logFail "Unable to fix broken dependencies"
+            logFail apt-get check status
+            return 1
```
camilamacedo86, line 114: This scenario is hard to check/test so I need your help because we are aware more of the code details, if by return 1 here the expected behaviour would occurs. That means fail and allow the user open a ticket within: https://github.com/replicatedhq/kURL/blob/main/scripts/install.sh#L532

scripts/common/host-packages.sh
```diff
@@ -93,6 +93,27 @@ function dpkg_install_host_packages() {
...
     local packages=("${@:2}")
     _dpkg_install_host_packages "$dir" "$dir_prefix" "${packages[@]}"
+    _dpkg_apt_get_status_and_fix_broken_deps
+}
+
+function _dpkg_apt_get_status_and_fix_broken_deps() {
+    _out=$(sudo apt-get check status 2>/dev/null)
+    _status=$?
+    if [ "$_status" -eq "0" ]; then
+        logSuccess "Apt cache status checked successfully"
```
ricardomaraschini, line 103: Can we return here and avoid the nested section under the `else` ?

scripts/common/host-packages.sh
```diff
@@ -93,6 +93,27 @@ function dpkg_install_host_packages() {
...
+    else
+        logWarn "Attempting to correct broken dependencies in place"
+        sudo apt-get install -f -y
+        _out=$(sudo apt-get check status 2>/dev/null)
+        _status=$?
+        if [ "$_status" -eq "0" ]; then
+            logSuccess "Broken dependencies fixed successfully"
```
ricardomaraschini, line 110: ditto

ricardomaraschini (commented):
Have you been able to uncover why exactly are we breaking the packaging system ? Are we installing a newer package than the one officially available ?

camilamacedo86:
Hi @ricardomaraschini,

> Have you been able to uncover why exactly are we breaking the packaging system ? Are we installing a newer package than the one officially available ?

Please see the description: https://github.com/replicatedhq/kURL/pull/3844#pullrequestreview-1215305341
Also, for further details and better understand the use case where we check this fix look for the internal draft doc shared. There, you can follow up all steps used to analyse the example scenario added here over how to reproduce the scenario.

scripts/common/host-packages.sh
```diff
@@ -95,6 +95,22 @@ function dpkg_install_host_packages() {
...
 
+function _dpkg_apt_get_status_and_maybe_fix_broken_pkgs() {
+    logStep "Checking package status"
+    if  sudo apt-get check status ; then
+        logSuccess "Status checked successfully. No broken packages were found."
+        return
+    fi
+
+    logWarn "Attempting to correct broken packages by running apt-get install -f -y"
+    sudo apt-get install -f -y
```
banjoh, line 106: _A bit of a nit pick here_

Perhaps writing the full command as `sudo apt-get install --fix-broken --yes` here would make things explicit for future readers

scripts/common/host-packages.sh
```diff
@@ -114,9 +130,11 @@ function _dpkg_install_host_packages() {
         return 0
     fi
 
-    DEBIAN_FRONTEND=noninteractive dpkg --install --force-depends-version --force-confold "${fullpath}"/*.deb
+    DEBIAN_FRONTEND=noninteractive dpkg --install --force-depends-version --force-confold --auto-deconfigure "${fullpath}"/*.deb
```
banjoh, line 133: For future consideration, should we start using `apt-get` instead to leverage its dependency management?

banjoh (approved):

scripts/common/host-packages.sh
banjoh: `-f` at times can be misunderstood as `--force`

banjoh: There is also `--fix-missing` option in the mix that makes things even more confusing :)

scripts/common/host-packages.sh
camilamacedo86: We cannot use apt-get because we are providing the packages, Therefore, it might fails when should not because tried to looking for a host packaged that is required and will be installed as part of the kURL install. 

Also, we need to keep in mind that the apt-get will only be able to fix the dependencies and get the new requirements when those are found available in the package manager index, which is updated from the repositories configured in the specific machine. Therefore, the user host might have configured a different repository from where we got the package, and it might fail in a use case that would still be able to work out. 

It seems that calling dpkg and the apt-get install -f is kind of the same as calling apt-get install ./deb directly. (Look [here](https://unix.stackexchange.com/questions/159094/how-to-install-a-deb-file-by-dpkg-i-or-by-apt) might help to better understand the thought)

scripts/common/host-packages.sh
camilamacedo86: Hun, if we run --help I think we can check that is fix-broken and yes. 
I think we can move forward with right. That does not seems a blocker for we get this one merged.

scripts/common/host-packages.sh
```diff
@@ -95,6 +95,22 @@ function dpkg_install_host_packages() {
...
 }
 
+function _dpkg_apt_get_status_and_maybe_fix_broken_pkgs() {
+    logStep "Checking package status"
+    if  sudo apt-get check status ; then
+        logSuccess "Status checked successfully. No broken packages were found."
+        return
+    fi
+
+    logWarn "Attempting to correct broken packages by running apt-get install -f -y"
```
camilamacedo86, line 105:
```suggestion
    logWarn "Attempting to correct broken packages by running sudo apt-get install --fix-broken --yes"
```

scripts/common/host-packages.sh
camilamacedo86, line 106:
```suggestion
    sudo apt-get install --fix-broken --yes
```

camilamacedo86: Done.

scripts/common/host-packages.sh
banjoh: What I meant was that `apt-get install ./deb` will go ahead and resolve dependencies in addition to installing the local debian package. This way, you would not need to run `sudo apt-get check status`. Either approach should be fine by me.

scripts/common/host-packages.sh
banjoh: For sure you can do that. It would be helpful if I'm reading code I don't have to search for apt-get man pages or google around to see what this options mean. Just making a readers life easier. Not a blocker anyway.

scripts/common/host-packages.sh
```diff
@@ -95,6 +95,22 @@ function dpkg_install_host_packages() {
     _dpkg_install_host_packages "$dir" "$dir_prefix" "${packages[@]}"
 }
 
+function _dpkg_apt_get_status_and_maybe_fix_broken_pkgs() {
+    logStep "Checking package status"
```
ricardomaraschini, line 99: Maybe use "package manager" status instead ?

scripts/common/host-packages.sh
```diff
@@ -95,6 +95,22 @@ function dpkg_install_host_packages() {
...
+    if  sudo apt-get check status ; then
```
ricardomaraschini, line 100: Do we need to use `sudo` here ? I think we are already root.

scripts/common/host-packages.sh
```diff
@@ -95,6 +95,22 @@ function dpkg_install_host_packages() {
...
+        logSuccess "Status checked successfully. No broken packages were found."
+        return
+    fi
+
+    logWarn "Attempting to correct broken packages by running sudo apt-get install --fix-broken --yes"
+    sudo apt-get install --fix-broken --yes
```
emosbaugh, line 106: Remove sudo

scripts/common/host-packages.sh
```diff
@@ -95,6 +95,22 @@ function dpkg_install_host_packages() {
...
+    if sudo apt-get check status ; then
```
emosbaugh, line 107: Remove sudo

scripts/common/host-packages.sh
camilamacedo86: However, yes described above we cannot use `apt-get install ./deb` we must use dpkg to do the installation.

scripts/common/host-packages.sh
camilamacedo86: Yes, we need the sudo otherwise we will check Permission denied

scripts/common/host-packages.sh
camilamacedo86, line 99:
```suggestion
    logStep "Checking package manager status"
```

scripts/common/host-packages.sh
camilamacedo86: Screenshot 2022-12-19 at 14 39 59

scripts/common/host-packages.sh
emosbaugh: This has the potential to not work on an airgapped instance. Let's wait for the host packages proposal before addressing this.

```
$ sudo apt-get install --fix-broken --yes
Reading package lists... Done
Building dependency tree
Reading state information... Done
Correcting dependencies... Done
The following packages were automatically installed and are no longer required:
  libatasmart4 libblockdev-fs2 libblockdev-loop2 libblockdev-part-err2 libblockdev-part2
  libblockdev-swap2 libblockdev-utils2 libblockdev2 libmbim-glib4 libmbim-proxy libmm-glib0
  libnspr4 libnss3 libnuma1 libparted-fs-resize0 libqmi-glib5 libqmi-proxy libudisks2-0 libxmlb2
  usb-modeswitch usb-modeswitch-data
Use 'sudo apt autoremove' to remove them.
The following additional packages will be installed:
  libperl5.30 perl perl-modules-5.30
Suggested packages:
  perl-doc libterm-readline-gnu-perl | libterm-readline-perl-perl make libb-debug-perl
  liblocale-codes-perl
The following packages will be upgraded:
  libperl5.30 perl perl-modules-5.30
3 upgraded, 0 newly installed, 0 to remove and 4 not upgraded.
Need to get 6914 kB of archives.
After this operation, 12.3 kB of additional disk space will be used.
0% [Connecting to us-central1.gce.archive.ubuntu.com (35.184.213.5)]
```

emosbaugh (changes requested):

scripts/common/host-packages.sh
camilamacedo86: HI @emosbaugh,

Today it will fail ALWAYS when a conflict be faced or worst finish the install and does not let the user know.
The changes here allows the package manager remove a host package if/when required and try to sorted out when possible. It does not matter if is airgap or not. If the package manager has the dependency it will be sorted out. Otherwise, it would fail at the same away.

emosbaugh: you will need to remove sudo. the script will always be invoked as root or with sudo.

scripts/common/host-packages.sh
emosbaugh: you will need to remove sudo. the script will always be invoked as root or with sudo.

scripts/common/host-packages.sh
camilamacedo86: That is true |o|

scripts/common/host-packages.sh
```diff
@@ -95,6 +95,22 @@ function dpkg_install_host_packages() {
...
 
+function _dpkg_apt_get_status_and_maybe_fix_broken_pkgs() {
+    logStep "Checking package manager status"
+    if  sudo apt-get check status ; then
+        logSuccess "Status checked successfully. No broken packages were found."
+        return
+    fi
+
+    logWarn "Attempting to correct broken packages by running sudo apt-get install --fix-broken --yes"
+    sudo apt-get install --fix-broken --yes
```
camilamacedo86, line 106:
```suggestion
    apt-get install --fix-broken --yes
```

scripts/common/host-packages.sh
```diff
@@ -95,6 +95,22 @@ function dpkg_install_host_packages() {
...
+    if sudo apt-get check status ; then
```
camilamacedo86, line 107:
```suggestion
    if apt-get check status ; then
```

scripts/common/host-packages.sh
camilamacedo86, line 106:
```suggestion
    apt-get install --fix-broken --yes
```

scripts/common/host-packages.sh
```diff
@@ -95,6 +95,22 @@ function dpkg_install_host_packages() {
     _dpkg_install_host_packages "$dir" "$dir_prefix" "${packages[@]}"
 }
 
+function _dpkg_apt_get_status_and_maybe_fix_broken_pkgs() {
+    logStep "Checking package manager status"
+    if  sudo apt-get check status ; then
```
camilamacedo86, line 100:
```suggestion
    if apt-get check status ; then
```

scripts/common/host-packages.sh
```diff
@@ -95,6 +95,22 @@ function dpkg_install_host_packages() {
...
+        return
+    fi
+
+    logWarn "Attempting to correct broken packages by running sudo apt-get install --fix-broken --yes"
+    apt-get install --fix-broken --yes
+    if apt-get check status ; then
+        logSuccess "Broken packages fixed successfully"
+        return
+    fi
+    bail "Unable to fix broken packages. It is required manual intervention. Run the command '$sudo apt-get check status' to get further information."
```
camilamacedo86, line 111:
```suggestion
    bail "Unable to fix broken packages. It is required manual intervention. Run the command '$ apt-get check status' to get further information."
```

emosbaugh (approved):

scripts/common/host-packages.sh
```diff
@@ -95,6 +95,22 @@ function dpkg_install_host_packages() {
...
 
+function _dpkg_apt_get_status_and_maybe_fix_broken_pkgs() {
+    logStep "Checking package manager status"
+    if apt-get check status ; then
+        logSuccess "Status checked successfully. No broken packages were found."
+        return
+    fi
+
+    logWarn "Attempting to correct broken packages by running sudo apt-get install --fix-broken --yes"
+    apt-get install --fix-broken --yes
```
rrpolanco, line 106: Will this work with airgap installs?

camilamacedo86: `apt-get install --fix-broken --yes` ask for the package manager try to fix by found, installing and etc the required dependencies. It will try to get from the local machine and from what is currently available in the index ( provided by the repos configured to be used by it)

It works exactly the same way for online or offline.
The only problem that we might to be faced here within that I can think about in the airgap bundle is:

- a) The apt-get index was updated when the machine was connected to the internet
- b) Then, you disconnect from the internet 
- c) The apt-get has a lot of versions and packages configured in the index that would only be retrieved by the internet

So, when we run this command that could be fail.

However, note that: 
- a) That would mean the offiline env is not configured correctly. That is not how works with offline envs. Offiline and online envs works very similar. Usually the only diff is that for an offiline env the package manager (host.list) will be configured to get the packages from a centralized hosting instead of the internet.
- b) We ONLY try to fix when we broke the package manager. That means, kURL install can only fail because it could not properly ensure that all host packages in the machine are properly installed as should be instead of finishes with success and does not let the users know about

I hope that answered all your questions and concerns. If not please, feel free to reach out. 

c/c @emosbaugh

PR #3850, opened 2022-12-13
Fix rook upgrade tasks

emosbaugh:

Fixes load images and task. Calls addon_load to actually load images. Shift task name from the args before passing them to the called function.

### DIFF
--- a/scripts/tasks.sh
+++ b/scripts/tasks.sh
@@ -74,10 +74,12 @@ function tasks() {
         rook-10-to-14-images|rook_10_to_14_images)
             rook_upgrade_tasks_load_images "from-version=1.0" "to-version=1.4"
             ;;
-        rook-upgrade)
+        rook-upgrade|rook_upgrade)
+            shift # the first param is rook-upgrade|rook_upgrade
             rook_upgrade_tasks_rook_upgrade "$@"
             ;;
-        rook-upgrade-load-images)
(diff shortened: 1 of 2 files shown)


scripts/tasks.sh
```diff
@@ -74,10 +74,12 @@ function tasks() {
         rook-10-to-14-images|rook_10_to_14_images)
             rook_upgrade_tasks_load_images "from-version=1.0" "to-version=1.4"
             ;;
-        rook-upgrade)
+        rook-upgrade|rook_upgrade)
```
camilamacedo86, line 77: is that right?

scripts/tasks.sh
```diff
@@ -74,10 +74,12 @@ function tasks() {
...
+            shift # the first param is rook-upgrade|rook_upgrade
             rook_upgrade_tasks_rook_upgrade "$@"
             ;;
-        rook-upgrade-load-images)
+        rook-upgrade-load-images|rook_upgrade_load_images)
```
camilamacedo86, line 81: Also, here?

camilamacedo86 (approved):

scripts/tasks.sh
camilamacedo86, line 77: that ringed the bell but after look the implementation I saw that we are doing in this away for all cases, so OK,

PR #3853, opened 2022-12-13
Remove lint and vet from kurl_util

emosbaugh:

We run lint and vet in the root Makefile. No need to duplicate here.

https://github.com/replicatedhq/kURL/blob/b7405229745676f5f924ed0c775bb6fba15671f0/Makefile#L633

### DIFF
--- a/.github/workflows/build-test.yaml
+++ b/.github/workflows/build-test.yaml
@@ -36,7 +36,7 @@ jobs:
 
       - uses: actions/checkout@v3
 
-      - run: make -C kurl_util deps test build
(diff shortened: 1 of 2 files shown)


ricardomaraschini (commented):
Seems like this has broken the CI:

```
make: *** No rule to make target 'deps'.  Stop.
```

rrpolanco:
Need to update: https://github.com/replicatedhq/kURL/blob/main/.github/workflows/build-test.yaml#L39

rrpolanco (approved):

PR #3854, opened 2022-12-13
Use kustomize bases rather than resources for k8s <= 1.20

emosbaugh:

Kubernetes <= 1.20 uses an older kustomize version that does not support bases in resources.

Error:
```
error: rawResources failed to read Resources: Load from path cephfs failed: 'cephfs' must be a file (got d='/var/lib/kurl/kustomize/rook/cluster/cephfs')
```


emosbaugh:
Testgrid is failing post-upgrade script

https://testgrid.kurl.sh/run/pr-3854-32cc416-rook-1.10.6-k8s-docker-2022-12-13T19:06:56Z

laverya (approved):

PR #3858, opened 2022-12-15
fix(centos|rhel|ol): allow allowerasing previous packages installed when conflicts are faced with yum install

camilamacedo86:

kURL distributed the packages, create a repo and will install them. However, if a conflict be faced because of a previous packaged installed it fails. Note that the `--allowerasing` has been used already. It has been used with the conditions if is `containerd` and rhel-8. However, conflicts can be faced in any rhel version as when the script will still any dependency.

Fixes # [sc-50237]
Also similar case: https://github.com/replicated-collab/cylance-kots/issues/16

## Steps to reproduce

- RHEL-4
- Run `curl https://kurl.sh/37de0ce | sudo bash`

```release-note
fix(centos|rhel|ol): allow allowerasing previous packages installed when conflicts are faced by kURL when it is installing the packages
```

NONE

### DIFF
--- a/scripts/common/host-packages.sh
+++ b/scripts/common/host-packages.sh
@@ -182,19 +182,13 @@ EOF
         previous_version="$(ctr -v | grep -o '[0-9][0-9]*\.[0-9][0-9]*\.[0-9][0-9]*')"
         logStep "Downgrading containerd, $previous_version -> $next_version"
         if semverCompare "$next_version" "$previous_version" && [ "$SEMVER_COMPARE_RESULT" -lt "0" ]; then
-            if uname -r | grep -q "el8" ; then
-                yum --disablerepo=* --enablerepo=kurl.local downgrade --allowerasing -y "${packages[@]}"
-            else
-                yum --disablerepo=* --enablerepo=kurl.local downgrade -y "${packages[@]}"
-            fi
+            yum --disablerepo=* --enablerepo=kurl.local downgrade --allowerasing -y "${packages[@]}"
         fi
         logSuccess "Downgraded containerd"
     fi
     # shellcheck disable=SC2086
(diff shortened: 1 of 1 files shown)


laverya (approved):
I would recommend making sure that this is covered in testgrid after merge (either a manual test or as part of the standard suites)

camilamacedo86:
HI @laverya, 

> I would recommend making sure that this is covered in testgrid after merge (either a manual test or as part of the standard suites)

All tests that are done against RHEL, Centos will be testing this change already.

**Therefore, are you suggesting ensuring that the --allowerasing works as expected? Am I right?**

Following some considerable thoughts: 

- 1 - How should must ensure that the allowerasing option works as it should? Kurl or YUM project that owns and provide the option? So, do not you agree that is part of yum domain and not of the kURL?

- 2 - Also, how would we do this test? Would we install a package that causes the conflict before running the KURL and then, check if the installation finishes succefully.  

However, the conflict may only exist for this specific version (latest) of kURL, and when we rebuild the packages to address a CEV for example, it no longer occurs because we will have new versions, etc. In this way, it would be hard to ensure that we could do an idempotent solution for this specific one. 

- 3-  It just allows us to solve automatically when possible and it is not a bulletprove, which will help reduce the support cases for now. 

By using this option we are say to the package manager: you are allowed to remove the package if you need to sort out the conflict. However, it does not means that the package manager will always and for any scenario be able to sort the conflict. 

However, please let me know if you disagree with the above thoughts.

camilamacedo86:
It was replaced by : https://github.com/replicatedhq/kURL/pull/3866 (just add the flag)
Just after check in the testgrid with all scenarios we can move forward to clean up the conditions.

PR #3870, opened 2022-12-19
feat: ask permission to allow remove docker in new installs to avoid …

camilamacedo86:

Improve the behaviour added in https://github.com/replicatedhq/kURL/pull/3839 by asking permission to the user before uninstall the docker in the preflights. 

Fixes # [sc-60835]

Screenshot 2022-12-19 at 11 25 46

## Steps to reproduce

$ sudo apt-get update

```release-note
feat: ask permission to allow remove docker in new installs to avoid conflicts with containerd
```

NONE

### DIFF
--- a/scripts/common/docker.sh
+++ b/scripts/common/docker.sh
@@ -226,11 +226,11 @@ function uninstall_docker_new_installs_with_containerd() {
 
      # if k8s is installed already then, the docker should not be uninstalled
      # so that it can be properly migrated to containerd
-     if commandExists kubectl ; then
+     if kubernetes_resource_exists kube-system configmap kurl-config; then
           return
      fi
 
-     logWarn "Docker already exists on this machine and it will be uninstalled to avoid conflicts with containerd.\n"
+     logStep "Uninstalling Docker to avoid conflicts with containerd package.\n"
 
      if [ "$(docker ps -aq | wc -l)" != "0" ] ; then
          docker ps -aq | xargs docker rm -f || true
--- a/scripts/common/preflights.sh
+++ b/scripts/common/preflights.sh
@@ -621,7 +621,9 @@ allow_remove_docker_new_install() {
           return
      fi
 
-     printf "\n${YELLOW}Docker already exists on this machine and Kubernetes is not installed. In order to avoid conflicts when installing containerd would be suggested to remove Docker. Remove Docker?${NC} "
+     printf "\n${YELLOW}Docker already exists on this machine and Kubernetes is not yet installed.${NC} "
+     printf "\n${YELLOW}In order to avoid conflicts when installing containerd, it is recommended that Docker be removed.
+     printf "\n${YELLOW}Remove Docker?${NC} "
      if confirmY ; then
           uninstall_docker_new_installs_with_containerd
      else
--- a/scripts/common/preflights.sh
+++ b/scripts/common/preflights.sh
@@ -621,7 +621,7 @@ allow_remove_docker_new_install() {
           return
      fi
 
-     printf "\n${YELLOW}Docker already exists on this machine and Kubernetes is not installed. In order to avoid conflicts when containerd will be installed do you allow Docker be uninstalled first?${NC} "
+     printf "\n${YELLOW}Docker already exists on this machine and Kubernetes is not installed. In order to avoid conflicts when installing containerd would be suggested to remove Docker. Remove Docker?${NC} "
      if confirmY ; then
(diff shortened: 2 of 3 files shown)


scripts/common/preflights.sh
```diff
@@ -605,3 +605,26 @@ function host_preflights_mkresults() {
...
+     # then, the docker should not be uninstalled
+     if ! commandExists docker || [ -z "$CONTAINERD_VERSION" ]; then
+          return
+     fi
+
+     # if k8s is installed already then, the docker should not be uninstalled
+     # so that it can be properly migrated to containerd
+     if commandExists kubectl ; then
+          return
+     fi
```
rrpolanco, line 622: I don't think checking the existence of `kubectl` command is enough to determine if this is a first install. We have this logic in `reporting.sh` to determine if there was a previous kURL installation: https://github.com/replicatedhq/kURL/blob/main/scripts/common/reporting.sh#L29

scripts/common/docker.sh
```diff
@@ -230,7 +230,7 @@ function uninstall_docker_new_installs_with_containerd() {
           return
      fi
 
-     logWarn "Docker already exists on this machine and it will be uninstalled to avoid conflicts with containerd.\n"
+     logStep "Uninstalling Docker to avoid conflicts when containerd will be installed.\n"
```
rrpolanco, line 233:
```suggestion
     logStep "Uninstalling Docker to avoid conflicts with containerd package.\n"
```

scripts/common/preflights.sh
```diff
@@ -605,3 +605,26 @@ function host_preflights_mkresults() {
...
+
+     printf "\n${YELLOW}Docker already exists on this machine and Kubernetes is not installed. In order to avoid conflicts when containerd will be installed do you allow Docker be uninstalled first?${NC} "
```
rrpolanco, line 624:
```suggestion
     printf "\n${YELLOW}Docker already exists on this machine and Kubernetes is not installed. In order to avoid conflicts when installing containerd Docker needs to be removed. Continue?${NC} "
```

scripts/common/preflights.sh
```diff
@@ -605,3 +605,26 @@ function host_preflights_mkresults() {
...
+     if confirmY ; then
+          uninstall_docker_new_installs_with_containerd
+     else
+          logWarn "\nThe installation will continue. However, if you face conflicts issue scenarios please uninstall Docker before re-run the install script"
```
rrpolanco, line 628:
```suggestion
          logWarn "\nThe installation will continue. However, if this script fails due to package conflicts, uninstall Docker and re-run the install script"
```

scripts/common/preflights.sh
camilamacedo86: After we merge https://github.com/replicatedhq/kURL/pull/3844 we might be able to fix in the biggest part of the scenarios.  So, because of this I think a more optional/recommended wording fits better here. Is that make sense?

Also, `Remove Docker?` is not gramarly wrong?

scripts/common/preflights.sh
camilamacedo86: Working on that :-)

camilamacedo86: Done and tested worked fine.

scripts/common/preflights.sh
```diff
@@ -605,3 +605,26 @@ function host_preflights_mkresults() {
...
+     # so that it can be properly migrated to containerd
+     if commandExists kubectl ; then
+          return
+     fi
+
+     printf "\n${YELLOW}Docker already exists on this machine and Kubernetes is not installed. In order to avoid conflicts when installing containerd would be suggested to remove Docker. Remove Docker?${NC} "
+     if confirmY ; then
+          uninstall_docker_new_installs_with_containerd
+     else
+          logWarn "\nThe installation will continue. However, if this script fails due to package conflicts, uninstall Docker and re-run the install script"
```
emosbaugh, line 628:
```suggestion
          logWarn "\nThe installation will continue, however, if this script fails due to package"
          logWarn "conflicts, please uninstall Docker and re-run the install script."
```

emosbaugh (changes requested):

camilamacedo86:
Hi @emosbaugh,

all suggestions were accepted. Therefore, what do you think? Could we get this one merged now?

rrpolanco (approved):

emosbaugh (approved):

PR #3871, opened 2022-12-19
chore: using rebuilt s3cmd image

ricardomaraschini:
#### What this PR does / why we need it:

This PR makes the Registry and the Velero add-ons default to a new `s3cmd` image. The new `s3cmd` image uses, as its base, the Alpine version 3.16.3 instead of 3.16.2 as the old image.

#### Which issue(s) this PR fixes:
Fixes #63702

#### Special notes for your reviewer:

I have manually dispatched the `build-image` GitHub action to rebuild the s3cmd image and verified that I can pull it and that it does not report the same issues as the old image:

```
$ docker pull kurlsh/s3cmd:20221219-ff60227
20221219-ff60227: Pulling from kurlsh/s3cmd
ca7dd9ec2225: Pull complete
eab94e9f41af: Pull complete
f60cb74099f7: Pull complete
Digest: sha256:08384f85979e48a52a6f8291c7b9fb1cf7def15afaeaf525d99495976f147a27
Status: Downloaded newer image for kurlsh/s3cmd:20221219-ff60227
docker.io/kurlsh/s3cmd:20221219-ff60227
$ trivy image kurlsh/s3cmd:20221219-ff60227
2022-12-19T16:42:14.035+0100    INFO    Vulnerability scanning is enabled
2022-12-19T16:42:14.036+0100    INFO    Secret scanning is enabled
2022-12-19T16:42:14.036+0100    INFO    If your scanning is slow, please try '--security-checks vuln' to disable secret scanning
2022-12-19T16:42:14.036+0100    INFO    Please see also https://aquasecurity.github.io/trivy/v0.35/docs/secret/scanning/#recommendation for faster secret detection
2022-12-19T16:42:14.039+0100    INFO    Detected OS: alpine
2022-12-19T16:42:14.039+0100    INFO    Detecting Alpine vulnerabilities...
2022-12-19T16:42:14.041+0100    INFO    Number of language-specific files: 0

kurlsh/s3cmd:20221219-ff60227 (alpine 3.16.3)

Total: 0 (UNKNOWN: 0, LOW: 0, MEDIUM: 0, HIGH: 0, CRITICAL: 0)

$
```

PR #3889, opened 2022-12-22
chore: address CVE for registry 2.71 and Velero 1.6.0, 1.6.1, 1.7.1, 1.8.1

camilamacedo86:

In the testgrid K8s 1.22x Airgap we use Velero 1.7.1 and it is failing with unmet dependecies because this version has a CVE. 
It does the same changes https://github.com/replicatedhq/kURL/pull/3871 but for those versions as well. 

Fixes # [sc-64917]

Not adding since it was not added to https://github.com/replicatedhq/kURL/pull/3871 as well.


camilamacedo86:
Closing in favor of https://github.com/replicatedhq/kURL/pull/3893 for we just fix the test since after a discussion we do not want to address the CVE for the old versions.

PR #3891, opened 2022-12-22
feat: add logs for kurl scripts

camilamacedo86:

Have the logs from what was executed with kurl is very helpful in the support.
After we get this PR merged we will be able to check all logs under /var/log/kurl

Examples:

Screenshot 2022-12-23 at 07 53 40

Fixes # [sc-35094]

**See that the output in the console still:** 

Screenshot 2022-12-23 at 07 51 42

**Also, if fails the error still showing in the output and log files:**

Screenshot 2022-12-23 at 07 54 46

**We are also hidden any line which might has a password:**

Following the test done by changing the script: 

- a) Change the script with the same done in: https://github.com/replicatedhq/kURL/pull/3891/files

Screenshot 2022-12-28 at 12 52 28

- b) Add in the main func the same print that will be done by kots example:

Screenshot 2022-12-28 at 12 19 38

- c) Run the sudo bash ./install
- d) Check that the line `"Login with password (will not be shown again): senha"`  cannot be found in the logs:

Screenshot 2022-12-28 at 12 58 46

**Also, to ensure that the script would fail after these changes:**

- a) Now, let's run without sudo
- b) See that it finished with the error code:

Screenshot 2022-12-28 at 12 55 23

**By last see that we need to change the funcs to read the data, so that we can have an interactive shell and the logs:**

Screenshot 2022-12-29 at 12 36 52

```release-note
feat: kURL execution logs can be found now under /var/log/kurl/
```

NONE

### DIFF
--- a/scripts/common/prompts.sh
+++ b/scripts/common/prompts.sh
@@ -5,7 +5,7 @@
 function prompts_can_prompt() {
     # Need the TTY to accept input and stdout to display
     # Prompts when running the script through the terminal but not as a subshell
-    if [ -t 1 ] && [ -c /dev/tty ]; then
+    if [ -c /dev/tty ]; then
         return 0
     fi
     return 1
@@ -266,7 +266,7 @@ function prompt_airgap_preload_images() {
     fi
 
     local unattended_nodes_missing_images=0
- 
+
     while read -r node; do
         local nodeName=$(echo "$node" | awk '{ print $1 }')
         if [ "$nodeName" = "$(get_local_node_name)" ]; then
--- a/scripts/install.sh
+++ b/scripts/install.sh
@@ -501,6 +501,7 @@ function report_kubernetes_install() {
 K8S_DISTRO=kubeadm
 
 function main() {
+    logStep "Running install with the argument(s): $@"
     require_root_user
     # ensure /usr/local/bin/kubectl-plugin is in the path
     path_add "/usr/local/bin"
@@ -569,4 +570,11 @@ function main() {
     report_install_success
 }
 
-main "$@"
+# tee logs into /var/log/kurl/install-<date>.log and stdout
+mkdir -p /var/log/kurl
+LOGFILE="/var/log/kurl/install-$(date +"%Y-%m-%dT%H-%M-%S").log"
+main "$@" 2>&1 | tee $LOGFILE
+# it is required to return the exit status of the script
+FINAL_RESULT="${PIPESTATUS[0]}"
+sed -i "/\b\(password\)\b/d" $LOGFILE > /dev/null 2>&1
+exit "$FINAL_RESULT"
--- a/scripts/join.sh
+++ b/scripts/join.sh
@@ -125,6 +125,7 @@ outro() {
 K8S_DISTRO=kubeadm
 
 function main() {
+    logStep "Running join with the argument(s): $@"
     export KUBECONFIG=/etc/kubernetes/admin.conf
     require_root_user
     # ensure /usr/local/bin/kubectl-plugin is in the path
@@ -171,4 +172,11 @@ function main() {
     popd_install_directory
 }
 
-main "$@"
+# tee logs into /var/log/kurl/install-<date>.log and stdout
+mkdir -p /var/log/kurl
+LOGFILE="/var/log/kurl/join-$(date +"%Y-%m-%dT%H-%M-%S").log"
+main "$@" 2>&1 | tee $LOGFILE
+# it is required to return the exit status of the script
+FINAL_RESULT="${PIPESTATUS[0]}"
+sed -i "/\b\(password\)\b/d" $LOGFILE > /dev/null 2>&1
+exit "$FINAL_RESULT"
--- a/scripts/tasks.sh
+++ b/scripts/tasks.sh
@@ -24,6 +24,7 @@ DIR=.
 
 K8S_DISTRO=
 function tasks() {
+    logStep "Running tasks with the argument(s): $@"
     # ensure /usr/local/bin/kubectl-plugin is in the path
     path_add "/usr/local/bin"
 
@@ -717,4 +718,10 @@ function install_host_dependencies_longhorn() {
     longhorn_host_init_common "${DIR}/packages/host/longhorn"
 }
 
-tasks "$@"
+mkdir -p /var/log/kurl
+LOGFILE="/var/log/kurl/tasks-$(date +"%Y-%m-%dT%H-%M-%S").log"
+tasks "$@" 2>&1 | tee $LOGFILE
+# it is required to return the exit status of the script
+FINAL_RESULT="${PIPESTATUS[0]}"
+sed -i "/\b\(password\)\b/d" $LOGFILE > /dev/null 2>&1
+exit "$FINAL_RESULT"
--- a/scripts/upgrade.sh
+++ b/scripts/upgrade.sh
@@ -1,4 +1,3 @@
-
 #!/bin/bash
 
 set -e
@@ -107,6 +106,7 @@ function outro() {
 K8S_DISTRO=kubeadm
 
 function main() {
+    logStep "Running upgrade with the argument(s): $@"
     export KUBECONFIG=/etc/kubernetes/admin.conf
     require_root_user
     # ensure /usr/local/bin/kubectl-plugin is in the path
@@ -148,4 +148,11 @@ function main() {
     popd_install_directory
 }
 
-main "$@"
+# tee logs into /var/log/kurl/upgrade-<date>.log and stdout
+mkdir -p /var/log/kurl
+LOGFILE="/var/log/kurl/upgrade-$(date +"%Y-%m-%dT%H-%M-%S").log"
+main "$@" 2>&1 | tee $LOGFILE
+# it is required to return the exit status of the script
+FINAL_RESULT="${PIPESTATUS[0]}"
+sed -i "/\b\(password\)\b/d" $LOGFILE > /dev/null 2>&1
+exit "$FINAL_RESULT"


scripts/install.sh
```diff
@@ -569,4 +569,7 @@ function main() {
     report_install_success
 }
 
-main "$@"
+# tee logs into /var/log/kurl/install-<date>.log and stdout
+mkdir -p /var/log/kurl
+main "$@" 2>&1 | tee -a /var/log/kurl/install-$(date +"%m-%d-%Y-%H-%M-%S").log
+
```
ricardomaraschini, line 575: I liked how simple this looks! Three things caught my attention though:

1. Maybe we should use a directory in `/var/lib/kurl` to store the log (not strong opinion here);
2. We loose access to `stderr`, now all is `stdout`. Not sure if this impacts anything;
3. We loose the return code of the `main` function (now it is always zero), aren't all testgrid become green because of that ?

As I don't know if these are real problems I prefer to hear what other have to say.

camilamacedo86: Hi @ricardomaraschini,

Thank you for the help and review. Following the answers.

> 1. Maybe we should use a directory in /var/lib/kurl to store the log (not strong opinion here);

/var/lib -> has only libs for the system
/var/log -> has the logs for the system. This dir has logs for other application and etc. The approach to create the dir with the name of the app from where the logs came from can also be checked there. 

> We loose access to stderr, now all is stdout. Not sure if this impacts anything;

Not sure, I could check all been outputting as usual and the same content in the logs
 
> We loose the return code of the main function (now it is always zero), aren't all testgrid become green because of that ?

I do not think so, because in the testgrid we have post scripts to check the results right. 
But that is a great point for we have attention. We can merge and check it out.

camilamacedo86: Hi @ricardomaraschini,

> 3. We loose the return code of the main function (now it is always zero), aren't all testgrid become green because of that ?

See the second commit : https://github.com/replicatedhq/kURL/pull/3891/commits/78f3281b9cf3e7ed22c8734cf64a72cc4fe14f39

Now the script will finish with the error. I did a test with:

Screenshot 2022-12-22 at 15 34 49

banjoh: > We loose the return code of the main function (now it is always zero), aren't all testgrid become green because of that ?

We don't thanks to `set -o pipefail` (line number 5 above)

scripts/install.sh
```diff
@@ -569,4 +570,7 @@ function main() {
     report_install_success
 }
 
-main "$@"
+# tee logs into /var/log/kurl/install-<date>.log and stdout
+mkdir -p /var/log/kurl
+main "$@" 2>&1 | tee -a /var/log/kurl/install-$(date +"%m-%d-%Y-%H-%M-%S").log
```
rrpolanco, line 575: I think it could be helpful to have the kURL installer ID as part of the filename.

banjoh, line 575: Wouldn't it be better to go with the format here? (`date +"%Y-%m-%dT%H-%M-%S"`) biggest to smallest...

banjoh (commented):
This is a welcome change. Thanks for that.

I know its non-trivial, but I wonder how we should treat passwords logged to `stdout` that will now end up in logs. Should they be one time passwords? Should we write them to a file and `cat` them later on? (just throwing ideas, not inclined to any of them :) )

banjoh:
I know these logs files are quite small, but should we have some form of simple log rotation implementation in the script? I'm thinking something like "keep the last 10-15 log files" scenario.

scripts/install.sh
camilamacedo86: Done

camilamacedo86:
Hi @banjoh,

> I know its non-trivial, but I wonder how we should treat passwords logged to stdout that will now end up in logs. Should they be one time passwords? Should we write them to a file and cat them later on, then delete the file? (just throwing ideas, not inclined to any of them :) )

I do not think that we have passwords in the kURL logs at all. Could you please provide an example where this passwords would appears usually?

scripts/install.sh
```diff
@@ -2,6 +2,7 @@
 
 set -e
 set -E # cause 'trap funcname ERR' to be inherited by child commands, see https://stackoverflow.com/questions/35800082/how-to-trap-err-when-using-set-e-in-bash
+set -o pipefail # Ensure the error is propagated past the piped commands
```
camilamacedo86, line 5: @banjoh we cannot use this option
It will make stop to output the data when we run the script.

So, we will need to figure out a way to output and finish with the error.

scripts/install.sh
camilamacedo86: @ricardomaraschini I need to remove that option but see that we can only use tee.

scripts/install.sh
camilamacedo86: Hi @rrpolanco,

> I think it would be helpful to have the kURL installer ID as part of the filename. 

When we will create the log we have not the installedID yet. But we can start to output the spec used for example then we can get it the logs. However, let's leave this one for a follow up? WDYT? Something like "started to output and log the spec". I think is better we try to keep the stories and the PRs shorter as possible. One for each need. 

Hi @banjoh,

> I wonder whether we can echo "$@" as well when the script starts. Logging the entire command would be good.
If we echo it is empty. Note sure what exactly you would like to echo here. 
However, let's it for a follow up as well, if needed. Something like begin to output and log X.

scripts/install.sh
camilamacedo86: Removed

camilamacedo86:
Hi @banjoh, 

> I know these logs files are quite small so disk consumption should not be high in our priority, but should we have some form of simple log rotation implementation in the script? I'm thinking something like "keep the last 10-15 log files" scenario.

Have the logs address the value and users can remove manually.
However, I think that could be another task "Create a log rotation and etc"

banjoh:
> Hi @banjoh,
> 
> > I know these logs files are quite small so disk consumption should not be high in our priority, but should we have some form of simple log rotation implementation in the script? I'm thinking something like "keep the last 10-15 log files" scenario.
> 
> Have the logs address the value and users can remove manually. However, I think that could be another task "Create a log rotation and etc" WDYT? Could you agree?

Makes sense. It can be addressed later on.

scripts/install.sh
banjoh: > 

I think I wasn't very clear :). I would like to see in the logs any [kurl CLI options](https://kurl.sh/docs/install-with-kurl/advanced-options) a script is invoked with. If there are no options, the list will be empty that's why nothing was printed out when you ran `echo "$@"`. Here is a snippet

```sh
function main() {
    echo "ARGS: $@"
}

main "$@"
```

And after running the script
```sh
$ ./tt.sh container-log-max-files=2 exclude-builtin-host-preflights force-reapply-addons
install.sh container-log-max-files=2 exclude-builtin-host-preflights force-reapply-addons

$ cat tt.sh | bash -s container-log-max-files=2 exclude-builtin-host-preflights force-reapply-addons
install.sh container-log-max-files=2 exclude-builtin-host-preflights force-reapply-addons
```

camilamacedo86: Done 👍

scripts/join.sh
```diff
@@ -171,4 +172,8 @@ function main() {
     popd_install_directory
 }
 
-main "$@"
+# tee logs into /var/log/kurl/join-<date>.log and stdout
+mkdir -p /var/log/kurl
+main "$@" 2>&1 | tee -a /var/log/kurl/join-$(date +"%Y-%m-%dT%H-%M-%S").log
+# it is required to return the exit status of the script
+echo ${PIPESTATUS[0]}
```
banjoh, line 179: +1 This is a good find and solution!

camilamacedo86:
Hi @banjoh,

All requests are done now, see that we are also removing from the logs any line that contains the word password. So, it seems good to fly :-) . Could you please give another look and let us know if we have your approval/lgtm here?

scripts/install.sh
```diff
@@ -569,4 +571,9 @@ function main() {
     report_install_success
 }
 
-main "$@"
+# tee logs into /var/log/kurl/install-<date>.log and stdout
+mkdir -p /var/log/kurl
+LOGFILE="/var/log/kurl/install-$(date +"%Y-%m-%dT%H-%M-%S").log"
+main "$@" 2>&1 | tee -a >(grep -vwE "(password)") $LOGFILE
```
banjoh, line 577: I could not get the `tee`d grep to work?

```sh
echo -e "Installing my awesome app\nYour temporary password is 'green'\nInstallation complete" | tee >(grep -vwE "(password)") my.log

...
[evans] $ cat my.log
Installing my awesome app
Your temporary password is 'green'
Installation complete
```

It works the other way though

```sh
[evans] $ echo -e "Installing my awesome app\nYour temporary password is 'green'\nInstallation complete" | grep -vwE "(password)"
Installing my awesome app
Installation complete
```

camilamacedo86: Hi @banjoh, 

It is fixed now. Good catcher, thx. Following the test made:

- a) Get the latest script `curl https://kurl.sh/latest > install.sh`
- b) Run `chmod +x install.sh`
- c) Change the script with the same done in: https://github.com/replicatedhq/kURL/pull/3891/files

Screenshot 2022-12-28 at 12 52 28

- e) Add in the main func the same print that will be done by kots example:

Screenshot 2022-12-28 at 12 19 38

- e) Run the sudo bash ./install
- d) Check that the line `"Login with password (will not be shown again): senha"`  cannot be found in the logs:

Screenshot 2022-12-28 at 12 54 04

Also, to ensure that the script would fail after these changes:

- a) Now, let's run without sudo
- b) See that it finished with the error code:

Screenshot 2022-12-28 at 12 55 23

banjoh (approved):

scripts/upgrade.sh
```diff
@@ -148,4 +148,8 @@ function main() {
     popd_install_directory
 }
 
-main "$@"
+# tee logs into /var/log/kurl/upgrade-<date>.log and stdout
+mkdir -p /var/log/kurl
+main "$@" 2>&1 | tee -a /var/log/kurl/upgrade-$(date +"%Y-%m-%dT%H-%M-%S").log
+# it is required to return the exit status of the script
+echo ${PIPESTATUS[0]}
```
banjoh, line 155: Should be

```
exit ${PIPESTATUS[0]}
```

scripts/tasks.sh
```diff
@@ -717,4 +718,8 @@ function install_host_dependencies_longhorn() {
     longhorn_host_init_common "${DIR}/packages/host/longhorn"
 }
 
-tasks "$@"
+# tee logs into /var/log/kurl/tasks-<date>.log and stdout
+mkdir -p /var/log/kurl
+tasks "$@" 2>&1 | tee -a /var/log/kurl/tasks-$(date +"%Y-%m-%dT%H-%M-%S").log
+# it is required to return the exit status of the script
+echo ${PIPESTATUS[0]}
```
banjoh, line 725: Should be

```
exit ${PIPESTATUS[0]}
```

banjoh (dismissed):
Ship it

laverya:
> Hi @banjoh,
> 
> > I know its non-trivial, but I wonder how we should treat passwords logged to stdout that will now end up in logs. Should they be one time passwords? Should we write them to a file and cat them later on, then delete the file? (just throwing ideas, not inclined to any of them :) )
> 
> I do not think that we have passwords in the kURL logs at all. Could you please provide an example where this passwords would appears usually?

A couple passwords are printed - kotsadm and grafana upon first installation, and the k8s node join tokens are also considered sensitive information and that is displayed on every installation.

laverya:
A few questions - have you tested this with prompts? (As in, do we still prompt for a load balancer address? Confirm Y/N? Etc)

The reason I ask is because I believe we have logic to detect whether the script is running within a pipe, and if so to not do any prompts.

Similarly, do we still trap ctrl+C?

camilamacedo86:
Hi @laverya,

Thank you for your time following the answers inline.

Regards (https://github.com/replicatedhq/kURL/pull/3891#issuecomment-1367182044): **The password is hidden now**. We are using sed to remove any line with this word from the logs. See: https://github.com/replicatedhq/kURL/pull/3891/files#diff-e076e43cb817ecc8b3c2ed18cd58baa50544ecc66293ed6e61129c43ca6a09ccR580

Regards (https://github.com/replicatedhq/kURL/pull/3891#issuecomment-1367183220): That is a good catcher. I am looking on that.

addons/longhorn/1.1.2/install.sh
```diff
@@ -209,11 +209,6 @@ function maybe_init_hosts() {
                     printf "\n\t${GREEN}${prefix}tasks.sh | sudo bash -s longhorn-node-initilize${NC}\n\n"
                 fi
 
-                if ! prompts_can_prompt ; then
-                    logWarn "Install Longhorn prerequisites prompt explicitly ignored"
-                    return
-                fi
-
```
laverya, line 216: I would prefer for this logic to still exist, but have prompts_can_prompt run once before starting the tee (and storing the result)

Does that seem plausible?

camilamacedo86: HI @laverya,

Thank you a lot for looking it.  

I am not sure if I could properly follow up your suggestion. The logic `prompts_can_prompt ` seems to exist only because we were getting the input with `read PROMPT_RESULT < /dev/tty` instead `read -p "<string output>: " PROMPT_RESULT`.

So, why would we need to have prompts_can_prompt running before we starting the tee? 
What is the use case that would require it? What is the workflow the requires it?

scripts/upgrade.sh
```diff
@@ -148,4 +148,11 @@ function main() {
     popd_install_directory
 }
 
-main "$@"
+# tee logs into /var/log/kurl/upgrade-<date>.log and stdout
+mkdir -p /var/log/kurl
+LOGFILE="/var/log/kurl/upgrade-$(date +"%Y-%m-%dT%H-%M-%S").log"
+tasks "$@" 2>&1 | tee -a $LOGFILE
```
emosbaugh, line 154:
```suggestion
main "$@" 2>&1 | tee -a $LOGFILE
```

scripts/common/prompts.sh
```diff
@@ -2,36 +2,18 @@
...
     set +e
-    read PROMPT_RESULT < /dev/tty
+    read -p "(y/n)?" PROMPT_RESULT
     set -e
 }
 
 function confirmY() {
-    printf "(Y/n) "
     if [ "$ASSUME_YES" = "1" ]; then
+        printf "(Y/n) "
```
emosbaugh, line 13: Why is this moving inside the conditional. This is prompting the user with options.

scripts/common/prompts.sh
```diff
@@ -2,36 +2,18 @@
...
         echo "Y"
         return 0
     fi
-    if ! prompts_can_prompt ; then
-        echo "Y"
-        logWarn "Automatically accepting prompt, shell is not interactive"
```
emosbaugh, line 32: What happens now for a non-interactive shell? Say i were to invoke the kurl script from cloud init. I would expect confirmY to continue and confirmN to exit unless ASSUME_YES.

scripts/common/addon.sh
```diff
@@ -191,10 +191,10 @@ function addon_outro() {
         fi
 
         if [ "${KURL_IGNORE_REMOTE_UPGRADE_PROMPT}" != "1" ]; then
-            if prompts_can_prompt ; then
-                echo "Press enter to proceed"
-                prompt
-            fi
+             echo "Press Y to proceed"
```
emosbaugh, line 194: The indentation seems off here

scripts/common/addon.sh
```diff
@@ -267,18 +267,12 @@ function addon_fetch_airgap() {
         printf "You can download it from %s with the following command:\n" "$(get_dist_url)/${package}"
         printf "\n${GREEN}    curl -LO %s${NC}\n\n" "$(get_dist_url)/${package}"
 
-        if ! prompts_can_prompt; then
-            # we can't ask the user to give us the file because there are no prompts, but we can say where to put it for a future run
-            printf "Please move this file to /var/lib/kurl/%s before rerunning the installer.\n" "$package_path"
-            return 1
```
emosbaugh, line 273: i do not understand why this is changing

scripts/common/prompts.sh
```diff
@@ -298,10 +250,6 @@ function prompt_airgap_preload_images() {
             logWarn "Remote load-images task prompt explicitly ignored"
         fi
     done < <(kubectl get nodes --no-headers)
-
-    if [ "$unattended_nodes_missing_images" = "1" ] ; then
-        bail "Preloading images required"
```
emosbaugh, line 303: i do not understand why this is changing

scripts/common/prompts.sh
```diff
@@ -341,18 +289,14 @@ function prompt_for_private_ip() {
         return
     fi
 
-    if ! prompts_can_prompt ; then
-        bail "Multiple network interfaces present, please select an IP address. Try passing the selected address to this script e.g. 'sudo ./install.sh private-address=1.2.3.4' or assign an IP address to the privateAddress field in the kurl add-on."
```
emosbaugh, line 345: i do not understand why this is changing

scripts/common/prompts.sh
```diff
@@ -88,16 +65,11 @@ function prompt_for_token() {
     if [ -n "$KUBEADM_TOKEN" ]; then
         return
     fi
-    if ! prompts_can_prompt ; then
-        bail "kubernetes.kubeadmToken required"
```
emosbaugh, line 92: i do not understand why this is changing

scripts/common/prompts.sh
```diff
@@ -107,14 +79,9 @@ function prompt_for_token_ca_hash() {
     if [ -n "$KUBEADM_TOKEN_CA_HASH" ]; then
         return
     fi
-    if ! prompts_can_prompt ; then
-        bail "kubernetes.kubeadmTokenCAHash required"
```
emosbaugh, line 111: i do not understand why this is changing

scripts/common/prompts.sh
```diff
@@ -126,15 +93,10 @@ function prompt_for_master_address() {
     if [ -n "$KUBERNETES_MASTER_ADDR" ]; then
         return
     fi
-    if ! prompts_can_prompt ; then
-        bail "kubernetes.masterAddress required"
```
emosbaugh, line 130: i do not understand why this is changing

scripts/common/prompts.sh
```diff
@@ -208,15 +170,10 @@ function prompt_for_load_balancer_address() {
     fi
 
     if [ -z "$LOAD_BALANCER_ADDRESS" ]; then
-        if ! prompts_can_prompt ; then
-            bail "kubernetes.loadBalancerAddress required"
```
emosbaugh, line 212: i do not understand why this is changing

emosbaugh (changes requested):

addons/longhorn/1.1.2/install.sh
laverya: The idea behind `prompts_can_prompt` is to determine whether the script is being run in an interactive terminal where the user can respond to dynamic questions (things like "can we upgrade k8s" or "what is your IP address, I can't find it"). Some users run kURL as part of automation where this is not possible. (And so does testgrid)

If we can't figure out the IP and the user can't answer dynamically, we need to exit the script immediately instead of waiting forever for input.

A 'good' way to test this would be to run something like `sudo bash install.sh | tee` - currently IRC 'prompts_can_prompt' would return false in that scenario.

scripts/upgrade.sh
emosbaugh: this does not need append does it?

```suggestion
main "$@" 2>&1 | tee $LOGFILE
```

emosbaugh (changes requested):

scripts/common/prompts.sh
camilamacedo86: Because we are printing it with the read already `https://github.com/replicatedhq/kURL/pull/3891/files#diff-04b0e97fb7895ebf075d4f880f40b32b7fc78978539ff2932c492400967feec8R7` so we do not need print it unless we are in the scenario where we assume yes.

scripts/common/prompts.sh
camilamacedo86: Because instead of `read PROMPT_RESULT < /dev/tty` we are using the read function `read -p "Kubernetes join token:" TOKEN_PROMPT_RESULT` (see the line 70).

Note that `prompts_can_prompt ` will fail always with tee.

scripts/common/prompts.sh
camilamacedo86: Same reason, we are `read -p "Kubernetes discovery token CA hash: " PROMPT_RESULT` instead.

scripts/common/prompts.sh
camilamacedo86: Same we are `read -p "Kubernetes master address: " PROMPT_RESULT` instead

scripts/common/prompts.sh
camilamacedo86: This one was removed because `unattended_nodes_missing_images=1` scenario could only occurs if:

``` 
                unattended_nodes_missing_images=1
                continue
            fi
```

Therefore, we removed the `prompts_can_prompt ` so we do not need this one anymore. The `unattended_nodes_missing_images` is unused code now.

scripts/common/prompts.sh
camilamacedo86: Because it would only occurs if `prompts_can_prompt ` which was removed.

scripts/install.sh
```diff
@@ -569,4 +571,11 @@ function main() {
     report_install_success
 }
 
-main "$@"
+# tee logs into /var/log/kurl/install-<date>.log and stdout
+mkdir -p /var/log/kurl
+LOGFILE="/var/log/kurl/install-$(date +"%Y-%m-%dT%H-%M-%S").log"
+main "$@" 2>&1 | tee -a $LOGFILE
```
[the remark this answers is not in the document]
camilamacedo86: I think you are right we do not need.

addons/longhorn/1.1.2/install.sh
camilamacedo86: In fact, I discovered that the problem was only the condition see: https://replicated.slack.com/archives/C0499EGNPNY/p1672836175693969?thread_ts=1672774404.239299&cid=C0499EGNPNY

So, I am closing it as outdated.

emosbaugh (approved):

PR #3892, opened 2022-12-22
chore: add .golangci config to add common used checks to help us improve the quality

camilamacedo86:

The golangci default config does not check all that is possible
This PR is adding common usage checks and sorting out some issues found with the new lints

PS.: See that some of them are commented/disabled so that we can sort those in a follow up to make easier the review.


.golangci.yml
```diff
@@ -0,0 +1,71 @@
...
+     - ineffassign
+#    - lll (TODO: Fiz size of the lines, as a good practice we should not need to use scroll to read the code)
+     - misspell
+     - nakedret
+     - revive
+     - staticcheck
+     - typecheck
+#    - unconvert TODO: check the 2 unnecessary conversions found to enable this one
+     - unparam
+     - unused
```
camilamacedo86, line 71:
```suggestion
     - unused

```

ricardomaraschini (commented):
You have my LGTM here for sure. Let's just wait for someone else's review as well.

emosbaugh (approved):

PR #3895, opened 2022-12-23
chore: fix longhorn generate script by skipping version 1.2.6

camilamacedo86:

Longhorn generate is not working, see

https://github.com/replicatedhq/kURL/actions/runs/3763084697/jobs/6396325888

The reason is because the script tried to add 1.2.6 (which is lower than the latest version) and fails because this one has not all expected manifests. 

Fixes # [sc-65099]

Screenshot 2022-12-23 at 08 50 02

## Steps to reproduce
Just try to run the script

NONE

NONE

### DIFF
--- a/addons/longhorn/template/generate.sh
+++ b/addons/longhorn/template/generate.sh
@@ -105,6 +105,12 @@ function add_as_latest() {
 
 function main() {
     get_latest_release_version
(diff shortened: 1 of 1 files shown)


ricardomaraschini (approved):

laverya (approved):
I'd prefer a more general solution, but this is okay I suppose

PR #3899, opened 2022-12-28
chore(npm dep): update npm install to try sorted out dependabot prs

camilamacedo86:

We are here only commiting the changes in the versions when we run `npm install` so that we can see if we can the dependabot fix prs run properly, see:  https://github.com/replicatedhq/kURL/security/dependabot

Fixes # NONE


emosbaugh (approved):

package-lock.json
```diff
@@ -1,943 +1,8 @@
 {
   "name": "kurl",
   "version": "1.0.0",
-  "lockfileVersion": 2,
+  "lockfileVersion": 1,
```
rrpolanco, line 4: Can you explain why this is changing? Doesn't NPM set this?

package-lock.json
```diff
@@ -1,943 +1,8 @@
...
   "requires": true,
-  "packages": {
```
rrpolanco, line 6: Is the goal to _not_ explicitly set the package versions so there aren't any conflicts with dependabot proposed changes? What's the difference between ths _packages_ section vs the package.json [_depenedencies_](https://github.com/replicatedhq/kURL/blob/1e21097e1844c1a6f91a46be60b0778075bd00e0/package.json#L9-L13) section?

package-lock.json
camilamacedo86: Yes, the changes in this PR happened by running npm install only.

camilamacedo86:
It seems no longer required.
After we merge all bots update and I update the npm version used locally the latest version is no longer modified.

PR #3920, opened 2023-01-06
Multi-node weave to flannel

laverya:

~In testing, single-master weave installs migrated to flannel with the same code as single-node installs - no additional scripts required. So this enables such migrations.~

Fixes #

```release-note
Allow migrating HA Weave installs to Flannel
```

### DIFF
--- a/addons/flannel/0.20.2/install.sh
+++ b/addons/flannel/0.20.2/install.sh
@@ -14,8 +14,8 @@ function flannel_pre_init() {
     local dst="$DIR/kustomize/flannel"
 
     if flannel_weave_conflict ; then
-        if ! flannel_is_single_node; then
-            bail "Migrations from Weave to Flannel are not supported"
+        if ! flannel_is_single_master; then
+            bail "Migrations from Weave to Flannel are not supported on HA installs"
         fi
     fi
     if flannel_antrea_conflict ; then
@@ -93,9 +93,9 @@ function flannel_antrea_conflict() {
     ls /etc/cni/net.d/*antrea* >/dev/null 2>&1
 }
 
-function flannel_is_single_node() {
+function flannel_is_single_master() {
     local nodecount=
-    nodecount=$(kubectl get nodes --no-headers | wc -l)
+    nodecount=$(kubectl get nodes --no-headers --selector='node-role.kubernetes.io/control-plane' | wc -l)
     if [ "$nodecount" -eq 1 ]; then
         return 0
     else
--- a/addons/flannel/0.20.2/install.sh
+++ b/addons/flannel/0.20.2/install.sh
@@ -13,11 +13,6 @@ function flannel_pre_init() {
     local src="$DIR/addons/flannel/$FLANNEL_VERSION"
     local dst="$DIR/kustomize/flannel"
 
-    if flannel_weave_conflict ; then
-        if ! flannel_is_single_master; then
-            bail "Migrations from Weave to Flannel are not supported on HA installs"
-        fi
-    fi
     if flannel_antrea_conflict ; then
         bail "Migrations from Antrea to Flannel are not supported"
     fi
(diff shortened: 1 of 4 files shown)


emosbaugh (approved):

addons/flannel/0.20.2/install.sh
```diff
@@ -14,10 +14,11 @@ function flannel_pre_init() {
     local dst="$DIR/kustomize/flannel"
 
     if flannel_weave_conflict ; then
-        if ! flannel_is_single_node; then
-            bail "Migrations from Weave to Flannel are not supported"
+        if [ -n "$DOCKER_VERSION" ]; then
+            bail "Migrations from Weave to Flannel are only supported with containerd"
```
camilamacedo86, line 18: Could we add here, please migrate your host to use containerd and then, re-run this script? What do you think?

addons/flannel/0.20.2/install.sh
```diff
@@ -114,6 +116,60 @@ function weave_to_flannel() {
...
+
+        if [ "$AIRGAP" = "1" ]; then
+            printf "\n\t${GREEN}cat ./tasks.sh | sudo bash -s weave-to-flannel-secondary${NC}\n\n"
+        else
+            local prefix=
+            prefix="$(build_installer_prefix "${INSTALLER_ID}" "${KURL_VERSION}" "${KURL_URL}" "${PROXY_ADDRESS}")"
+            printf "\n\t${GREEN}${prefix}tasks.sh | sudo bash -s weave-to-flannel-secondary${NC}\n\n"
+        fi
+
+        printf "${YELLOW}Once this has been run on all nodes, press enter to continue.${NC}"
```
camilamacedo86, line 168: how the user knows that it was executed in all nodes?
Are they watching the changes?

addons/flannel/template/base/install.sh
```diff
@@ -114,6 +116,60 @@ function weave_to_flannel() {
...
+    # if there is more than one node, prompt to run on each primary/master node, and then on each worker/secondary node
+    local master_node_count=
+    master_node_count=$(kubectl get nodes --no-headers --selector='node-role.kubernetes.io/control-plane' | wc -l)
+    local hostnamevar
+    hostnamevar=$(hostname)
+    local master_node_names=
+    master_node_names=$(kubectl get nodes --no-headers --selector='node-role.kubernetes.io/control-plane' -o custom-columns=NAME:.metadata.name | grep -v "$hostnamevar")
+    if [ "$master_node_count" -gt 0 ]; then
+        printf "${YELLOW}Moving primary nodes from Weave to Flannel requires removing certain weave files and restarting kubelet.${NC}\n"
+        printf "${YELLOW}Please run the following command on each of the listed primary nodes:${NC}\n\n"
```
camilamacedo86, line 129: Could we not run the commands instead of ask for the user to do it?

scripts/tasks.sh
```diff
@@ -718,6 +724,97 @@ function install_host_dependencies_longhorn() {
...
+
+function weave_to_flannel_secondary() {
+    task_requires_root
+
+    rm -f /opt/cni/bin/weave-*
+    rm -rf /etc/cni/net.d
+    ip link delete weave
+    systemctl restart kubelet containerd
+
+    logSuccess "Successfully updated $(hostname) to use Flannel"
```
camilamacedo86, line 807: is possible to use an kubectl command to ensure that all is ok before print it out?

scripts/tasks.sh
```diff
@@ -718,6 +724,97 @@ function install_host_dependencies_longhorn() {
...
+}
+
+function task_requires_root() {
```
camilamacedo86, line 810: have we not a func that does that. Could we re-use instead of create another one?

scripts/tasks.sh
```diff
@@ -718,6 +724,97 @@ function install_host_dependencies_longhorn() {
...
+                ;;
+        esac
+        shift
+    done
+
+    task_requires_root
+
+    # if CERT_KEY was not provided, we cannot continue
+    if [ -z "$CERT_KEY" ]; then
+        bail "cert-key is required"
```
camilamacedo86, line 748: could we try to improve the message and add the steps / where / how the CERT KEY should be provided so that if it fails users knows what they must to do?

scripts/tasks.sh
```diff
@@ -718,6 +724,97 @@ function install_host_dependencies_longhorn() {
...
+    shift
+    while [ "$1" != "" ]; do
+        _param="$(echo "$1" | cut -d= -f1)"
+        _value="$(echo "$1" | grep '=' | cut -d= -f2-)"
+        case $_param in
+            cert-key)
+                CERT_KEY="$_value"
+                ;;
+            *)
+                echo >&2 "Error: unknown parameter \"$_param\""
```
camilamacedo86, line 737: could we not use here logFail to ensure that will print with its colours and etc?

addons/flannel/template/base/install.sh
laverya: We don't have a way to run commands on other nodes ourselves

scripts/tasks.sh
laverya: I didn't want to import all of the preflight stuff into tasks

scripts/tasks.sh
laverya: Not really, because it won't necessarily be okay until all of the main notes have been updated.

And of course worker nodes can't run kubectl at all

scripts/tasks.sh
laverya: The cert key should always be provided as part of the command that was given for a user to run

emosbaugh (changes requested):

scripts/tasks.sh
```diff
@@ -718,6 +724,97 @@ function install_host_dependencies_longhorn() {
...
+    node-ip: $current_node_ip
+    node-labels: kurl.sh/cluster=true,
+  taints: []
+EOM
+
+    rm -f /opt/cni/bin/weave-*
+    rm -rf /etc/cni/net.d
+    ip link delete weave
+
+    kubeadm join phase control-plane-prepare control-plane --config=/tmp/kubeadm-join.conf
```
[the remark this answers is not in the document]
laverya: Secondary nodes don't have kube-controller-manager as far as I'm aware, and that was the only thing that changed:
```
diff -r manifests.bak/ manifests-updated/
diff -r manifests.bak/manifests/kube-controller-manager.yaml manifests-updated/manifests/kube-controller-manager.yaml
13a14
>     - --allocate-node-cidrs=true
17a19
>     - --cluster-cidr=10.32.16.0/20
27a30
>     - --service-cluster-ip-range=10.96.0.0/22
```

addons/flannel/template/base/install.sh
```diff
@@ -93,9 +95,9 @@ function flannel_antrea_conflict() {
     ls /etc/cni/net.d/*antrea* >/dev/null 2>&1
 }
 
-function flannel_is_single_node() {
+function flannel_is_single_master() {
```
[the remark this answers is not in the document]
laverya: it is now, I can delete it

addons/flannel/template/base/install.sh
```diff
@@ -34,7 +35,8 @@ function flannel() {
     flannel_render_config
 
     if flannel_weave_conflict; then
-        printf "${YELLOW}Would you like to migrate from Weave to Flannel?${NC}"
+        printf "${YELLOW}Would you like to migrate from Weave to Flannel?${NC}\n"
+        printf "${YELLOW}This will require whole-cluster downtime during the transition process. ${NC}"
```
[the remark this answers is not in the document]
laverya: can do

addons/flannel/template/base/install.sh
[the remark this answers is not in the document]
laverya: I'll look

addons/flannel/template/base/install.sh
```diff
@@ -114,6 +116,60 @@ function weave_to_flannel() {
...
     kubectl -n kube-flannel apply -k "$dst/"
+
+    # if there is more than one node, prompt to run on each primary/master node, and then on each worker/secondary node
+    local master_node_count=
+    master_node_count=$(kubectl get nodes --no-headers --selector='node-role.kubernetes.io/control-plane' | wc -l)
+    local hostnamevar
+    hostnamevar=$(hostname)
+    local master_node_names=
+    master_node_names=$(kubectl get nodes --no-headers --selector='node-role.kubernetes.io/control-plane' -o custom-columns=NAME:.metadata.name | grep -v "$hostnamevar")
+    if [ "$master_node_count" -gt 0 ]; then
```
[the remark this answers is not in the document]
laverya: Already done at time of comment 😆

scripts/tasks.sh
laverya: Well, initially my response was "because copilot", I'll fix

but apparently copilot copied the code from `migrate_rgw_to_minio_task`, `migrate_pvcs`, and `join_token` in the same file 😆, so I'm keeping it

addons/flannel/0.20.2/install.sh
laverya: maaaaaybe, but this is really a 'vendor needs to fix' problem and not a 'user can fix' problem if that makes any sense

the user can't migrate from docker->containerd on their own, they need the vendor to update the spec

addons/flannel/template/base/install.sh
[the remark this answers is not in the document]
laverya: Any objections to merging this and adding better termination later?

Right now it'll stall during the wait-for-flannel-health step shortly after this, so it's not like it'll continue forever even if you hit enter early

emosbaugh (approved):

PR #3921, opened 2023-01-06
bug: don't attempt to migrate from longhorn on unsupported version

ricardomaraschini:

Due to compatibility issues, if the version of Kubernetes being used is v1.25 or higher, it will not be possible for Longhorn to operate properly and the migration process will fail. This commit introduces a message that will be displayed and the process will immediately terminate upon detection of a Kubernetes version of v1.25 or higher.

Fixes #58505

### DIFF
--- a/addons/openebs/template/base/install.sh
+++ b/addons/openebs/template/base/install.sh
@@ -374,6 +374,16 @@ function openebs_prompt_migrate_from_longhorn() {
     logWarn "    Detected Longhorn is running in the cluster. Data migration will be initiated to move data from Longhorn to storage class $OPENEBS_LOCALPV_STORAGE_CLASS."
     logWarn "    As part of this, all pods mounting PVCs will be stopped, taking down the application."
     logWarn "    It is recommended to take a snapshot or otherwise back up your data before proceeding."
+
+    semverParse "$KUBERNETES_VERSION"
+    if [ "$minor" -gt 24 ] ; then
+        logFail "    It appears that the Kubernetes version you are attempting to install ($KUBERNETES_VERSION) is incompatible with the version of Longhorn currently installed"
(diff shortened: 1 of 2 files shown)


addons/openebs/template/base/install.sh
```diff
@@ -374,6 +374,16 @@ function openebs_prompt_migrate_from_longhorn() {
     logWarn "    Detected Longhorn is running in the cluster. Data migration will be initiated to move data from Longhorn to storage class $OPENEBS_LOCALPV_STORAGE_CLASS."
     logWarn "    As part of this, all pods mounting PVCs will be stopped, taking down the application."
     logWarn "    It is recommended to take a snapshot or otherwise back up your data before proceeding."
+
+    semverParse "$KUBERNETES_VERSION"
```
camilamacedo86, line 378: Should not the check be in the preflights (https://github.com/replicatedhq/kURL/blob/main/scripts/common/preflights.sh) ?

ricardomaraschini: I don't have strong opinion on the matter. I read the `preflight.sh` and it seems to contain a lot of checks regarding the node where we are planning to install, in the other hand this file here is pretty much "openebs" specific. If we move this from here to there we gonna have to migrate other logic that actually exist in other migrations. Yeah, I think it is better to keep it here but I am open to be convinced otherwise.

camilamacedo86: So, that is fine. I do not have strong opinion either.
I do not think that is a blocker at all.

emosbaugh (approved):
is there an incompatibility here we could add to the linter?

ricardomaraschini:
> is there an incompatibility here we could add to the linter?

The problem here is that Longhorn is not selected in the Installer but Rook is. On this scenario we can't upgrade to a Kubernetes version where Longhorn is not supported otherwise we can't migrate from it. We can't catch this scenario in the linter, and that is unfortunate.

PR #3960, opened 2023-01-18
add kurl_machine_uuid reporting, remove redundant/unused machine_id reporting

laverya:

This creates a file `/var/lib/kurl/uuid` when not present, and stores a UUID there. That UUID is reported in addition to the existing `machine_id` that is apparently not always unique on cloned VMs.

Reporting of `machine_id` was also reduced to only be where the parameter was useful/used - at the init step. There's no point to sending it with each request, since we only store it from the "beginning of the install" reporting line anyways. (and it doesn't change)

Fixes #

```release-note

```

### DIFF
--- a/scripts/common/common.sh
+++ b/scripts/common/common.sh
@@ -860,8 +860,16 @@ function maybe() {
 }
 
 MACHINE_ID=
+KURL_INSTANCE_UUID=
 function get_machine_id() {
     MACHINE_ID="$(${DIR}/bin/kurl host protectedid || true)"
+    if [ -f /var/lib/kurl/uuid ]; then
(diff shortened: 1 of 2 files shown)


scripts/common/common.sh
```diff
@@ -860,8 +860,16 @@ function maybe() {
 }
 
 MACHINE_ID=
+KURL_INSTANCE_UUID=
 function get_machine_id() {
     MACHINE_ID="$(${DIR}/bin/kurl host protectedid || true)"
+    if [ -f /var/lib/kurl/uuid ]; then
```
emosbaugh, line 866: will this work in dev? do you need to mkdir -p /var/lib/kurl or use $DIR?

emosbaugh (approved):

PR #3963, opened 2023-01-18
check that flannel images are present on remote nodes in airgap

laverya:
When the on-each-node weave to flannel migration script is run, it checks if there is a flannel image present if the instance is airgapped. This way, we can catch issues that would prevent the flannel daemonset from ever becoming healthy.

Fixes #

```release-note

```

### DIFF
--- a/addons/flannel/0.20.2/install.sh
+++ b/addons/flannel/0.20.2/install.sh
@@ -132,7 +132,7 @@ function weave_to_flannel() {
         else
             local prefix=
             prefix="$(build_installer_prefix "${INSTALLER_ID}" "${KURL_VERSION}" "${KURL_URL}" "${PROXY_ADDRESS}")"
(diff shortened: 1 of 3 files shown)


emosbaugh (approved):
does the script prompt the user to load the images on the remote nodes prior to running the migration?

laverya:
Yes - the main install script calls common_prompts here: https://github.com/replicatedhq/kURL/blob/f7ad6b1125db48dd3bed4ef20dc346a684dc588b/scripts/install.sh#L535, which calls prompt_airgap_preload_images https://github.com/replicatedhq/kURL/blob/fbae1a26b5d4d920813114592dd155797a4b391a/scripts/common/prompts.sh#L145-L156 which does the preload prompting https://github.com/replicatedhq/kURL/blob/fbae1a26b5d4d920813114592dd155797a4b391a/scripts/common/prompts.sh#L259

PR #3967, opened 2023-01-19
Add metrics-server 0.6.2

laverya:

Our currently shipped version of metrics-server is quite out of date, but we don't list it as being deprecated. So this adds the latest.

Fixes #

```release-note
Adds metrics-server 0.6.2
```

### DIFF
--- a/addons/metrics-server/template/generate.sh
+++ b/addons/metrics-server/template/generate.sh
@@ -1,0 +1,89 @@
+#!/bin/bash
+
+set -euo pipefail
+
+function generate() {
+    local dir="../$version"
+
+    mkdir -p "$dir"
+
+    cp -r ./base/* "$dir"
+
+    download_yaml
+
+    find_images_in_yaml
+}
+
+function download_yaml() {
+    curl -Lo "$dir/components.yaml" "https://github.com/kubernetes-sigs/metrics-server/releases/download/v$version/components.yaml"
+}
+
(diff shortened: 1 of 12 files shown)


camilamacedo86 (dismissed):

rrpolanco (dismissed):
lgtm

rrpolanco (dismissed):

rrpolanco:
@laverya Should we have a template installer for this add-on? 
I think it's worthwhile to consider automating the version releases like we do for the other add-ons.

laverya:
IMO "maybe"
I'll give it a look

addons/metrics-server/template/generate.sh
```diff
@@ -0,0 +1,89 @@
...
+        else
+            echo "not updating existing version of metrics-server"
+            return
+        fi
+    fi
+
+    generate "$version"
+    add_as_latest "$version"
+
+    echo "::set-output name=metrics_server_version::$version"
```
rrpolanco, line 86: This will work for _now_ but it's a deprecated way of setting state: https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/

laverya: yeah, but it's what we use in other generate scripts, and we should probably update all of those at once

rrpolanco (approved):

PR #3970, opened 2023-01-19
Experimental support for k0s

emosbaugh:

Adds experimental support for k0s.

Add-ons
| Add-on | Working |
| ------- | --------- |
| rook | ✓ |
| ekco | |
| openebs | |
| minio | |
| contour | |
| registry | |
| prometheus | |
| kotsadm | ✓ |
| velero | |

Fixes #

```release-note

```

### DIFF
--- a/addons/rook/1.10.8/install.sh
+++ b/addons/rook/1.10.8/install.sh
@@ -163,6 +163,10 @@ function rook_operator_deploy() {
         insert_patches_strategic_merge "$dst/kustomization.yaml" patches/deployment-privileged.yaml
     fi
 
+    if [ "$K8S_DISTRO" = "k0s" ]; then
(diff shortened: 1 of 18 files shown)


chris-sanders:
This seems like it's headed in the right direction for what we discussed. Seeing it here with hundreds of lines of bash makes me curious if k0sctl even for a single node install where we can just provide a config file and not worry about state _at all_ would be cleaner/better. It seems to support [local host installs](https://wiki.replicated.io/doc/engineering-career-framework-1TLfSGeYrj). Maybe this is harder than I'm thinking I'm just wondering if we can take all 'discovery' out.

Even with that, this seems really cool is it possible to try this out or does it have to be merged before you could try it on an actual machine?

emosbaugh:
I am closing this as we are unlikely to go in this direction

PR #3971, opened 2023-01-20
add serverFlags field to velero add-on

cbodonnell:

This PR adds support for an additional `serverFlags` field in the Velero add-on.  This field accepts a list of strings that will be added as additional arguments to the `velero server` command in the velero deployment.

Depends on: https://github.com/replicatedhq/kurlkinds/pull/45

Related kurl-api: https://github.com/replicatedhq/kURL-api/pull/128

Fixes [SC-35682](https://app.shortcut.com/replicated/story/35682/deleting-snapshot-does-not-reclaim-space-from-restic-for-5-days)

The existing `resticTimeout` implementation was modified so that it will instead leverage this same logic for appending additional arguments to the `velero server` command.  This field could potentially be deprecated at some point since it could be configured in the `serverFlags` field now as well.

## Steps to reproduce

Install Spec:
```
...
    velero:
        version: 1.9.5
        serverFlags:
        - --log-level
        - debug
        - --default-backup-ttl=24h
```

CLI Flag:
```
cat install.sh | sudo bash -s velero-server-flags=--log-level,debug,--default-backup-ttl=12h
```

Result:
```
$ kubectl get deploy velero -n velero -o yaml
...
      - args:
        - server
        - --features=
        - --log-level
        - debug
        - --default-backup-ttl=24h
        command:
        - /velero
...
```

```release-note
Adds a `serverFlags` configuration option to the [Velero add-on](https://kurl.sh/docs/add-ons/velero) to allow users to pass additional flags to the `velero server` command in the Velero pod. This can also be set using the [velero-server-flags](https://kurl.sh/docs/install-with-kurl/advanced-options#reference) cli flag when running the install script.
```

https://github.com/replicatedhq/kurl.sh/pull/929


cbodonnell:
Should this be added to _all_ Velero add-on versions?  Or only the template and most recent version?

ricardomaraschini:
> Should this be added to _all_ Velero add-on versions? Or only the template and most recent version?

That is an interesting question. Are the flags supported in all releases ? What is the impact of some versions supporting them and others not ? I think that if we are going to introduce this partially (not across all versions) then we should add a [lint rule](https://github.com/replicatedhq/kurlkinds/blob/main/pkg/lint/rego/output.rego) indicating that this param does not work with versions < X.

cbodonnell:
> > Should this be added to _all_ Velero add-on versions? Or only the template and most recent version?
> 
> That is an interesting question. Are the flags supported in all releases ? What is the impact of some versions supporting them and others not ? I think that if we are going to introduce this partially (not across all versions) then we should add a [lint rule](https://github.com/replicatedhq/kurlkinds/blob/main/pkg/lint/rego/output.rego) indicating that this param does not work with versions < X.

I believe they should be supported in all add-on versions since we did the same for `resticTimeout`.  I'll do some testing and update this PR.

emosbaugh (commented):
@cbodonnell is it a requirement to backport this change to all previous velero releases? typically we only update the latest and sometimes backport a few versions if there is a requirement. this can also help drive customers to new versions.

sgalsaleh (approved):

cbodonnell:
> @cbodonnell is it a requirement to backport this change to all previous velero releases? typically we only update the latest and sometimes backport a few versions if there is a requirement. this can also help drive customers to new versions.

No, I don't believe that's a requirement here (and if it becomes one we can do that at a later time).  If updating the template and latest add-on version is enough, then that is probably best.

sgalsaleh (approved):

addons/velero/template/base/install.tmpl.sh
```diff
@@ -244,12 +244,33 @@ function velero_patch_args() {
     local src="$1"
     local dst="$2"
 
-    if [ "${VELERO_DISABLE_RESTIC}" = "1" ] || [ -z "${VELERO_RESTIC_TIMEOUT}" ]; then
-        return 0
+    # if the user has specified any additional velero server flags, add them to the velero deployment
+    if [ -n "$VELERO_SERVER_FLAGS" ]; then
+        # iterate over the flags in reverse order since they are prepended to the list of kustomize patches
+        IFS=',' read -ra flags <<< "$VELERO_SERVER_FLAGS"
```
emosbaugh, line 250: it may be safest to restore the value of IFS for any subsequent calls to the read utility

```suggestion
        oIFS="$IFS"; IFS=',' read -ra flags <<< "$VELERO_SERVER_FLAGS"; IFS="$oIFS"
```

emosbaugh (changes requested):

addons/velero/template/base/install.tmpl.sh
emosbaugh: This seems ok as is based on https://unix.stackexchange.com/a/18889

emosbaugh (approved):

addons/velero/template/base/install.tmpl.sh
cbodonnell: leaving this as-is since it would only be defined for the read invocation

PR #3972, opened 2023-01-22
(chore): testgrid: add some tests to check kubectl with kube/config

camilamacedo86:

Adding some tests to check kubectl with kube/config

Fixes # [sc-63954]

Add the for all tests that we have postinstal scripts today just to ensure that we do not break the kubectl behaviour if we unset KUBECONFIG

### DIFF
--- a/testgrid/specs/deploy.yaml
+++ b/testgrid/specs/deploy.yaml
@@ -182,6 +182,10 @@
     kube_bench_version="$(curl -s https://api.github.com/repos/aquasecurity/kube-bench/releases/latest | grep '"tag_name":' | sed -E 's/.*"v([^"]+)".*/\1/')"
     curl -L https://github.com/aquasecurity/kube-bench/releases/download/v${kube_bench_version}/kube-bench_${kube_bench_version}_linux_amd64.tar.gz | tar -xz
     ./kube-bench --config-dir=`pwd`/cfg --config=`pwd`/cfg/config.yaml --exit-code=1
+    echo "Checking kubectl with kube/config"  
+    echo unset KUBECONFIG >> ~/.bash_profile
+    bash -l
+    kubectl get namespaces
 - name: k8s126x_reserved_resources
   installerSpec:
     kubernetes:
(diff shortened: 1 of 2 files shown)


testgrid/specs/deploy.yaml
```diff
@@ -182,6 +182,10 @@
     kube_bench_version="$(curl -s https://api.github.com/repos/aquasecurity/kube-bench/releases/latest | grep '"tag_name":' | sed -E 's/.*"v([^"]+)".*/\1/')"
     curl -L https://github.com/aquasecurity/kube-bench/releases/download/v${kube_bench_version}/kube-bench_${kube_bench_version}_linux_amd64.tar.gz | tar -xz
     ./kube-bench --config-dir=`pwd`/cfg --config=`pwd`/cfg/config.yaml --exit-code=1
+    echo "Checking kubectl with kube/config"  
+    echo unset KUBECONFIG >> ~/.bash_profile
+    bash -l
+    kubectl get namespaces
```
laverya, line 188:
```suggestion
    echo "Checking kubectl with kube/config"  
    echo "Kubeconfig was $KUBECONFIG"
    unset KUBECONFIG
    kubectl get namespaces
```

I'd rather not do a `bash -l` here

camilamacedo86: Hi @laverya,

Why should we use `unset KUBECONFIG` instead of `echo unset KUBECONFIG >> ~/.bash_profile`  which is the command outputted by kURL?

emosbaugh: We may want to consider changing this https://github.com/replicatedhq/kURL-testgrid/blob/bcd4284baa85a1df84e73a073fff8f482dbcdd77/tgrun/pkg/runner/vmi/embed/runcmd.sh#L53

camilamacedo86: Kurl/kubeadm setup ` KUBECONFIG=/etc/kubernetes/admin.conf`, if we remove https://github.com/replicatedhq/kURL-testgrid/blob/bcd4284baa85a1df84e73a073fff8f482dbcdd77/tgrun/pkg/runner/vmi/embed/runcmd.sh#L53 I understand that would not have any change. 

Also, would not better we just change a bunch of tests instead of all to cover both scenarios?

laverya: @camilamacedo86 because just running the command directly (instead of putting it in the bash profile and reloading) lets us do this without reloading

And personally I expect there to be issues with our test code when reloading

camilamacedo86: Hi @laverya and @emosbaugh,

I followed up the @laverya suggestion here. WDYT ? 
Could we move forward now? 
OR should we find a better away to test it out?

laverya (approved):

PR #3973, opened 2023-01-22
feat: improve the output message to clarify to users kubectl config

camilamacedo86:

Kurl uses and export KUBECONFIG with kubernetes/admin.conf.
However, it also created the .kube/config by copying kubernetes/admin.conf to kube/config

Therefore, the output does not clarifies why unset KUBECONFIG and what is used by kurl.
Also, it is not required unset KUBECONFIG to use kubectl as described in the output. 

So, the goal of this PR is try to clarifies it to the users and sorted out the current misleading message 

Fixes # [sc-63954]

Before this PR: 

Screenshot 2023-01-22 at 09 55 40

After this PR:

Screenshot 2023-01-22 at 10 14 08

Also:

Screenshot 2023-01-22 at 10 19 51

## Steps to reproduce
Just install and re-install

```release-note
Improve output message to clarifies the kubectl config used by kurl
```

NONE

### DIFF
--- a/scripts/common/common.sh
+++ b/scripts/common/common.sh
@@ -604,15 +604,21 @@ function kubeconfig_setup_outro() {
             cp "$(${K8S_DISTRO}_get_kubeconfig)" $ownerdir/.kube/config
             chown -R $owner $ownerdir/.kube
 
-            printf "To access the cluster with kubectl, ensure the KUBECONFIG environment variable is unset:\n"
+            printf "To access the cluster with kubectl:\n"
+            printf "\n"
+            printf "${GREEN}    bash -l${NC}\n"
+            printf "Kurl uses "$(${K8S_DISTRO}_get_kubeconfig)", you might want to unset KUBECONFIG to use .kube/config:\n"
             printf "\n"
             printf "${GREEN}    echo unset KUBECONFIG >> ~/.bash_profile${NC}\n"
-            printf "${GREEN}    bash -l${NC}\n"
             return
         fi
     fi
 
-    printf "To access the cluster with kubectl, copy kubeconfig to your home directory:\n"
+    printf "To access the cluster with kubectl:\n"
+    printf "\n"
+    printf "${GREEN}    bash -l${NC}\n"
+    printf "\n"
+    printf "Kurl uses "$(${K8S_DISTRO}_get_kubeconfig)", you might want to copy kubeconfig to your home directory:\n"
     printf "\n"
     printf "${GREEN}    cp "$(${K8S_DISTRO}_get_kubeconfig)" ~/.kube/config${NC}\n"
     printf "${GREEN}    chown -R ${owner} ~/.kube${NC}\n"


scripts/common/common.sh
```diff
@@ -604,15 +604,21 @@ function kubeconfig_setup_outro() {
             cp "$(${K8S_DISTRO}_get_kubeconfig)" $ownerdir/.kube/config
             chown -R $owner $ownerdir/.kube
 
-            printf "To access the cluster with kubectl, ensure the KUBECONFIG environment variable is unset:\n"
+            printf "To access the cluster with kubectl:\n"
             printf "\n"
-            printf "${GREEN}    echo unset KUBECONFIG >> ~/.bash_profile${NC}\n"
             printf "${GREEN}    bash -l${NC}\n"
+            printf "Kurl uses "$(${K8S_DISTRO}_get_kubeconfig)", you might want to unset KUBECONFIG to use .kube/config:\n"
```
laverya, line 610: This is IMO still missing the point that _we're the ones that are adding the `KUBECONFIG` env var to the bash profile_, and we should stop doing that

camilamacedo86: Hi @laverya, @ricardomaraschini.

Following the comments inline 

> This is IMO still missing the point that we're the ones that are adding the KUBECONFIG env var to the bash profile, and we should stop doing that

kurl (kubeadmin) uses  `/etc/kubernetes/admin.conf` we never spoke in stop to do so or change that. 


> I agree. It is also not clear what we are trying to solve with this

What we are trying to do here is:
- Stop to say for the user that is required unset KUBECONFIG to have access to the kubectl because that is NOT required at all. 
- Make clear that kURL uses and updates   `/etc/kubernetes/admin.conf`  and NOT kube/config. 

>  IMHO we should keep the file under /etc/kubernetes/admin.conf and just inform that (as we have been doing)
Yes, it is what we are doing here., We are only clarifying it for the user. 

>  That is a file that shouldn't be moved anywhere as it provides administrative rights to the cluster.
We are not moving any file in this PR, so could you please clarify what do you mean with?

ricardomaraschini: Yeah, my comments do not belong here and I think the train has already departed so, never mind. My sole point was: we should not copy `/etc/kubernetes/admin.conf` to the user's home directory (it is a file providing administrative rights to the cluster and I prefer to have a single source of truth).

camilamacedo86: Hi @ricardomaraschini,

> we should not copy /etc/kubernetes/admin.conf to the user's home directory

We created the .kube/config using it today and before. 
That is the kubeadmin recommendation (also it is outputted by) see:

Screenshot 2023-01-23 at 09 55 20

Kubeadmin docs: https://kubernetes.io/docs/setup/production-environment/tools/kubeadm/create-cluster-kubeadm/#more-information

Therefore, change how it works never was discussed. This PR only change the output texted because it is not accurate (it is not required to unset KUBECONFIG to get the kubectl working at all) and clarifies that  `/etc/kubernetes/admin.conf` is what is used by kURL.

ricardomaraschini: Yes, thank you.

camilamacedo86: @laverya, could you please check out the above comment and let us know if that is also making sense for you now and if we have your /lgtm /approval to move forward here? if not, could you please let us know why not?

laverya: this is printing:
```
To access the cluster with kubectl:

    bash -l

Kurl uses "/etc/kubernetes/admin.conf", you might want to unset KUBECONFIG to use .kube/config:

    echo unset KUBECONFIG >> ~/.bash_profile
```

I'm still not sure this wording is ok.

Give me a minute to try and figure out something better

camilamacedo86: Sure thx for the help.

laverya (approved):

camilamacedo86:
Hi @laverya,

You approved this one,  so I understand that at the end you are OK with the changes in the text and did not wished came up with wording suggestion (https://github.com/replicatedhq/kURL/pull/3973#discussion_r1084366505). I am moving forward, but feel free to let me us me if you came with a better message. We can update it always.

PR #3984, opened 2023-01-24
feat(flannel): log warning flannel udp port 8472 must be open

emosbaugh:

In order to make Flannel add-on GA, preflights must be added where necessary.

From my investigation it is not possible to check if a connection is possible over a UDP port from the client. In the absence of a preflight check I thought it best to add a warning.

Fixes NONE

```release-note
Adds a warning message when Flannel is the cluster CNI suggesting the user check that UDP port 8472 is open when joining a node or migrating from Weave to Flannel.
```

NONE

### DIFF
--- a/addons/flannel/0.20.2/install.sh
+++ b/addons/flannel/0.20.2/install.sh
@@ -26,6 +26,11 @@ function flannel_pre_init() {
     flannel_init_pod_subnet
 }
 
+function flannel_join() {
+    logWarn "Flannel requires UDP port 8472 for communication between nodes."
+    logWarn "Failure to open this port will cause connection failures between containers on different nodes."
+}
+
 function flannel() {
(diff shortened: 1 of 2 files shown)


addons/flannel/0.20.2/install.sh
```diff
@@ -26,6 +26,11 @@ function flannel_pre_init() {
     flannel_init_pod_subnet
 }
 
+function flannel_join() {
+    logWarn "Flannel requires UDP port 8472 for communication between nodes."
+    logWarn "Failure to open this port will cause connection failures between containers on different nodes."
```
emosbaugh, line 31: This language is meant to suggest that the user make sure that the port is open. Does this wording make it seem that the port is indeed not open? If so please suggest a copy edit.

laverya (approved):

PR #3988, opened 2023-01-25
chore: standardize minio deployment await timeout.

ricardomaraschini:

In some cases testgrid is failing to bring minio up, the failure reported is:

```
2022-12-28 07:26:30+00:00 awaiting minio deployment
2022-12-28 07:29:00+00:00 [/] An error occurred on line 199
```

By inspecting the testgrid log the pod seems to be running but not in a healthy state yet:

```
NAMESPACE     NAME                    STATUS    RESTARTS   AGE
minio         minio-5b8df4f997-vhhws  Running   0          2m31s
```

Worthy to mention that in the test in case the host preflight had a warning regarding disk performance (that may be related). As this problem does not happen consistently (the same test passed in the next daily run) this commit changes the timeout applied when waiting for Minio to come up.

Fixes #sc-65193

### DIFF
--- a/addons/minio/2022-07-06T20-29-49Z/install.sh
+++ b/addons/minio/2022-07-06T20-29-49Z/install.sh
@@ -153,7 +153,7 @@ function minio_object_store_output() {
 
 function minio_wait_for_health() {
     printf "awaiting minio deployment\n"
-    spinner_until 120 deployment_fully_updated minio minio
+    spinner_until 300 deployment_fully_updated minio minio
 
     MINIO_CLUSTER_IP=$(kubectl -n ${MINIO_NAMESPACE} get service minio | tail -n1 | awk '{ print $3}')
     printf "awaiting minio readiness\n"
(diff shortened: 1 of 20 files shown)


camilamacedo86 (approved):
Just a nit : https://github.com/replicatedhq/kURL/pull/3988/files#r1086568106

Otherwise, /lgtm

addons/openebs/2.6.0/install.sh
```diff
@@ -319,7 +319,7 @@ function openebs_spc_cspc_migration() {
 
 function openebs_cspc_upgrade() {
     # start upgrade process - see https://github.com/openebs/upgrade/blob/v2.6.0/docs/upgrade.md
-    kubectl delete csidriver cstor.csi.openebs.io || true
+    kubectl delete csidriver cstor.csi.openebs.io --ignore-not-found=true || true
```
camilamacedo86, line 322: By using the  --ignore-not-found=true  I think we do not need the || true anymore @ricardomaraschini

PR #3990, opened 2023-01-25
feat(ci): Speedup S3 Copy when Releasing to Prod

rrpolanco:

The `copy-staging-release-to-prod` GH workflow job is optimized to now perform multiple parallel s3api `copy-object` commands. I decided to use [GNU Parallel](https://www.gnu.org/software/parallel/sphinx.html) as the job orchestrator. The goal is to explicitly leverage all the CPU cores on the Linux Github runners to run multiple s3 copy operations so that we can release faster.

Fixes [SC-66950](https://app.shortcut.com/replicated/story/66950/speed-up-kurl-release-process)

I tested `bin/copy-packages-s3-parallel.sh` on an Amazon Linux VM via AWS CloudShell.

```release-note

```

### DIFF
--- a/bin/copy-packages-s3-parallel.sh
+++ b/bin/copy-packages-s3-parallel.sh
@@ -40,7 +40,7 @@ function retry {
     done
     return 0
 }
-# export the retry function so that the parallel jobs can use it
+# export the retry function so that the parallel command can use it
 export -f retry
 
 require AWS_ACCESS_KEY_ID "${AWS_ACCESS_KEY_ID}"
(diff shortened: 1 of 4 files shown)


bin/copy-packages-s3-parallel.sh
```diff
@@ -0,0 +1,103 @@
...
+
+  log "The following aws s3api commands will be executed in parallel:"
+  cat "${JOBS_FILE}"
+
+  # run s3 uploads in parallel
+  # --wil-cite => don't display citation nag
+  # --hatl now,fail=1 => running jobs will be killed immediately if any other job fails
+  # --jobs 0 => will run as many jobs in parallel as possible (depends on number of cores)
+  # --eta => show progress
+  parallel --will-cite --eta --halt now,fail=1 --jobs "${PARALLEL_JOBS}" < "${JOBS_FILE}"
```
emosbaugh, line 97: i'd prefer to parallelize this using actions jobs. that way we can leverage re-run failed.

rrpolanco: That's a good point to consider. However, I don't expect s3 copy-object operations to fail frequent enough to restart the job.

rrpolanco:
Closing in favor of https://github.com/replicatedhq/kURL/pull/3993

PR #3999, opened 2023-01-27
fix check to verify if multiple rook versions are found

camilamacedo86:

Note that when we are upgrading to 1.16.11 we are facing the error that multiple versions were found when they were not. We just fail in the timeout to check rook status deployments. See:

Screenshot 2023-01-27 at 12 51 21

Therefore, this PR fix the check for we do not warning something that is not true.

PS.: Also, note that in this scenario we can move forward and finish the upgrade with success. (It seems that we have some know issues in rock operator which gets stuck see: https://github.com/rook/rook/issues/5090 and because of this I am assuming that we do not fail. Anyway, the scope of this PR is just sorted out the check in order to not provide an wrong info and ensure that all places are following the same wording standards)

Fixes # [sc-67863]

## Steps to reproduce
Test upgrade from 1.15.12 to 1.6.11

```release-note
fix: check to verify multiple rook versions in for rook upgrade 
```

### DIFF
--- a/addons/rook/1.7.11/install.sh
+++ b/addons/rook/1.7.11/install.sh
@@ -237,9 +237,17 @@ function rook_cluster_deploy_upgrade() {
 
     # 4. https://rook.io/docs/rook/v1.6/ceph-upgrade.html#4-wait-for-the-upgrade-to-complete
     echo "Awaiting rook-ceph operator"
-    if ! $DIR/bin/kurl rook wait-for-rook-version "$ROOK_VERSION" --timeout=1200 ; then
-        logWarn "Rook version not yet rolled out"
+    if ! "$DIR"/bin/kurl rook wait-for-rook-version "$ROOK_VERSION" --timeout=1200 ; then
+        logWarn "Timeout waiting for Rook version rolled out"
+        logStep "Checking Rook versions and replicas"
         kubectl -n rook-ceph get deployment -l rook_cluster=rook-ceph -o jsonpath='{range .items[*]}{.metadata.name}{"  \treq/upd/avl: "}{.spec.replicas}{"/"}{.status.updatedReplicas}{"/"}{.status.readyReplicas}{"  \trook-version="}{.metadata.labels.rook-version}{"\n"}{end}'
+        local rook_versions=
+        rook_versions="$(kubectl -n rook-ceph get deployment -l rook_cluster=rook-ceph -o jsonpath='{range .items[*]}{"rook-version="}{.metadata.labels.rook-version}{"\n"}{end}' | sort | uniq)"
+        if [ -n "${rook_versions}" ] && [ "$(echo "${rook_versions}" | wc -l)" -gt "1" ]; then
+            logWarn "Detected multiple Rook versions"
+            logWarn "${rook_versions}"
+            logWarn "Failed to verify the Rook upgrade, multiple Rook versions detected"
+        fi
     fi
 
     # 5. https://rook.io/docs/rook/v1.6/ceph-upgrade.html#5-verify-the-updated-cluster
--- a/addons/rookupgrade/10to14/install.sh
+++ b/addons/rookupgrade/10to14/install.sh
@@ -101,7 +101,7 @@ function rookupgrade_10to14_upgrade() {
             if [ -n "${rook_versions}" ] && [ "$(echo "${rook_versions}" | wc -l)" -gt "1" ]; then
                 logWarn "Detected multiple Rook versions"
                 logWarn "${rook_versions}"
-                logWarn "Failed to verify the Rook upgrade, multiple Rook versions detected"
+                bail "Failed to verify the Rook upgrade, multiple Rook versions detected"
             fi
         fi
 
@@ -159,7 +159,7 @@ function rookupgrade_10to14_upgrade() {
             if [ -n "${rook_versions}" ] && [ "$(echo "${rook_versions}" | wc -l)" -gt "1" ]; then
                 logWarn "Detected multiple Rook versions"
                 logWarn "${rook_versions}"
-                logWarn "Failed to verify the Rook upgrade, multiple Rook versions detected"
+                bail "Failed to verify the Rook upgrade, multiple Rook versions detected"
             fi
         fi
 
@@ -203,7 +203,7 @@ function rookupgrade_10to14_upgrade() {
             if [ -n "${rook_versions}" ] && [ "$(echo "${rook_versions}" | wc -l)" -gt "1" ]; then
                 logWarn "Detected multiple Rook versions"
                 logWarn "${rook_versions}"
-                logWarn "Failed to verify the Rook upgrade, multiple Rook versions detected"
+                bail "Failed to verify the Rook upgrade, multiple Rook versions detected"
             fi
         fi
 
@@ -240,7 +240,7 @@ function rookupgrade_10to14_upgrade() {
             if [ -n "${rook_versions}" ] && [ "$(echo "${rook_versions}" | wc -l)" -gt "1" ]; then
                 logWarn "Detected multiple Rook versions"
                 logWarn "${rook_versions}"
-                logWarn "Failed to verify the Rook upgrade, multiple Rook versions detected"
+                bail "Failed to verify the Rook upgrade, multiple Rook versions detected"
             fi
         fi
 
--- a/addons/rookupgrade/10to14/install.sh
+++ b/addons/rookupgrade/10to14/install.sh
@@ -93,9 +93,16 @@ function rookupgrade_10to14_upgrade() {
 
         log "Waiting for Rook 1.1.9 to rollout throughout the cluster, this may take some time"
         if ! "$DIR"/bin/kurl rook wait-for-rook-version "v1.1.9" --timeout=1200 ; then
-            logWarn "Detected multiple Rook versions"
+            logWarn "Timeout waiting for Rook version 1.1.9 rolled out"
+            logStep "Checking Rook versions and replicas"
             kubectl -n rook-ceph get deployment -l rook_cluster=rook-ceph -o jsonpath='{range .items[*]}{.metadata.name}{"  \treq/upd/avl: "}{.spec.replicas}{"/"}{.status.updatedReplicas}{"/"}{.status.readyReplicas}{"  \trook-version="}{.metadata.labels.rook-version}{"\n"}{end}'
-            bail "Failed to verify the Rook upgrade, multiple Rook versions detected"
+            local rook_versions=
+            rook_versions="$(kubectl -n rook-ceph get deployment -l rook_cluster=rook-ceph -o jsonpath='{range .items[*]}{"rook-version="}{.metadata.labels.rook-version}{"\n"}{end}' | sort | uniq)"
+            if [ -n "${rook_versions}" ] && [ "$(echo "${rook_versions}" | wc -l)" -gt "1" ]; then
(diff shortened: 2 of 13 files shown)


emosbaugh (changes requested):
The issue is not that the version didnt roll out, its that the ceph fs pod never became ready. The correct change here would be to change the messaging similar to this

https://github.com/replicatedhq/kURL/blob/e313c3db4ab77dbb027f9a80759f59fd6544d60e/addons/rook/template/base/install.sh#L270-L271

emosbaugh (commented):
I'm not sure what this pull request is attempting to accomplish.

The existing logic will warn and continue if all rook deployments are not using the new version and ready.

They new logic will warn if not using the new version and ready and bail if not all deployments are using the new version.

camilamacedo86:
Hi @emosbaugh,

Regards https://github.com/replicatedhq/kURL/pull/3999#pullrequestreview-1273364067.
If the deploy and the health check does not work after we try to upgrade and we found 2 rook versions it probably fails and it is the root cause here. However, I totally agree with you.

Began to bail when we are not might introduce a new behaviour and some workflow that works currently might no longer work. I changed that. See that we are not using the bail when before we just warned anymore. please feel free to review this one, I think it is good to fly.

addons/rook/1.7.11/install.sh
```diff
@@ -237,9 +237,16 @@ function rook_cluster_deploy_upgrade() {
 
     # 4. https://rook.io/docs/rook/v1.6/ceph-upgrade.html#4-wait-for-the-upgrade-to-complete
     echo "Awaiting rook-ceph operator"
-    if ! $DIR/bin/kurl rook wait-for-rook-version "$ROOK_VERSION" --timeout=1200 ; then
-        logWarn "Rook version not yet rolled out"
-        kubectl -n rook-ceph get deployment -l rook_cluster=rook-ceph -o jsonpath='{range .items[*]}{.metadata.name}{"  \treq/upd/avl: "}{.spec.replicas}{"/"}{.status.updatedReplicas}{"/"}{.status.readyReplicas}{"  \trook-version="}{.metadata.labels.rook-version}{"\n"}{end}'
```
emosbaugh, line 242: in my opinion this was correct the way it was. The version is not yet rolled out because there is either multiple rook versions or the replicas are not ready and up to date which is the same check that the `kurl rook wait-for-rook-version` validates. The proposed change will hide the output if all versions are the same but the replicas are not ready and up to date. This seems like a regression.

The correct fix in my opinion would be to backport this code to versions 1.4.9 - 1.6.11

emosbaugh (changes requested):

addons/rook/1.7.11/install.sh
camilamacedo86: Hi @emosbaugh,

I think I understand your comment. It is helpful we get the output from the rook replicas and versions every time that we face an timeout. We should not only WARNING saying that we found more than one version when not more the one version was found.

addons/rookupgrade/10to14/install.sh
```diff
@@ -93,9 +93,16 @@ function rookupgrade_10to14_upgrade() {
 
         log "Waiting for Rook 1.1.9 to rollout throughout the cluster, this may take some time"
         if ! "$DIR"/bin/kurl rook wait-for-rook-version "v1.1.9" --timeout=1200 ; then
-            logWarn "Detected multiple Rook versions"
+            logWarn "Timeout waiting for Rook version 1.1.9 rolled out"
+            logStep "Checking Rook versions and replicas"
             kubectl -n rook-ceph get deployment -l rook_cluster=rook-ceph -o jsonpath='{range .items[*]}{.metadata.name}{"  \treq/upd/avl: "}{.spec.replicas}{"/"}{.status.updatedReplicas}{"/"}{.status.readyReplicas}{"  \trook-version="}{.metadata.labels.rook-version}{"\n"}{end}'
-            bail "Failed to verify the Rook upgrade, multiple Rook versions detected"
```
emosbaugh, line 98: this is a functional change rather than only messaging. this should bail as it did before.

addons/rookupgrade/10to14/install.sh
```diff
@@ -144,9 +151,16 @@ function rookupgrade_10to14_upgrade() {
 
         log "Waiting for Rook 1.2.7 to rollout throughout the cluster, this may take some time"
         if ! "$DIR"/bin/kurl rook wait-for-rook-version "v1.2.7" --timeout=1200 ; then
-            logWarn "Detected multiple Rook versions"
+            logWarn "Timeout waiting for Rook version 1.2.7 rolled out"
+            logStep "Checking Rook versions and replicas"
             kubectl -n rook-ceph get deployment -l rook_cluster=rook-ceph -o jsonpath='{range .items[*]}{.metadata.name}{"  \treq/upd/avl: "}{.spec.replicas}{"/"}{.status.updatedReplicas}{"/"}{.status.readyReplicas}{"  \trook-version="}{.metadata.labels.rook-version}{"\n"}{end}'
-            bail "Failed to verify the Rook upgrade, multiple Rook versions detected"
```
emosbaugh, line 149: this is a functional change rather than only messaging. this should bail as it did before.

addons/rookupgrade/10to14/install.sh
```diff
@@ -211,9 +232,16 @@ function rookupgrade_10to14_upgrade() {
 
         log "Waiting for Rook 1.4.9 to rollout throughout the cluster, this may take some time"
         if ! "$DIR"/bin/kurl rook wait-for-rook-version "v1.4.9" --timeout=1200 ; then
-            logWarn "Detected multiple Rook versions"
+            logWarn "Timeout waiting for Rook version 1.4.9 rolled out"
+            logStep "Checking Rook versions and replicas"
             kubectl -n rook-ceph get deployment -l rook_cluster=rook-ceph -o jsonpath='{range .items[*]}{.metadata.name}{"  \treq/upd/avl: "}{.spec.replicas}{"/"}{.status.updatedReplicas}{"/"}{.status.readyReplicas}{"  \trook-version="}{.metadata.labels.rook-version}{"\n"}{end}'
-            bail "Failed to verify the Rook upgrade, multiple Rook versions detected"
```
emosbaugh, line 216: this is a functional change rather than only messaging. this should bail as it did before.

emosbaugh (changes requested):

addons/rookupgrade/10to14/install.sh
```diff
@@ -93,9 +93,16 @@ function rookupgrade_10to14_upgrade() {
...
+            local rook_versions=
+            rook_versions="$(kubectl -n rook-ceph get deployment -l rook_cluster=rook-ceph -o jsonpath='{range .items[*]}{"rook-version="}{.metadata.labels.rook-version}{"\n"}{end}' | sort | uniq)"
+            if [ -n "${rook_versions}" ] && [ "$(echo "${rook_versions}" | wc -l)" -gt "1" ]; then
+                logWarn "Detected multiple Rook versions"
+                logWarn "${rook_versions}"
+                bail "Failed to verify the Rook upgrade, multiple Rook versions detected"
+            fi
```
emosbaugh, line 105: this should be moved outside this conditional

```suggestion
            fi
            bail "Failed to verify the Rook upgrade"
```

addons/rookupgrade/10to14/install.sh
```diff
@@ -144,9 +151,16 @@ function rookupgrade_10to14_upgrade() {
...
+            logWarn "Timeout waiting for Rook version 1.2.7 rolled out"
+            logStep "Checking Rook versions and replicas"
             kubectl -n rook-ceph get deployment -l rook_cluster=rook-ceph -o jsonpath='{range .items[*]}{.metadata.name}{"  \treq/upd/avl: "}{.spec.replicas}{"/"}{.status.updatedReplicas}{"/"}{.status.readyReplicas}{"  \trook-version="}{.metadata.labels.rook-version}{"\n"}{end}'
-            bail "Failed to verify the Rook upgrade, multiple Rook versions detected"
+            local rook_versions=
+            rook_versions="$(kubectl -n rook-ceph get deployment -l rook_cluster=rook-ceph -o jsonpath='{range .items[*]}{"rook-version="}{.metadata.labels.rook-version}{"\n"}{end}' | sort | uniq)"
+            if [ -n "${rook_versions}" ] && [ "$(echo "${rook_versions}" | wc -l)" -gt "1" ]; then
+                logWarn "Detected multiple Rook versions"
+                logWarn "${rook_versions}"
+                bail "Failed to verify the Rook upgrade, multiple Rook versions detected"
```
emosbaugh, line 162: this should be moved outside this conditional

addons/rookupgrade/10to14/install.sh
```diff
@@ -181,9 +195,16 @@ function rookupgrade_10to14_upgrade() {
...
+            logWarn "Timeout waiting for Rook version 1.3.11 rolled out"
+            logStep "Checking Rook versions and replicas"
             kubectl -n rook-ceph get deployment -l rook_cluster=rook-ceph -o jsonpath='{range .items[*]}{.metadata.name}{"  \treq/upd/avl: "}{.spec.replicas}{"/"}{.status.updatedReplicas}{"/"}{.status.readyReplicas}{"  \trook-version="}{.metadata.labels.rook-version}{"\n"}{end}'
-            bail "Failed to verify the Rook upgrade, multiple Rook versions detected"
+            local rook_versions=
+            rook_versions="$(kubectl -n rook-ceph get deployment -l rook_cluster=rook-ceph -o jsonpath='{range .items[*]}{"rook-version="}{.metadata.labels.rook-version}{"\n"}{end}' | sort | uniq)"
+            if [ -n "${rook_versions}" ] && [ "$(echo "${rook_versions}" | wc -l)" -gt "1" ]; then
+                logWarn "Detected multiple Rook versions"
+                logWarn "${rook_versions}"
+                bail "Failed to verify the Rook upgrade, multiple Rook versions detected"
```
emosbaugh, line 206: this should be moved outside this conditional

addons/rookupgrade/10to14/install.sh
```diff
@@ -211,9 +232,16 @@ function rookupgrade_10to14_upgrade() {
...
+            logWarn "Timeout waiting for Rook version 1.4.9 rolled out"
+            logStep "Checking Rook versions and replicas"
             kubectl -n rook-ceph get deployment -l rook_cluster=rook-ceph -o jsonpath='{range .items[*]}{.metadata.name}{"  \treq/upd/avl: "}{.spec.replicas}{"/"}{.status.updatedReplicas}{"/"}{.status.readyReplicas}{"  \trook-version="}{.metadata.labels.rook-version}{"\n"}{end}'
-            bail "Failed to verify the Rook upgrade, multiple Rook versions detected"
+            local rook_versions=
+            rook_versions="$(kubectl -n rook-ceph get deployment -l rook_cluster=rook-ceph -o jsonpath='{range .items[*]}{"rook-version="}{.metadata.labels.rook-version}{"\n"}{end}' | sort | uniq)"
+            if [ -n "${rook_versions}" ] && [ "$(echo "${rook_versions}" | wc -l)" -gt "1" ]; then
+                logWarn "Detected multiple Rook versions"
+                logWarn "${rook_versions}"
+                bail "Failed to verify the Rook upgrade, multiple Rook versions detected"
```
emosbaugh, line 243: this should be moved outside this conditional

emosbaugh (changes requested):

emosbaugh (approved):

PR #4021, opened 2023-02-01
feat: migrating from longhorn to openebs or rook

ricardomaraschini:

Scale down pods mounting Longhorn volumes during OpenEBS and Rook pre-init. This is done so we can start scaling things down before we start to migrate the cluster, we have noticed problems with Longhorn stability later on during the process.

Fixes #65977


emosbaugh (approved):

ricardomaraschini:
I will be reopening this.

PR #4025, opened 2023-02-02
Clean up Rook packages after upgrades

laverya:

This frees up disk space by deleting the packages for past rook versions as part of the upgrade process. When upgrading from Rook 1.0 to 1.10, this frees up over 20GB that would have otherwise been persisted.

~This also stores current and previous addon docker images in a pair of in-cluster configmaps. This data is currently unused, but is planned to be used to prune images after upgrading addons.~ EDIT: moved to another branch

Fixes #

```release-note

```


camilamacedo86 (changes requested):
HI @laverya,

That is nice but I have a few questions: 

You are here:
- 1. Creating a configmap to store the old images
- 2. Removing the assets from the filesystem

However:
- 1. Where and when we would use the config map?
- 2. The images were load to containerd or docker previously and we are not removing them?
- 3. After remove the assets and the images previously loaded when/where the configmap will be useful? Therefore, why do we need to keep it? 

Why not asap after the upgrade be made remove the filesystem and leaded images from the previous version installed (upgrade from)?

scripts/common/addon.sh
```diff
@@ -307,3 +309,49 @@ function addon_source() {
...
+# addon_update_used_images takes an addon name and version as parameters.
+# optionally, a third 'pathoverride' parameter can be passed to specify a path to the addon's images file.
+# it checks the kurl-current-images configmap for a match, and adds those images to kurl-old-images.
+# it then adds the new addon version's images to the kurl-current-images map and removes previous addon version's images.
+function addon_update_used_images() {
+    local name=$1
+    local version=$2
+    local pathoverride=$3
+
+    if ! kubectl get configmap -n kurl kurl-current-images; then
```
emosbaugh, line 322: can you use `--ignore-not-found` or `kubernetes_resource_exists` function?

scripts/common/addon.sh
```diff
@@ -307,3 +309,49 @@ function addon_source() {
...
+        kubectl create configmap -n kurl kurl-current-images
+    fi
+    if ! kubectl get configmap -n kurl kurl-old-images; then
```
emosbaugh, line 325: can you use `--ignore-not-found` or `kubernetes_resource_exists` function?

scripts/common/addon.sh
```diff
@@ -307,3 +309,49 @@ function addon_source() {
...
+        kubectl create configmap -n kurl kurl-old-images
+    fi
+
+    # update kurl-old-images configmap with images from the previous addon version
```
emosbaugh, line 329: What if you were to store a list of images with the key `[addon]-[version]` and another key `[addon]` with value `[version]` to look up the previous version? Wouldn't that greatly simplify the logic in this function? This is only a suggestion.

laverya: It might simplify the logic here, but I'm not sure it would simplify the logic overall.

That being said, I'm not sure if storing the list of images is even _useful_, so...

laverya: (Because there's not really a way to delete a host containerd image from within a k8s pod)

I think we might just want to rely on the k8s automatic image eviction here, but there's always the risk it'll prune an image we're migrating to in airgap. Another reason to use a registry)

emosbaugh: if we'd rather solve images holistically im fine with that

emosbaugh (approved):

scripts/common/rook-upgrade.sh
```diff
@@ -141,6 +141,10 @@ function rook_upgrade() {
     if rook_upgrade_is_version_included "$from_version" "$to_version" "1.4" ; then
         addon_source "rookupgrade" "10to14"
         rookupgrade_10to14_upgrade "$from_version"
+
+        # delete both the compressed and decompressed addon files to free up space
+        rm "$DIR/assets/rookupgrade-10to14.tar.gz"
```
emosbaugh, line 146: i dont think this will work for airgap since we use multi-package bundles, you should probably add a -f...

scripts/common/rook-upgrade.sh
```diff
@@ -184,6 +188,13 @@ function rook_upgrade_do_rook_upgrade() {
         # NOTE: there is no way to know this is the correct rook version function
         rook # upgrade to the step version
         ROOK_VERSION="$old_rook_version"
+
+        # if this is not the last version in the loop, then delete the addon files to free up space
+        if ! [[ "$step" =~ $to_version ]]; then
+            rm "$DIR/assets/rook-$step.tar.gz"
```
emosbaugh, line 194: same

emosbaugh (changes requested):

scripts/common/rook-upgrade.sh
laverya: Don't the multi-package bundles still decompress to these bundles?

emosbaugh: No they decompress directly to the file tree, not another archive

laverya: ouch, ok

emosbaugh (approved):

scripts/common/rook-upgrade.sh
```diff
@@ -179,8 +183,20 @@ function rook_upgrade_do_rook_upgrade() {
...
+        # if this is not the last version in the loop, then delete the addon files to free up space
+        if ! [[ "$step" =~ $to_version ]]; then
+            rm -f "$DIR/assets/rook-$step.tar.gz"
+            rm -rf "$DIR/addons/rook/$step"
+        fi
+
         logSuccess "Upgraded to Rook $step successfully"
     done <<< "$(rook_upgrade_step_versions "ROOK_STEP_VERSIONS[@]" "$from_version" "$to_version")"
+
+    if [ -z $AIRGAP_MULTI_ADDON_PACKAGE_PATH ]; then
```
emosbaugh, line 196: This should be -n

emosbaugh (changes requested):

scripts/common/rook-upgrade.sh
laverya: oh for...

emosbaugh (approved):

camilamacedo86 (approved):

PR #4027, opened 2023-02-02
feat: add prompt when preflights warnings are raised

camilamacedo86:

Currently, users are not requested for confirmation before moving on with the installation if the host pre-flight has warnings. As we don't ask for confirmation we don't give the user the option of aborting the install and fix the warning' root cause before resuming.

Therefore, this PR will let them be aware of and give this oppritunity.

Fixes # [sc-66713]

See:

Screenshot 2023-02-03 at 16 17 59

## Steps to reproduce
Create a VM  which will raise:

[WARN] Number of CPUs: At least 4 CPU cores are recommended
[WARN] Amount of Memory: At least 8G of memory is recommended

```release-note
- Adds a prompt when preflight warnings are faced to ensure that user is aware and given the opportunity to cancel the install to and fix the root cause before resuming.
```

NONE

### DIFF
--- a/scripts/common/preflights.sh
+++ b/scripts/common/preflights.sh
@@ -583,6 +583,13 @@ function host_preflights() {
                 ;;  
             2)
                 logWarn "Host preflights have warnings"
+                logWarn "It is highly recommended to sort out the warning conditions before proceeding."
+                logWarn "Be aware that continuing with preflight warnings can result in failures."
+                log ""
+                logWarn "Would you like to continue?"
+                if ! confirmY ; then
+                    bail "The installation will not continue"
+                fi
                 return 0
                 ;;
             1)


scripts/common/preflights.sh
```diff
@@ -583,6 +583,13 @@ function host_preflights() {
                 ;;  
             2)
                 logWarn "Host preflights have warnings"
+                logWarn "It is highly recommended sort out the warnings conditions before continue."
+                logWarn "Be aware that continue with preflight warnings can result in failures."
+                log ""
+                logWarn "Would you like to continue? "
+                if ! confirmN ; then
+                    bail "The install will not continue"
+                fi
```
emosbaugh, line 592:
```suggestion
                logWarn "It is highly recommended to sort out the warning conditions before proceeding."
                logWarn "Be aware that continuing with preflight warnings can result in failures."
                log ""
                logWarn "Would you like to continue? "
                if ! confirmN ; then
                    bail "The installation will not continue"
                fi
```

emosbaugh (changes requested):

scripts/common/preflights.sh
```diff
@@ -583,6 +583,13 @@ function host_preflights() {
                 ;;  
             2)
                 logWarn "Host preflights have warnings"
+                logWarn "It is highly recommended sort out the warnings conditions before continue."
+                logWarn "Be aware that continue with preflight warnings can result in failures."
+                log ""
+                logWarn "Would you like to continue? "
+                if ! confirmN ; then
```
emosbaugh, line 590: I'd like to call out that this will now cause kURL installations to fail by default when run via scripts or in non-interactive terminals when warnings are encountered. We should run this by @kyleprice before making this change and call it out in the release notes. Otherwise we could take a less drastic approach and change this to `confirmY`. I think we should discuss this with the team either way and determine the desired change.

```release-note
- kURL installations will now fail by default for non-interactive terminals when preflight warnings are encountered.
```

camilamacedo86: That is a very good point !. IHMO we should not fail by default and I agree that we should use confirmY here, so that if it is not interactive then it will continue to work as usual:

```
        echo "N"
        logWarn "Automatically declining prompt, shell is not interactive"
        return 1
    fi
```

camilamacedo86: I pushed the change.

emosbaugh (approved):

PR #4030, opened 2023-02-03
bug: check if prometheus is installed before attempt to scale it.

ricardomaraschini:

If prometheus is not installed then we fail if we try to scale it up or down. This PR checks if "monitoring" namespace is present in the cluster, if not then just skips scale up and down processes.


ricardomaraschini:
I will be reopening this in the feature. Closing this as it is incomplete (does not contain the relevant shell script part).

PR #4032, opened 2023-02-03
feat: add utility commands for finding private interface

emosbaugh:

in support of https://github.com/replicatedhq/kURL/pull/4029

I'd like to merge this first so testgrid passes.


pkg/cli/util.go
```diff
@@ -0,0 +1,78 @@
...
+			return cli.GetViper().BindPFlags(cmd.PersistentFlags())
+		},
+		PreRunE: func(cmd *cobra.Command, args []string) error {
+			return cli.GetViper().BindPFlags(cmd.Flags())
+		},
+	}
+	return cmd
+}
+
+func newUtilIfaceFromIPCommand(cli CLI) *cobra.Command {
```
camilamacedo86, line 25:
```suggestion
func newUtilIfaceFromIPCommand(_ CLI) *cobra.Command {
```

To fix the lint issue

pkg/cli/util.go
```diff
@@ -0,0 +1,78 @@
...
+				cmd.Println(iface.Name)
+				return nil
+			}
+			return err
+		},
+	}
+	return cmd
+}
+
+func newUtilDefaultIfaceCommand(cli CLI) *cobra.Command {
```
camilamacedo86, line 54: same
```suggestion
func newUtilDefaultIfaceCommand(_ CLI) *cobra.Command {
```

pkg/netutils/iface.go
```diff
@@ -0,0 +1,127 @@
...
+)
+
+func GetDefaultGatewayInterface() (*net.Interface, error) {
+	routes, err := netlink.RouteList(nil, syscall.AF_INET)
+	if err != nil {
+		return nil, err
+	}
+
+	for _, route := range routes {
+		if route.Dst == nil || route.Dst.String() == "0.0.0.0/0" {
```
camilamacedo86, line 18: create a const

pkg/netutils/iface.go
```diff
@@ -0,0 +1,127 @@
...
+		err := ifaceIPv4AddrMatch(&iface, ip)
+		if err == nil {
+			return &iface, nil
+		}
+	}
+
+	return nil, errors.New("no interface with given IP address found")
+}
+
+func GetInterfaceByIPv6(ip net.IP) (*net.Interface, error) {
```
camilamacedo86, line 63: Since we are exporting this methods could we add a comment on top to document them?
Also, do we need to export? Do we want someone using kurl a lib re-using this code? if not, could we no longer export things that we do not want anyone using?

pkg/cli/util.go
```diff
@@ -0,0 +1,78 @@
...
+	cmd := &cobra.Command{
+		Use:   "default-gateway-iface",
+		Short: "Gets the default gateway interface name",
+		RunE: func(cmd *cobra.Command, args []string) error {
+			if ipv6 {
+				iface, err := netutils.GetDefaultV6GatewayInterface()
+				if err != nil {
+					return err
+				}
+				cmd.Println(iface.Name)
```
camilamacedo86, line 65: do we really want this outputs or are they for debug only?
Why use cmd.Println?

pkg/cli/util.go
```diff
@@ -0,0 +1,78 @@
...
+
+	"github.com/pkg/errors"
+	"github.com/replicatedhq/kurl/pkg/netutils"
+	"github.com/spf13/cobra"
+)
+
+func newUtilCommand(cli CLI) *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "util",
+		Short: "Utility commands",
```
camilamacedo86, line 14: My concern here is that anything could fit under this command
Is this feature something that is valid for other cases? if not could we have a flannel command and some subcommand to it like `flannel find-private-interface`? WDYT?

pkg/cli/util.go
emosbaugh: This is a cli. Output needs to be parseable so i dont think special characters are appropriate. to see how these are used see https://github.com/replicatedhq/kURL/pull/4029/files#diff-898fa6e3c189cf8168e2a7c50d5900083094a39b8ef1a28a9b4c82189d29abcaR28-R29

pkg/cli/util.go
emosbaugh: ive changed this to netutil

laverya (approved):

PR #4045, opened 2023-02-06
bug: migrating minio object storage (longhorn to rook migration).

ricardomaraschini:
When migrating to Rook in a cluster where Minio is running we need to migrate all object storage from the latter. This commit creates the function to migrate the data. This function will called from the Rook add-on installer (coming up in a next PR).

Also with this commit:

- Removes default storage class annotation once the data has been migrated.
- Only attempts to scale down and up Prometheus if it is installed.

### DIFF
--- a/scripts/common/object_store.sh
+++ b/scripts/common/object_store.sh
@@ -257,3 +257,146 @@ EOF
     printf "\n${GREEN}Object store migration completed successfully${NC}\n"
     report_addon_success "rook-ceph-to-minio" "v1"
 }
+
+function migrate_minio_to_rgw() {
+    local minio_ns="$MINIO_NAMESPACE"
+    if [ -z "$minio_ns" ]; then
+        minio_ns=minio
+    fi
+
+    if ! kubernetes_resource_exists $minio_ns deployment minio; then
+        return 0
+    fi
+
+    report_addon_start "minio-to-rook-ceph" "v1"
+
+    if kubernetes_resource_exists kurl deployment ekc-operator; then
+        kubectl -n kurl scale deploy ekc-operator --replicas=0
+        echo "Waiting for ekco pods to be removed"
+        spinner_until 120 ekco_pods_gone
+    fi
+
+    RGW_HOST="rook-ceph-rgw-rook-ceph-store.rook-ceph"
+    RGW_ACCESS_KEY_ID=$(kubectl -n rook-ceph get secret rook-ceph-object-user-rook-ceph-store-kurl -o yaml | grep AccessKey | head -1 | awk '{print $2}' | base64 --decode)
+    RGW_ACCESS_KEY_SECRET=$(kubectl -n rook-ceph get secret rook-ceph-object-user-rook-ceph-store-kurl -o yaml | grep SecretKey | head -1 | awk '{print $2}' | base64 --decode)
+    RGW_CLUSTER_IP=$(kubectl -n rook-ceph get service rook-ceph-rgw-rook-ceph-store | tail -n1 | awk '{ print $3}')
+
+    MINIO_HOST="minio.${minio_ns}"
+    MINIO_ACCESS_KEY_ID=$(kubectl -n ${minio_ns} get secret minio-credentials -ojsonpath='{ .data.MINIO_ACCESS_KEY }' | base64 --decode)
+    MINIO_ACCESS_KEY_SECRET=$(kubectl -n ${minio_ns} get secret minio-credentials -ojsonpath='{ .data.MINIO_SECRET_KEY }' | base64 --decode)
+
+    get_shared
+
+    kubectl delete pod sync-object-store --force --grace-period=0 &>/dev/null || true
+
+    cat <<EOF | kubectl apply -f -
+apiVersion: v1
+kind: Pod
+metadata:
+  name: sync-object-store
+  namespace: default
+spec:
+  restartPolicy: OnFailure
+  containers:
+  - name: sync-object-store
+    image: $KURL_UTIL_IMAGE
+    command:
+    - /usr/local/bin/kurl
+    - sync-object-store
+    - --source_host=$MINIO_HOST
+    - --source_access_key_id=$MINIO_ACCESS_KEY_ID
+    - --source_access_key_secret=$MINIO_ACCESS_KEY_SECRET
+    - --dest_host=$RGW_HOST
+    - --dest_access_key_id=$RGW_ACCESS_KEY_ID
+    - --dest_access_key_secret=$RGW_ACCESS_KEY_SECRET
+EOF
+
+    echo "Waiting up to 2 minutes for sync-object-store pod to start"
+    if ! spinner_until 120 kubernetes_pod_started sync-object-store default; then
+        bail "sync-object-store pod failed to start within 2 minutes"
+    fi
+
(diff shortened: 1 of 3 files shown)


scripts/common/object_store.sh
```diff
@@ -257,3 +257,146 @@ EOF
     printf "\n${GREEN}Object store migration completed successfully${NC}\n"
     report_addon_success "rook-ceph-to-minio" "v1"
 }
+
+function migrate_minio_to_rgw() {
```
ricardomaraschini, line 261: This is an almost equal copy of `migrate_rgw_to_minio` above.

laverya: I hate to ask you to rewrite this
would it be possible to make both of these into wrapper scripts that set things like "source/dest ns" "source/dest service" "source/dest access key ID" and then call the same underlying logic?

scripts/common/object_store.sh
```diff
@@ -257,3 +257,146 @@ EOF
...
+        if [ "$s3_url" = "http://${MINIO_HOST}" ]; then
+            kubectl -n velero patch backupstoragelocation default --type=merge -p "{\"spec\":{\"config\":{\"s3Url\":\"http://${RGW_HOST}\",\"publicUrl\":\"http://${rgw_ip}\"}}}"
+
+            while read -r resticrepo; do
+                oldResticIdentifier=$(kubectl -n velero get resticrepositories "$resticrepo" -ojsonpath="{ .spec.resticIdentifier }")
+                newResticIdentifier=$(echo "$oldResticIdentifier" | sed "s/${MINIO_HOST}/${RGW_HOST}/")
+                kubectl -n velero patch resticrepositories "$resticrepo" --type=merge -p "{\"spec\":{\"resticIdentifier\":\"${newResticIdentifier}\"}}"
+            done < <(kubectl -n velero get resticrepositories --selector=velero.io/storage-location=default --no-headers | awk '{ print $1 }')
+        else
+            echo "default backupstoragelocation was not minio, skipping"
```
emosbaugh, line 377:
```suggestion
            echo "default velero backupstoragelocation is not minio, skipping"
```

scripts/common/object_store.sh
```diff
@@ -257,3 +257,146 @@ EOF
...
+    fi
+
+    local rgw_ip=$($DIR/bin/kurl format-address "$RGW_CLUSTER_IP")
+    # Update registry to use rook
+    if kubernetes_resource_exists kurl configmap registry-config; then
+        echo "Updating registry to use rook"
+        kubectl -n kurl get configmap registry-config -ojsonpath='{ .data.config\.yml }' | sed "s/regionendpoint: http.*/regionendpoint: http:\/\/${rgw_ip}/" > config.yml
+        kubectl -n kurl delete configmap registry-config
+        kubectl -n kurl create configmap registry-config --from-file=config.yml=config.yml
+        rm config.yml
```
emosbaugh, line 354:
```suggestion
        echo "Updating registry to use rook"
        local temp_file=
        temp_file=$(mktemp --suffix=.yml)
        kubectl -n kurl get configmap registry-config -ojsonpath='{ .data.config\.yml }' | sed "s/regionendpoint: http.*/regionendpoint: http:\/\/${rgw_ip}/" > "$temp_file"
        kubectl -n kurl delete configmap registry-config
        kubectl -n kurl create configmap registry-config --from-file=config.yml="$temp_file"
        rm "$temp_file"
```

emosbaugh (changes requested):

scripts/common/object_store.sh
```diff
@@ -202,49 +198,49 @@ EOF
...
-            sed -i "s/aws_access_key_id=.*/aws_access_key_id=${MINIO_ACCESS_KEY_ID}/" cloud
-            sed -i "s/aws_secret_access_key=.*/aws_secret_access_key=${MINIO_ACCESS_KEY_SECRET}/" cloud
+            sed -i "s/aws_access_key_id=.*/aws_access_key_id=${destination_access_key}/" cloud
+            sed -i "s/aws_secret_access_key=.*/aws_secret_access_key=${destination_secret_key}/" cloud
             cloud=$(cat cloud | base64 -w 0)
             kubectl -n velero patch secret cloud-credentials -p "{\"data\":{\"cloud\":\"${cloud}\"}}"
             rm cloud
         else
-            echo "cloud-credentials secret were not for rgw, skipping"
+            echo "cloud-credentials secret were not for the source host, skipping"
```
emosbaugh, line 243: something is wrong about this sentence

scripts/common/object_store.sh
```diff
@@ -157,20 +154,19 @@ spec:
...
-    - --source_access_key_secret=$RGW_ACCESS_KEY_SECRET
-    - --dest_host=$MINIO_HOST
-    - --dest_access_key_id=$MINIO_ACCESS_KEY_ID
-    - --dest_access_key_secret=$MINIO_ACCESS_KEY_SECRET
+    - --source_host=$source_host
+    - --source_access_key_id=$source_access_key
+    - --source_access_key_secret=$source_secret_key
+    - --dest_host=$destination_host
+    - --dest_access_key_id=$destination_access_key
+    - --dest_access_key_secret=$destination_secret_key
```
ricardomaraschini, line 164: I think we already have a function that copies object stores [here](https://github.com/replicatedhq/kURL/blob/main/scripts/common/object_store.sh#L69-L72). Maybe we can replace this with a call to `migrate_object_store`.

laverya: They're not quiiiiiite the same, but I'll give it a go

laverya: easy enough

scripts/common/object_store.sh
```diff
@@ -204,49 +167,53 @@ EOF
...
-            kubectl -n velero get secret cloud-credentials -ojsonpath='{ .data.cloud }' | base64 -d > cloud
-            sed -i "s/aws_access_key_id=.*/aws_access_key_id=${MINIO_ACCESS_KEY_ID}/" cloud
-            sed -i "s/aws_secret_access_key=.*/aws_secret_access_key=${MINIO_ACCESS_KEY_SECRET}/" cloud
-            cloud=$(cat cloud | base64 -w 0)
+        if kubectl -n velero get secret cloud-credentials -ojsonpath='{ .data.cloud }' | base64 -d | grep -q "$source_access_key"; then
+            local temp_file=
+            temp_file=$(mktemp)
+            kubectl -n velero get secret cloud-credentials -ojsonpath='{ .data.cloud }' | base64 -d > "$temp_file"
+            sed -i "s/aws_access_key_id=.*/aws_access_key_id=${destination_access_key}/" cloud
+            sed -i "s/aws_secret_access_key=.*/aws_secret_access_key=${destination_secret_key}/" cloud
```
ricardomaraschini, line 211: Seems to me that we mixed the "cloud" and the new temp_file variable here, no ? I mean, shouldn't `cloud` be `$temp_file` in the previous two lines ?

laverya: yes
yes it should be

laverya, line 211:
```suggestion
            sed -i "s/aws_access_key_id=.*/aws_access_key_id=${destination_access_key}/" "$temp_file"
            sed -i "s/aws_secret_access_key=.*/aws_secret_access_key=${destination_secret_key}/" "$temp_file"
```

rrpolanco (approved):

PR #4047, opened 2023-02-06
Add restic restore helper image from gcr.io

sgalsaleh:

It seems like the default image name for the restic restore helper image in Velero v1.9.5 changed from `velero/velero-restic-restore-helper:v1.9.5` to `gcr.io/velero-gcp/velero-restic-restore-helper:v1.9.5` either by mistake or intentionally, we're still waiting on an answer from the Velero team. In the meantime, this works around the issue by including both images for airgapped installations.

Fixes #

~I did not make the same change in the `template` directory since I'm not sure if the next Velero version would have the same issue or not.~

```release-note
- Fixes an issue in [Velero add-on](https://kurl.sh/docs/add-ons/velero) version 1.9.5 where restores fail to pull the `velero-restic-restore-helper` image in air gapped environments.
```

NONE

### DIFF
--- a/addons/velero/1.9.5/Manifest
+++ b/addons/velero/1.9.5/Manifest
@@ -1,5 +1,6 @@
 image velero velero/velero:v1.9.5
 image restic-restore velero/velero-restic-restore-helper:v1.9.5
+image restic-restore-gcr gcr.io/velero-gcp/velero-restic-restore-helper:v1.9.5
 image velero-aws velero/velero-plugin-for-aws:v1.6.0
 image velero-gcp velero/velero-plugin-for-gcp:v1.6.0
 image velero-azure velero/velero-plugin-for-microsoft-azure:v1.6.0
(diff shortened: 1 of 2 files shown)


addons/velero/1.9.5/Manifest
```diff
@@ -1,5 +1,6 @@
 image velero velero/velero:v1.9.5
 image restic-restore velero/velero-restic-restore-helper:v1.9.5
+image restic-restore-gcr gcr.io/velero-gcp/velero-restic-restore-helper:v1.9.5
```
laverya, line 3: Could you add this to the velero template too?

laverya (approved):

laverya:
I'd still rather have this image added to template too, even if we don't know for sure that it'll be required yet.

Better to have too large a file than to accidentally break the next release too

rrpolanco (approved):

PR #4038, opened 2023-02-06
chore: verifying rook updated prior report success

camilamacedo86:
#### What this PR does / why we need it:
Before finish the upgrade and reported that it was done with success check the status as described in the docs from version 1.6.11+. That is required in order to:

- a) Not report that the upgrade was done successfully when it might not be
- b) avoid continue the upgrading for the next versions before check that the current upgrade was performed with success and avoid issues to the next ones. 

#### Which issue(s) this PR fixes:

Fixes # [sc-68491] 
Fixes # [sc-68490]

#### Special notes for your reviewer:

- Testgrid : https://testgrid.kurl.sh/run/rook_upgrade_feb_6_check_upgrades_with_rook_check
- Testgrid After Apply the suggestions: https://testgrid.kurl.sh/run/rook_upgrade_feb_6_check_upgrades_with_rook_check_with_suggestions

Example upgrade to 1.10.6 with this changes:

```
2023-02-06 15:33:56+00:00 ⚙  Addon rook 1.10.6
....
⚙  Checking if Rook Upgrade is completed successfully
2023-02-06 15:40:15+00:00 Verifying Rook Version Deployed
2023-02-06 15:40:16+00:00 Verifying Ceph Version 17.2.5 Deployed
2023-02-06 15:40:16+00:00 Verifying Ceph Status
2023-02-06 15:40:16+00:00 Waiting for Rook-Ceph to be healthy
2023-02-06 15:40:16+00:00 Rook is healthy
2023-02-06 15:40:16+00:00 ✔ Rook-ceph cluster upgraded successfully**
...
```

#### Does this PR introduce a user-facing change?

```release-note
Adds check to verify the updated cluster from rook upgrade version 1.6.11 prior report success and continue the workflow with multiple upgrades
```

#### Does this PR require documentation?
NONE

PR #4052, opened 2023-02-07
chore: replacing ceph version check in shell by golang command

camilamacedo86:

To address the request as a follow up: https://github.com/replicatedhq/kURL/pull/4038#discussion_r1097805802

Fixes # [sc-68491]

We executed the following tests after we get merged : 

https://testgrid.kurl.sh/run/rook_upgrade_feb_7_rook_upgrades_after_chages

Therefore, after we merge this changes we should run the same tests and ensure that we will get the same result. Otherwise, we might need to work on to fix the golang command used to check the ceph version.

### DIFF
--- a/addons/rook/template/base/install.sh
+++ b/addons/rook/template/base/install.sh
@@ -306,8 +306,8 @@ function rook_cluster_deploy_upgrade() {
 
     # https://rook.io/docs/rook/v1.6/ceph-upgrade.html#2-wait-for-the-daemon-pod-updates-to-complete
     if ! "$DIR"/bin/kurl rook wait-for-ceph-version "${ceph_version}-0" --timeout=1200 ; then
-        logWarn "Timeout waiting for Ceph version rolled out"
-        logStep "Checking Ceph versions and replicas"
+        logWarn "Timeout waiting for Ceph version to be rolled out"
+        log "Checking Ceph versions and replicas"
         kubectl -n rook-ceph get deployment -l rook_cluster=rook-ceph -o jsonpath='{range .items[*]}{.metadata.name}{"  \treq/upd/avl: "}{.spec.replicas}{"/"}{.status.updatedReplicas}{"/"}{.status.readyReplicas}{"  \tceph-version="}{.metadata.labels.ceph-version}{"\n"}{end}'
         local ceph_versions_found=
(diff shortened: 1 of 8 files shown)


addons/rook/template/base/install.sh
```diff
@@ -361,7 +361,21 @@ function verify_rook_updated_cluster() {
     fi
 
     log "Verifying Ceph version ${ceph_version} deployed"
-    if ! spinner_until 1200 rook_ceph_version_deployed "${ceph_version}" ; then
+    if ! "$DIR"/bin/kurl rook wait-for-ceph-version "${ceph_version}-0" --timeout=1200 ; then
+        logWarn "Timeout waiting for Ceph version rolled out"
```
emosbaugh, line 365:
```suggestion
        logWarn "Timeout waiting for Ceph version to be rolled out"
```

addons/rook/template/base/install.sh
```diff
@@ -361,7 +361,21 @@ function verify_rook_updated_cluster() {
...
+        logStep "Checking Ceph versions and replicas"
```
emosbaugh, line 366:
```suggestion
        log "Checking Ceph versions and replicas"
```

emosbaugh (changes requested):

addons/rook/template/base/install.sh
```diff
@@ -361,7 +361,21 @@ function verify_rook_updated_cluster() {
     fi
 
     log "Verifying Ceph version ${ceph_version} deployed"
-    if ! spinner_until 1200 rook_ceph_version_deployed "${ceph_version}" ; then
+    if ! "$DIR"/bin/kurl rook wait-for-ceph-version "${ceph_version}-0" --timeout=1200 ; then
```
rrpolanco, line 364: I'm curious why the argument to the command is `${cpeh_version}-0` with `-0` suffix.

rrpolanco (approved):

addons/rook/template/base/install.sh
camilamacedo86: So far, it seems to work and that ceph version added is contacted with `-0`.
However, it might not work with future versions

emosbaugh (approved):

PR #4053, opened 2023-02-07
chore: fix testgrid by updating it to the latest versions to cover rook upgrade workflow

camilamacedo86:

Currently this test using old versions is getting stuck.
However, we should no longer be testing with so older versions and by updating it to use the latest after the fix we can check that test can be performed with success which allows us to cover the workflow. 

This PR proposes 
- Update the current test to check rook upgrade with the latest versions. From rook 1.7.x to latest and k8s 1.24
- Add a new test to check rook upgrade from 1.4.x to 1.8.x using docker and old k8s release since this one we can check that pass in the tests. 

Fixes # [sc-68484]

See the testgrid: https://testgrid.kurl.sh/run/rook_upgrade_fix_test_feb_7_3?kurlLogsInstanceId=vcrtefrlgmsfftyy&nodeId=vcrtefrlgmsfftyy-initialprimary

### DIFF
--- a/testgrid/specs/full.yaml
+++ b/testgrid/specs/full.yaml
@@ -627,40 +627,66 @@
   airgap: true
   unsupportedOSIDs:
   - ubuntu-2204 # containerd 1.4.x is not supported on ubuntu 22.04
-- name: k8s1205_rook_upgrade
+- name: k8s24x_rook_upgrade_17x_latest
   flags: "yes"
   cpu: 6
   installerSpec:
     kubernetes:
-      version: 1.18.4
+      version: 1.24.x
+    weave:
+      version: 2.8.1
+    rook:
+      isBlockStorageEnabled: true
+      version: 1.7.x
+    kotsadm:
+      version: 1.93.x
+    containerd:
+      version: 1.6.x
+    ekco:
+      version: latest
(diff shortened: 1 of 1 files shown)


testgrid/specs/full.yaml
```diff
@@ -627,40 +627,39 @@
   airgap: true
   unsupportedOSIDs:
   - ubuntu-2204 # containerd 1.4.x is not supported on ubuntu 22.04
-- name: k8s1205_rook_upgrade
+- name: k8s24x_rook_upgrade_143_1108
```
camilamacedo86, line 630:
```suggestion
- name: k8s24x_rook_upgrade_16x_1108
```

testgrid/specs/full.yaml
```diff
@@ -627,40 +627,39 @@
...
   installerSpec:
     kubernetes:
-      version: 1.18.4
+      version: 1.24.x
     weave:
       version: 2.8.1
     rook:
       isBlockStorageEnabled: true
-      version: 1.4.3
+      version: 1.7.x
```
camilamacedo86, line 640:
```suggestion
      version: 1.6.x
```

emosbaugh (approved):

PR #4057, opened 2023-02-08
fix: start to delete flannel pods to workaround issue scenario

camilamacedo86:

To workaround the issue scenario: https://github.com/flannel-io/flannel/issues/1721 where the pods will get stuck to create the containers. It is sorted out by re-creating the flannel pods. 

Fixes # [sc-68370]

Testgrid: https://testgrid.kurl.sh/run/flannel_k8s_upgrade_feb_8_workaround
Tests with the latest changes (1c7db82) : https://testgrid.kurl.sh/run/rook_upgrade_flannel_feb_9_1c7db82

## Steps to reproduce

https://staging.kurl.sh/47129cf -> https://staging.kurl.sh/1a98e97

```release-note
Fixes issue scenario faced where the containers are not created after k8s upgrades because flannel issues. 
```

NONE

### DIFF
--- a/addons/flannel/template/base/install.sh
+++ b/addons/flannel/template/base/install.sh
@@ -256,9 +256,11 @@ function flannel_already_applied() {
     # We will remove the flannel pods to let it be re-created
     # in order to workaround the issue scenario described in
     # https://github.com/flannel-io/flannel/issues/1721
-    log "Deleting pods from kube-flannel namespace"
-    kubectl delete --all pods --namespace=kube-flannel
-    sleep 60
+    if [ "$KUBERNETES_UPGRADE" == "1" ]; then
+       log "Deleting pods from kube-flannel namespace"
+       kubectl delete --all pods --namespace=kube-flannel
+       sleep 60
+    fi
 
     flannel_ready_spinner
     check_network
--- a/addons/flannel/template/base/install.sh
+++ b/addons/flannel/template/base/install.sh
@@ -253,6 +253,13 @@ function flannel_kubeadm() {
 }
 
 function flannel_already_applied() {
+    # We will remove the flannel pods to let it be re-created
+    # in order to workaround the issue scenario described in
+    # https://github.com/flannel-io/flannel/issues/1721
+    log "Deleting pods from kube-flannel namespace"
+    kubectl delete --all pods --namespace=kube-flannel
+    sleep 60
+
     flannel_ready_spinner
     check_network
 }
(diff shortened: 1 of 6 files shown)


addons/flannel/template/base/install.sh
```diff
@@ -253,6 +253,13 @@ function flannel_kubeadm() {
 }
 
 function flannel_already_applied() {
+    # We will remove the flannel pods to let it be re-created
+    # in order to workaround the issue scenario described in
+    # https://github.com/flannel-io/flannel/issues/1721
+    log "Deleting pods from kube-flannel namespace"
+    kubectl delete --all pods --namespace=kube-flannel
```
emosbaugh, line 260: i dont think we should delete flannel pods each time we run the script. perhaps we should target this to only when we upgrade kubernetes? i cant think of any other way to do this than to put the code around here https://github.com/replicatedhq/kURL/blob/34feb2f47ef808dbee1d025825834e03dba0c7a9/scripts/common/kubernetes.sh#L183

emosbaugh (changes requested):

addons/flannel/template/base/install.sh
laverya: or have a 'did upgrade k8s host packages' env var

camilamacedo86: This code implementation will only be called when the package is installed already.
So, should not it only be called in the upgrade?

emosbaugh:
It will be called each time the Kurl script is run. Not only when
kubernetes is being upgraded. Also please take into account multi node
clusters. We will have to restart flannel after it has been upgraded on all
nodes

addons/flannel/template/base/install.sh
camilamacedo86: @emosbaugh,  

Follow up your suggestion https://github.com/replicatedhq/kURL/pull/4057#discussion_r1100677337 did not sorted out the problem. That is not right moment to do this action. 

HI @laverya,

I tried to do what you suggested https://github.com/replicatedhq/kURL/pull/4057#discussion_r1100679805 but not all tests passed as well. (https://testgrid.kurl.sh/run/flannel_k8s_upgrade_feb_9_test_envvar)

Honest, I do not see any problem with we delete the pods since they should be re-created expected it takes 1 minute more to run the script when it might not be required. if they not be re-created then, we have other issues in the install and it should fail in the health check. 

However, in order to address the request let's see if we can use the ENV VAR to know when is an upgrade and just run it when is an upgrade and has flannel installed. See: 5c2ba9f5805f443218c5c73a99f1a1f2b6479d6e

laverya (approved):

PR #4063, opened 2023-02-09
chore(Rook): Check for Rook pods in namespace to be in Running or Completed phase

rrpolanco:

After Rook is installed, ensure the Rook operator is ready and all pods in `rook-ceph` namespace are in `Running` phase before completing the install. This _could_ help with some failing tests where post-install/upgrade scripts fail due to the Rook/Ceph deployment still being rolled-out.

Fixes #

```release-note

```

### DIFF
--- a/addons/rook/template/base/install.sh
+++ b/addons/rook/template/base/install.sh
@@ -53,6 +53,10 @@ function rook_post_init() {
     if [ "$ROOK_DID_DISABLE_EKCO_OPERATOR" = "1" ]; then
         rook_enable_ekco_operator
     fi
+
+    # wait for all pods in the rook-ceph namespace to rollout
+    log "Rook Post-init: Awaiting Rook rollout in rook-ceph namespace"
+    rook_maybe_wait_for_rollout
 }
 
 ROOK_DID_DISABLE_EKCO_OPERATOR=0
@@ -987,4 +991,15 @@ function rook_prompt_migrate_from_longhorn() {
     if ! confirmN; then
         bail "Not migrating"
     fi
+}
+
+# wait for all pods in rook-ceph namespace to transition to Running status
+function rook_maybe_wait_for_rollout() {
+
+    # allow the Rook operator to start its reconcile loop
+    sleep 5
+
+    log "Awaiting Rook pods to transition to Running"
+    spinner_until 120 wait_for_running_pods "rook-ceph"
+
 }
--- a/addons/rook/template/base/install.sh
+++ b/addons/rook/template/base/install.sh
@@ -997,7 +997,7 @@ function rook_prompt_migrate_from_longhorn() {
 function rook_maybe_wait_for_rollout() {
 
     # allow the Rook operator to start its reconcile loop
-    sleep 5
+    sleep 10
 
     log "Awaiting Rook pods to transition to Running"
     spinner_until 120 wait_for_running_pods "rook-ceph"
--- a/addons/rook/template/base/install.sh
+++ b/addons/rook/template/base/install.sh
@@ -57,6 +57,8 @@ function rook_post_init() {
     # wait for all pods in the rook-ceph namespace to rollout
     log "Rook Post-init: Awaiting Rook rollout in rook-ceph namespace"
     rook_maybe_wait_for_rollout
+
+    log "Rook Post-init: End"
 }
 
 ROOK_DID_DISABLE_EKCO_OPERATOR=0
@@ -993,13 +995,15 @@ function rook_prompt_migrate_from_longhorn() {
     fi
 }
 
-# wait for all pods in rook-ceph namespace to transition to Running status
+# wait for Rook deployment to be running and healthy
 function rook_maybe_wait_for_rollout() {
-
-    # allow the Rook operator to start its reconcile loop
-    sleep 10
-
     log "Awaiting Rook pods to transition to Running"
-    spinner_until 120 wait_for_running_pods "rook-ceph"
+    if ! spinner_until 120 wait_for_running_pods "rook-ceph"; then
+        logWarn "Rook-ceph rollout did not complete"
+    fi
 
+    log "Awaiting for Ceph Cluster to be Ready"
+    if ! spinner_until 120 rook_cephcluster_ready; then
+        logWarn "Ceph cluster is not ready"
+    fi
 }
--- a/addons/rook/template/base/install.sh
+++ b/addons/rook/template/base/install.sh
@@ -53,12 +53,6 @@ function rook_post_init() {
     if [ "$ROOK_DID_DISABLE_EKCO_OPERATOR" = "1" ]; then
         rook_enable_ekco_operator
     fi
-
-    # wait for all pods in the rook-ceph namespace to rollout
-    log "Rook Post-init: Awaiting Rook rollout in rook-ceph namespace"
-    rook_maybe_wait_for_rollout
-
-    log "Rook Post-init: End"
 }
 
 ROOK_DID_DISABLE_EKCO_OPERATOR=0
@@ -992,18 +986,5 @@ function rook_prompt_migrate_from_longhorn() {
     log "Would you like to continue? "
     if ! confirmN; then
         bail "Not migrating"
-    fi
-}
-
-# wait for Rook deployment to be running and healthy
-function rook_maybe_wait_for_rollout() {
-    log "Awaiting Rook pods to transition to Running"
-    if ! spinner_until 120 wait_for_running_pods "rook-ceph"; then
-        logWarn "Rook-ceph rollout did not complete"
-    fi
-
-    log "Awaiting for Ceph Cluster to be Ready"
-    if ! spinner_until 120 rook_cephcluster_ready; then
-        logWarn "Ceph cluster is not ready"
     fi
 }
--- a/scripts/common/common.sh
+++ b/scripts/common/common.sh
@@ -1060,6 +1060,11 @@ function wait_for_running_pods() {
             is_job_controller=1
         fi
 
+        # ignore pods that have been Evicted
+        if [ "$status" == "Failed" ] && [[ $(kubectl get pod "$pod" -n "$namespace" -o jsonpath='{.status.reason}') == "Evicted" ]]; then
+            continue
+        fi
+
         if [ "$status" != "Running" ] && [ "$status" != "Succeeded" ]; then
             log "  Pod, $pod, is not ready: $status"
             return 1
--- a/scripts/common/common.sh
+++ b/scripts/common/common.sh
@@ -1037,8 +1037,9 @@ function canonical_image_name() {
     echo "$image"
 }
 
-# wait_for_running_pods waits for pod(s) in a given namspace to transition to Running status
-function wait_for_running_pods() {
+# check_for_running_pods scans for pod(s) in a namespace and checks whether their status is running/completed
+# note: Evicted pods are exempt from this check
+function check_for_running_pods() {
     local namespace=$1
     local is_job_controller=0
     local ns_pods=
@@ -1048,7 +1049,6 @@ function wait_for_running_pods() {
     ns_pods=$(kubectl get pods -n "$namespace" -o jsonpath='{.items[*].metadata.name}')
 
     if [ -z "$ns_pods" ]; then
-        log "No pods found in namespace $namespace."
         return 0
     fi
 
@@ -1066,7 +1066,6 @@ function wait_for_running_pods() {
         fi
 
         if [ "$status" != "Running" ] && [ "$status" != "Succeeded" ]; then
-            log "  Pod, $pod, is not ready: $status"
             return 1
         fi
 
@@ -1076,12 +1075,10 @@ function wait_for_running_pods() {
             
             # ignore container ready status for pods managed by the Job controller
             if [ "$container_status" != "true" ] && [ "$is_job_controller" = "0" ]; then
-                log "  Container, $container ($pod), is not ready: $container_status"
                 return 1
             fi
         done
     done
 
-    log "All pods and their containers in namespace $namespace are running."
     return 0
 }
--- a/scripts/common/common.sh
+++ b/scripts/common/common.sh
@@ -1070,7 +1070,7 @@ function wait_for_running_pods() {
             container_status=$(kubectl get pod "$pod" -n "$namespace" -o jsonpath="{.status.containerStatuses[?(@.name==\"$container\")].ready}")
             
             # ignore container ready status for pods managed by the Job controller
-            if [ "$container_status" != "true" ] && [ $is_job_controller -eq 0 ]; then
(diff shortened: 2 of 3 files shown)


scripts/common/common.sh
```diff
@@ -1036,3 +1036,54 @@ function canonical_image_name() {
...
+
+    for pod in $ns_pods; do
+        status=$(kubectl get pod "$pod" -n "$namespace" -o jsonpath='{.status.phase}')
+
+        # ignore pods spawned by Jobs
+        if kubectl get pod "$pod" -n "$namespace" -o jsonpath='{.metadata.ownerReferences[*].kind}' | grep -q "Job"; then
+            continue
+        fi
+
+        if [ "$status" != "Running" ]; then
```
camilamacedo86, line 1063: The provisioner pod will have status completed so that will fail always.

rrpolanco: Pods with status `Completed     0/1` are managed by the Job controller. I didn't think it was worth checking the status of the Job pods but I'm amenable to change my mind.

addons/rook/template/base/install.sh
```diff
@@ -988,3 +992,14 @@ function rook_prompt_migrate_from_longhorn() {
         bail "Not migrating"
     fi
 }
+
+# wait for all pods in rook-ceph namespace to transition to Running status
+function rook_maybe_wait_for_rollout() {
```
emosbaugh, line 997: i wonder if rather than waiting for the pods here we could wait for the conditions in the cephcluster crd?

https://rook.io/docs/rook/v1.10/CRDs/Cluster/ceph-cluster-crd/#conditions

emosbaugh (changes requested):

addons/rook/template/base/install.sh
camilamacedo86: Also, what about we add this check under: https://github.com/replicatedhq/kURL/blob/main/addons/rook/template/base/install.sh#L344

scripts/common/rook.sh
```diff
@@ -225,3 +225,17 @@ function current_ceph_version() {
...
         | awk -F'-' '{ print $1 }'
 }
+
+function rook_cephcluster_ready() {
+    local ceph_cluster_phase=
+    local ceph_cluster_status=
+    ceph_cluster_phase=$(kubectl -n rook-ceph get cephcluster rook-ceph --template '{{.status.phase}}')
+    if [  "$ceph_cluster_phase" = "Ready"  ]; then
+        ceph_cluster_status=$(kubectl -n rook-ceph get cephcluster rook-ceph --template '{{.status.ceph.health}}')
+        if [ "$CEPH_POOL_REPLICAS" -gt 1 ] && [ "$ceph_cluster_status" = "HEALTH_WARN" ]; then
```
camilamacedo86, line 235: We cannot to do that because we have many warns that we need to ignore.
For example, if we install kurl in a host that has only 1 master node we will see warnings about the pools. 

See we ignoring some warnings: https://github.com/replicatedhq/kURL/blob/main/pkg/rook/health.go#L66-L125

Also, see that we check the ceph status already: https://github.com/replicatedhq/kURL/blob/2f119ae38bc88a5ab4fc8ad1a6d14aa7cb674fbe/addons/rook/1.9.12/install.sh#L330-L335

camilamacedo86: https://github.com/replicatedhq/kURL/blob/2f119ae38bc88a5ab4fc8ad1a6d14aa7cb674fbe/addons/rook/1.9.12/install.sh#L307-L324

emosbaugh: As an alternative you can use Kurl rook wait-for-health command

camilamacedo86: > As an alternative you can use Kurl rook wait-for-health command

But we do those checks already. So, how could that sort out the problem? 
Rook creates a lot of pods, so are not those others that might be with ContainerCreating, etc. status? 
So, I think we just can ensure that all pods in the namespace are with status as completed and running.

emosbaugh: In my initial comment I had hoped that the cephcluster status.phase or conditions would be enough to check for rook ready, but it seems that the pod check is still required. if so you can forgo my request.

scripts/common/common.sh
```diff
@@ -1054,23 +1055,26 @@ function wait_for_running_pods() {
 
     for pod in $ns_pods; do
         status=$(kubectl get pod "$pod" -n "$namespace" -o jsonpath='{.status.phase}')
-
-        # ignore pods spawned by Jobs
+
```
emosbaugh, line 1058: can you remove this trailing whitespace?

addons/rook/template/base/install.sh
```diff
@@ -53,6 +53,12 @@ function rook_post_init() {
     if [ "$ROOK_DID_DISABLE_EKCO_OPERATOR" = "1" ]; then
         rook_enable_ekco_operator
     fi
+
+    # wait for all pods in the rook-ceph namespace to rollout
+    log "Rook Post-init: Awaiting Rook rollout in rook-ceph namespace"
+    rook_maybe_wait_for_rollout
```
emosbaugh: I dont think this belongs in the post_init. it affects subsequent addons.

addons/rook/template/base/install.sh
```diff
@@ -988,3 +994,16 @@ function rook_prompt_migrate_from_longhorn() {
...
+
+# wait for Rook deployment to be running and healthy
+function rook_maybe_wait_for_rollout() {
+    log "Awaiting Rook pods to transition to Running"
+    if ! spinner_until 120 wait_for_running_pods "rook-ceph"; then
+        logWarn "Rook-ceph rollout did not complete"
+    fi
+
+    log "Awaiting for Ceph Cluster to be Ready"
+    if ! spinner_until 120 rook_cephcluster_ready; then
```
emosbaugh: does it make sense to wait for ready before pods? the operator may restart pods if its still updating.

scripts/common/common.sh
```diff
@@ -1036,3 +1036,58 @@ function canonical_image_name() {
...
+            # ignore container ready status for pods managed by the Job controller
+            if [ "$container_status" != "true" ] && [ $is_job_controller -eq 0 ]; then
+                pods_not_ready=1
+                log "  Container, $container ($pod), is not ready: $container_status"
+                break
+            fi
+        done
+
+        if [ $pods_not_ready -ne 0 ]; then
+            break
```
emosbaugh, line 1083: why not just return 1 here?

scripts/common/common.sh
```diff
@@ -1036,3 +1036,58 @@ function canonical_image_name() {
...
+
+        containers=$(kubectl get pod "$pod" -n "$namespace" -o jsonpath="{.spec.containers[*].name}")
+        for container in $containers; do
+            container_status=$(kubectl get pod "$pod" -n "$namespace" -o jsonpath="{.status.containerStatuses[?(@.name==\"$container\")].ready}")
+            
+            # ignore container ready status for pods managed by the Job controller
+            if [ "$container_status" != "true" ] && [ $is_job_controller -eq 0 ]; then
+                pods_not_ready=1
+                log "  Container, $container ($pod), is not ready: $container_status"
+                break
```
emosbaugh, line 1078: why not just return 1 here?

scripts/common/common.sh
```diff
@@ -1036,3 +1036,58 @@ function canonical_image_name() {
...
+        
+        # determine if pod is manged by a Job
+        if kubectl get pod "$pod" -n "$namespace" -o jsonpath='{.metadata.ownerReferences[*].kind}' | grep -q "Job"; then
+            is_job_controller=1
+        fi
+
+        if [ "$status" != "Running" ] && [ "$status" != "Succeeded" ]; then
+            pods_not_ready=1
+            log "  Pod, $pod, is not ready: $status"
+            break
```
emosbaugh, line 1067: why not just return 1 here?

scripts/common/common.sh
```diff
@@ -1036,3 +1036,58 @@ function canonical_image_name() {
...
+                break
+            fi
+        done
+
+        if [ $pods_not_ready -ne 0 ]; then
+            break
+        fi
+    done
+
+    if [ $pods_not_ready -ne 0 ]; then
```
emosbaugh, line 1087: this is not needed if you return 1 in the loop

emosbaugh (changes requested):
also i dont think this should be called from the post init as other addons depend on rook ready to serve requests. pehaps it is better in the rook() and rook_already_applied() functions?

addons/rook/template/base/install.sh
rrpolanco: I will address this in this PR: https://github.com/replicatedhq/kURL/pull/4101

laverya (commented):
I would recommend excluding evicted pods, because those don't get cleaned up by default

addons/rook/template/base/install.sh
rrpolanco: addressed in #4101

scripts/common/common.sh
```diff
@@ -1036,3 +1036,47 @@ function canonical_image_name() {
...
+
+    for pod in $ns_pods; do
+        status=$(kubectl get pod "$pod" -n "$namespace" -o jsonpath='{.status.phase}')
+
+        # determine if pod is manged by a Job
+        if kubectl get pod "$pod" -n "$namespace" -o jsonpath='{.metadata.ownerReferences[*].kind}' | grep -q "Job"; then
+            is_job_controller=1
+        fi
+
+        if [ "$status" != "Running" ] && [ "$status" != "Succeeded" ]; then
```
camilamacedo86, line 1063: Should we not also add Completed?

scripts/common/rook.sh
```diff
@@ -225,3 +225,15 @@ function current_ceph_version() {
...
         | awk -F'-' '{ print $1 }'
 }
+
+function rook_operator_ready() {
+    local rook_status_phase=
+    local rook_status_msg=
+    rook_status_phase=$(kubectl -n rook-ceph get cephcluster rook-ceph --template '{{.status.phase}}')
+    rook_status_msg=$(kubectl -n rook-ceph get cephcluster rook-ceph --template '{{.status.message}}')
+    if [  "$rook_status_phase" != "Ready"  ]; then
+        log "Rook operator is not ready: $rook_status_msg"
```
camilamacedo86, line 235:
```suggestion
        logWarn "Rook operator is not ready: $rook_status_msg"
```
Could we use logWarn to write in yellow and make easier we check it out?
Also, what about we use wait to still checking for some time and does not bail in this case.
It does not seems a problem in real installs since the things would be ready after a little more time than in the tests.

scripts/common/common.sh
```diff
@@ -1036,3 +1036,47 @@ function canonical_image_name() {
...
+    local containers=
+
+    ns_pods=$(kubectl get pods -n "$namespace" -o jsonpath='{.items[*].metadata.name}')
+
+    if [ -z "$ns_pods" ]; then
+        log "No pods found in namespace $namespace."
+        return 0
+    fi
+
+    for pod in $ns_pods; do
```
camilamacedo86, line 1055: Would be possible we have a away to pass a timeout so that we could check that for a period of time ?

scripts/common/common.sh
rrpolanco: `Completed` pods have a status.phase of `Succeeded`.

camilamacedo86: We are testing this one here : https://testgrid.kurl.sh/run/pr-4101-5175666-rook-1.10.11-k8s-docker-2023-02-14T22:34:55Z Am I right? So let's just wait the tests to move forward, otherwise /lgtm

scripts/common/common.sh
```diff
@@ -1036,3 +1036,47 @@ function canonical_image_name() {
...
+            log "  Pod, $pod, is not ready: $status"
+            return 1
+        fi
+
+        containers=$(kubectl get pod "$pod" -n "$namespace" -o jsonpath="{.spec.containers[*].name}")
+        for container in $containers; do
+            container_status=$(kubectl get pod "$pod" -n "$namespace" -o jsonpath="{.status.containerStatuses[?(@.name==\"$container\")].ready}")
+            
+            # ignore container ready status for pods managed by the Job controller
+            if [ "$container_status" != "true" ] && [ $is_job_controller -eq 0 ]; then
```
emosbaugh, line 1073: for equality comparison i always prefer string comparison in bash as you will not get an error for an empty or non numeric value

```suggestion
            if [ "$container_status" != "true" ] && [ "$is_job_controller" = "0" ]; then
```

emosbaugh (changes requested):

emosbaugh (dismissed):

scripts/common/common.sh
```diff
@@ -1036,3 +1036,52 @@ function canonical_image_name() {
...
+            
+            # ignore container ready status for pods managed by the Job controller
+            if [ "$container_status" != "true" ] && [ "$is_job_controller" = "0" ]; then
+                log "  Container, $container ($pod), is not ready: $container_status"
+                return 1
+            fi
+        done
+    done
+
+    log "All pods and their containers in namespace $namespace are in Running phase."
```
laverya, line 1085:
```suggestion
    log "All pods and their containers in namespace $namespace are running."
```

scripts/common/common.sh
```diff
@@ -1036,3 +1036,52 @@ function canonical_image_name() {
     fi
     echo "$image"
 }
+
+# wait_for_running_pods waits for pod(s) in a given namspace to transition to Running status
+function wait_for_running_pods() {
```
laverya, line 1041: does this function really _wait_ for anything?
Or is it intended to be used with a spinner?

If it's intended to be used with a spinner, it shouldn't be logging things IIRC - that messes up the spinner

rrpolanco: It's meant to be used as an arg to `spinner_until()`. I'll change the name and remove the logs.

scripts/common/common.sh
camilamacedo86: I am just looking for we ensure that we will not fail because pods are completed.

addons/rook/template/base/install.sh
camilamacedo86: I think we can close this one

scripts/common/rook.sh
camilamacedo86: Closing this one I think it is addressed / discussed now.

camilamacedo86:
IMO to get this one merged it is only missing we have the same code applied to https://github.com/replicatedhq/kURL/pull/4101.

So, we can check the tests and ensure that all worked fine and get this one merged.

camilamacedo86 (approved):
We need to check the code working to ensure that is fine.
However, shows that we can only check it in the PR https://github.com/replicatedhq/kURL/pull/4101 after we get this merged . 

So, approved it will not also harm anything and we can test the things out in the follow up PR.

PR #4066, opened 2023-02-09
fixes: add retry to apply original.yaml after crd is installed

zhongweili2010:
To accommodate slower machines when apply original installer spec, install spec sometimes fails after crd is already installed. 
  
Fixes #

```release-note
NONE
```

NONE

### DIFF
--- a/scripts/common/yaml.sh
+++ b/scripts/common/yaml.sh
@@ -125,7 +125,7 @@ function apply_installer_crd() {
         cat > $ORIGINAL_INSTALLER_SPEC <<EOL
 ${INSTALLER_YAML}
 EOL
-        kubectl apply -f "$ORIGINAL_INSTALLER_SPEC"
+        try_1m kubectl apply -f "$ORIGINAL_INSTALLER_SPEC"
     fi
 
     try_1m kubectl apply -f "$MERGED_YAML_SPEC"


emosbaugh:
Hello. We appreciate the contribution! Would it be possible to attach logs here from the output of the installation script reproducing the issue to help me better understand? Thank you!

scripts/common/yaml.sh
```diff
@@ -125,7 +125,7 @@ function apply_installer_crd() {
         cat > $ORIGINAL_INSTALLER_SPEC <<EOL
 ${INSTALLER_YAML}
 EOL
-        kubectl apply -f "$ORIGINAL_INSTALLER_SPEC"
+        try_1m kubectl apply -f "$ORIGINAL_INSTALLER_SPEC"
```
camilamacedo86, line 128: The line just apply the manifest.
The try will be re-doing the same command each 2 seconds : https://github.com/replicatedhq/kURL/blob/c09c7fb818be8cfab5d5d51b0a1e2bd26c4d9a59/scripts/common/common.sh#L440-L458

So, why would need to be required to re-duo kubectl apply many times? 
What would be solved with that?

+1 to the @emosbaugh comments. That would be great if you could share the problem, and all logs of the issue that you are facing so we can better check it out and see how we can help.

PR #4069, opened 2023-02-10
check available disk space before performing rook upgrade

laverya:
add ability for user to continue despite potentially lacking disk space

This PR estimates the amount of disk space required to upgrade Rook to the specified version and presents a warning if insufficient space is detected.

Both docker and containerd are supported, a limit of 85% is imposed on filesystems containing docker/containerd, and different logic is used if /var/lib/kurl is on the same filesystem as the image store.

An example message (when upgrading from Rook 1.0.4 to 1.10.8) is:
```bash
Would you like to continue? (y/N) y
Not enough disk space to upgrade Rook.
You need at least 78650 MB of free space on the filesystem containing /var/lib/kurl and /var/lib/containerd - and to have 15% free space after that to avoid image pruning.
Currently, only 18763 MB of free space is available before reaching 85% capacity.
If you have already loaded images or started this Rook upgrade, it is possible that less space will be required. Would you like to continue anyways?
```

Fixes #

```release-note
Warn the user if there is likely to be insufficient space to upgrade Rook many versions.
```

### DIFF
--- a/scripts/common/rook-upgrade.sh
+++ b/scripts/common/rook-upgrade.sh
@@ -115,20 +115,20 @@ function rook_upgrade_storage_check() {
         container_directory="/var/lib/containerd"
     fi
 
-    # if $container_directory and /var/lib/kurl are on the same filesystem, we need to check that there is space for all of the files
-    if [ "$(df -P "$container_directory" | awk 'END{print $1}')" = "$(df -P /var/lib/kurl | awk 'END{print $1}')" ]; then
+    # if $container_directory and $DIR are on the same filesystem, we need to check that there is space for all of the files
+    if [ "$(df -P "$container_directory" | awk 'END{print $1}')" = "$(df -P $DIR | awk 'END{print $1}')" ]; then
         # in total, we need space for 5.5x the archive size, AND there must be 15% free space on the filesystem afterwards
         local total_required_size=
         total_required_size=$((archive_size * 11 / 2)) # 5.5x archive size, rounded to an integer
 
         local free_kb=
         local free_mb=
-        free_kb="$(df -P /var/lib/kurl | awk 'END{print $4}')"
+        free_kb="$(df -P $DIR | awk 'END{print $4}')"
         free_mb="$((free_kb / 1024))"
 
         local total_kb=
         local total_mb=
-        total_kb="$(df -P /var/lib/kurl | awk 'END{print $2}')"
+        total_kb="$(df -P $DIR | awk 'END{print $2}')"
         total_mb="$((total_kb / 1024))"
 
         local available_mb=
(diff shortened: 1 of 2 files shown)


camilamacedo86 (approved):
It shows fine to me. 
The hard thing is that we cannot test it out prior get merged.

So, @laverya please run a testgrid afterwords with the scenarios to ensure that all is fine.

emosbaugh:
What about secondary nodes? Would it be possible to add a similar check to

rook_upgrade_tasks_rook_upgrade

or

rook_upgrade_tasks_load_images

laverya:
> ./scripts/common/rook-upgrade-test.sh: line 95: rook_upgrade_required_disk_space: command not found
> ASSERT:1.0 to 1.5 expected:<6100> but was:<>


🤦 I changed the function name

laverya:
> What about secondary nodes? Would it be possible to add a similar check to
>
> rook_upgrade_tasks_rook_upgrade
>
>or
>
>rook_upgrade_tasks_load_images

yeah, that makes sense to do

scripts/common/rook-upgrade.sh
```diff
@@ -95,6 +100,96 @@ function rook_upgrade_prompt() {
...
+        local available_mb=
+        available_mb="$((free_mb - total_mb * 3 / 20))" # free space, excluding 15% of the total
+
+        if [ "$available_mb" -lt "$total_required_size" ]; then
+            logWarn "Not enough disk space to upgrade Rook."
+            logWarn "You need at least $total_required_size MB of free space on the filesystem containing /var/lib/kurl and $container_directory - and to have 15%% free space after that to avoid image pruning."
+            logWarn "Currently, only $available_mb MB of free space is available before reaching 85%% capacity."
+            logWarn "If you have already loaded images or started this Rook upgrade, it is possible that less space will be required. Would you like to continue anyways?"
+            if ! confirmN; then
+                return 1
```
camilamacedo86, line 143: Why we are not using bail here? WDYT about use bail?

scripts/common/rook-upgrade.sh
```diff
@@ -95,6 +100,96 @@ function rook_upgrade_prompt() {
...
+            fi
+        fi
+    else
+        local kurl_dir_size=
+        kurl_dir_size=$((archive_size * 2))
+
+        local kurl_free_kb=
+        local kurl_free_mb=
+        kurl_free_kb="$(df -P /var/lib/kurl | awk 'END{print $4}')"
```
camilamacedo86, line 152: I think we need to use the DIR here right? Otherwise, we are unable to test it from the source code.

emosbaugh: Good call. Can we use the vars for this? `$BIN_KURL`

https://github.com/replicatedhq/kURL/blob/592f3ac1bcca78d91c3b7611fbc3b99752d4e40b/scripts/common/utilbinaries.sh#L8

laverya: yeah, not hard to change

(just `$DIR` though)

scripts/common/rook-upgrade.sh
laverya: because for every other bit in this flow we're returning and not bailing

laverya:
```
    if ! rook_upgrade_prompt "$current_version" "$desired_version" ; then
        bail "Not upgrading Rook"
    fi

    if ! rook_upgrade_storage_check "$current_version" "$desired_version" ; then
        bail "Not upgrading Rook"
    fi
```

camilamacedo86: But in the `confirmN` we do not bail. Here:

```
if ! confirmN; then
                return 1
```

Should we not as we do in the other places?

laverya: I can change it to just be `confirmN`, which is what we do in the `rook_upgrade_prompt` section

It's the same thing though, just spelled out

laverya: `rook_upgrade_storage_check` returns 1, and so there is a bail in the snippet I pasted

camilamacedo86: I see I think that is fine 👍 
Many ways to do the same :-)

camilamacedo86 (dismissed):

camilamacedo86 (approved):

PR #4071, opened 2023-02-10
WIP fix: flannel issues when force reapply

camilamacedo86:

Fixes #

```release-note

```


addons/flannel/template/base/install.sh
```diff
@@ -67,6 +67,16 @@ function flannel() {
         kubectl -n kube-flannel apply -k "$dst/"
     fi
 
+    # We will remove the flannel pods to let it be re-created
+    # in order to workaround the issue scenario described in
+    # https://github.com/flannel-io/flannel/issues/1721
+    if [ "$KUBERNETES_UPGRADE" == "1" ]; then
+       sleep 30
```
emosbaugh, line 74: i have an issue with timeouts of arbitrary length. it introduces flakes in our product. could we get creative here?

addons/flannel/template/base/install.sh
```diff
@@ -67,6 +67,16 @@ function flannel() {
...
+       log "Deleting pods from kube-flannel namespace"
+       kubectl delete --all pods --namespace=kube-flannel
+       sleep 60
```
emosbaugh, line 77: and here?

emosbaugh (changes requested):

addons/flannel/template/base/install.sh
camilamacedo86: The sleep if just to give a sometime to ensure that all was applied (line 67 ` kubectl -n kube-flannel apply -k "$dst/"`), not sure how creative we would be.

addons/flannel/template/base/install.sh
camilamacedo86: After think further here. I think is safe we have this one, I added just to ensure that the bin would be created but in my tests it is very fast and take a couple of seconds. On top of that,  the `check_network` has a waiting timeout already.

addons/flannel/template/base/install.sh
emosbaugh: could we not check that the flannel ds pods are running and ready?

camilamacedo86: I think we can just move the test bellow the health check. That is a good point thank you.

emosbaugh (approved):

PR #4073, opened 2023-02-10
feat: preparing longhorn for migration.

ricardomaraschini:

Calls `longhorn_prepare_for_migration` function in both OpenEBS and Rook add-ons code. `longhorn_prepare_for_migration` scales down pods using Longhorn volumes, as it is called earlier in the migration process we don't have pods pending on "Terminating" state anymore (I could not reproduce it anymore, at least).

This PR also migrates data from  Minio object store to Rook object store when moving away from Minio + Longhorn duo. Migrations from Longhorn to Rook has also been blocked in single node clusters.

Fixes #65977

### DIFF
--- a/addons/openebs/3.3.0/install.sh
+++ b/addons/openebs/3.3.0/install.sh
@@ -388,4 +388,8 @@ function openebs_prompt_migrate_from_longhorn() {
     if ! confirmN; then
         bail "Not migrating"
     fi
+
+    if ! longhorn_prepare_for_migration; then
+        bail "Not migrating"
+    fi
 }
--- a/addons/openebs/3.3.0/install.sh
+++ b/addons/openebs/3.3.0/install.sh
@@ -92,7 +92,6 @@ function openebs_maybe_rook_migration_checks() {
     if [ -n "$rook_scs_pvmigrate_dryrun_output" ] || [ -n "$rook_default_sc_pvmigrate_dryrun_output" ] ; then
         log "$rook_scs_pvmigrate_dryrun_output"
         log "$rook_default_sc_pvmigrate_dryrun_output"
-
         bail "Cannot upgrade from Rook to OpenEBS due to previous error."
     fi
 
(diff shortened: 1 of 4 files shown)


camilamacedo86 (commented):
The changes shows great. But would be nice we check the tests results before we get merged right? Let's wait a little .. I am happy to approve this one.

camilamacedo86 (dismissed):
It shows nice for me 👍

addons/openebs/3.3.0/install.sh
```diff
@@ -384,8 +384,18 @@ function openebs_prompt_migrate_from_longhorn() {
         bail "Not migrating"
     fi
 
+    if [ -z "$MINIO_VERSION" ] ; then
+        logFail "    You can only migrate from Longhorn to OpenEBS if you are specifying Minio as your Object Store."
+        logFail "    Please make sure you are including Minio in your kURL Installer spec and try again."
+        bail "Not migrating"
+    fi
```
laverya, line 391: what if they have no object store at all?

camilamacedo86: That is a good point.

ricardomaraschini: Does that mean "no kots" ?

ricardomaraschini: Screenshot 2023-02-13 at 18 50 51

laverya: You can have kots with no object store 😄

camilamacedo86: theoretically we could have a kurl install without kots. 
In the reality I think you are right :-)

In this way, would be possible:

- warning instead of bail
- in the waning suggest add minio and say what are the implications
- confirm if they want or not continue to ensure
- if they continue without minio skip the specific step and just migrate ?

laverya: There is a flag (in kURL) to run kots with no object store

ricardomaraschini: I have removed the check for Minio. My reasoning is that if the cluster is running Kots+Longhorn without Minio then a simple PVC migration will be enough. I will test this to make sure it works.

ricardomaraschini: Tested with Kots' `disableS3` option and it works (after removing the Minio presence check introduced by this PR).

rrpolanco (approved):

emosbaugh (approved):

PR #4074, opened 2023-02-10
adds new command to get hostname as it is done by k8s api to use it crosswise

camilamacedo86:

Currently, kurl scripts and ekco use the func `$(hostname | tr '[:upper:]' '[:lower:]')` to get the node name. However, in some edge cases scenarios it seems that we have a mismatch. 

Therefore, by looking at kubernetes implementation we can check that they are using [os.Hostname()](https://github.com/kubernetes/kubernetes/blob/f58f70bd5730658505042cd9baa80f72d3b6e31e/staging/src/k8s.io/component-helpers/node/util/hostname.go#L28-L46) which its implementation will look for `cat /proc/sys/kernel/hostname.` This PR ensure that we are getting the information at the same way that kubernetes does for Linux environments to avoid problematic edge cases scenarios. 

As agreed in the review this PR creates a new command to do the same k8s implementation

Screenshot 2023-02-13 at 10 52 46

In a follow up we will start to use it. 

Partial Fixes # [sc-68468]

```release-note
Adds new command to kurl binary to get hostname `kurl host hostname`
```

### DIFF
--- a/pkg/host/hostname.go
+++ b/pkg/host/hostname.go
@@ -7,6 +7,8 @@ import (
 )
 
 // GetHostname returns the node hostname using kubernetes as lib to ensure that will obtain the same value
+// This code is the same used kubernetes project: (We copied to avoid to add a dependency only because of this snipt)
+//https://github.com/kubernetes/kubernetes/blob/6b34fafdaf5998039c7e01fa33920a641b216d3e/staging/src/k8s.io/component-helpers/node/util/hostname.go#L25-L46
 func GetHostname() (string, error) {
 	nodeName, err := os.Hostname()
 	if err != nil {
--- a/pkg/host/hostname.go
+++ b/pkg/host/hostname.go
@@ -8,7 +8,7 @@ import (
 
 // GetHostname returns the node hostname using kubernetes as lib to ensure that will obtain the same value
 // This code is the same used kubernetes project: (We copied to avoid to add a dependency only because of this snipt)
-//https://github.com/kubernetes/kubernetes/blob/6b34fafdaf5998039c7e01fa33920a641b216d3e/staging/src/k8s.io/component-helpers/node/util/hostname.go#L25-L46
+// https://github.com/kubernetes/kubernetes/blob/6b34fafdaf5998039c7e01fa33920a641b216d3e/staging/src/k8s.io/component-helpers/node/util/hostname.go#L25-L46
 func GetHostname() (string, error) {
 	nodeName, err := os.Hostname()
 	if err != nil {
--- a/pkg/host/hostname.go
+++ b/pkg/host/hostname.go
@@ -1,0 +1,25 @@
+package host
+
+import (
+	"fmt"
+	"os"
+	"strings"
+)
+
+// GetHostname returns the node hostname using kubernetes as lib to ensure that will obtain the same value
+func GetHostname() (string, error) {
+	nodeName, err := os.Hostname()
+	if err != nil {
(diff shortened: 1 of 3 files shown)


addons/containerd/1.3.7/install.sh
```diff
@@ -178,7 +178,7 @@ function containerd_migrate_from_docker() {
 
     local kubeconfigFlag="--kubeconfig=/etc/kubernetes/kubelet.conf"
 
-    local node=$(hostname | tr '[:upper:]' '[:lower:]')
+    local node=$(cat /proc/sys/kernel/hostname | tr '[:upper:]' '[:lower:]')
```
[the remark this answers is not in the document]
camilamacedo86: See the k8s implementation and comment:

https://github.com/kubernetes/kubernetes/blob/f58f70bd5730658505042cd9baa80f72d3b6e31e/staging/src/k8s.io/component-helpers/node/util/hostname.go#L38-L40

Then, see the code source for os.Hostname() to linux platforms:

https://github.com/golang/go/blob/2ebe77a2fda1ee9ff6fd9a3e08933ad1ebaea039/src/os/sys_linux.go#L12-L53

And the line 38 from the link that you added : https://cs.opensource.google/go/go/+/refs/tags/go1.20:src/os/sys_linux.go;l=38

So, I think we should stick with `cat /proc/sys/kernel/hostname` . Try uname first, shows that is required because it will run in other platforms such as Android.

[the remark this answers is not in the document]
camilamacedo86: In this PR we are adding the command and in a follow up we will use it.

pkg/host/hostname.go
```diff
@@ -0,0 +1,25 @@
+package host
+
+import (
+	"fmt"
+	"os"
+	"strings"
+)
+
+// GetHostname returns the node hostname using kubernetes as lib to ensure that will obtain the same value
+func GetHostname() (string, error) {
```
emosbaugh, line 10: should we be using node/util.GetHostname()?

if we dont use it directly can you add a comment with a link to the code you borrowed this from?

emosbaugh (changes requested):

pkg/host/hostname.go
camilamacedo86: I did not want to add more deps to kurl only because of this piece of code. 
I will add the link for it.

pkg/host/hostname.go
```diff
@@ -0,0 +1,25 @@
+package host
+
+import (
+	"fmt"
+	"os"
+	"strings"
+)
+
+// GetHostname returns the node hostname using kubernetes as lib to ensure that will obtain the same value
```
camilamacedo86, line 9:
```suggestion
// GetHostname returns the node hostname using kubernetes as lib to ensure that will obtain the same value
// This code is the same used kubernetes project: (We copied to avoid to add a dependency only because of this piece)
//https://github.com/kubernetes/kubernetes/blob/6b34fafdaf5998039c7e01fa33920a641b216d3e/staging/src/k8s.io/component-helpers/node/util/hostname.go#L25-L46
```

emosbaugh:
Looks like ci is failing https://github.com/replicatedhq/kURL/actions/runs/4167040563/jobs/7212129642

rrpolanco (dismissed):

camilamacedo86:
Hi @emosbaugh,

> Looks like ci is failing https://github.com/replicatedhq/kURL/actions/runs/4167040563/jobs/7212129642

Fixed. it is just a lint issue.

emosbaugh (approved):

PR #4076, opened 2023-02-10
chore: sleep before post upgrade test.

ricardomaraschini:

This PR solves what seems to be a flake during Longhorn to Rook/OpenEBS migration tests (testgrid). During the upgrade Pods mounting Longhorn volumes are scaled down and they can take some time to get back up once their data has been migrated to Rook or OpenEBS.

This pr adds a sleep of 2 minutes before running the post upgrade test (this time should be more than enough for the simple Pod we are using to test to come up).


laverya (approved):

laverya:
I'll approve this because we're not waiting for a kurl pod here, we're waiting for an 'application' pod

(kurl pods should have waits in the kurl code)

camilamacedo86 (approved):

PR #4078, opened 2023-02-10
use actual flannel version in addons

laverya:

Fixes #

```release-note

```


laverya:
I also need to update the automation to no longer depend on the upstream updating the version in their documentation, but that'll be a different PR

rrpolanco (approved):

PR #4094, opened 2023-02-14
Replace shell by hostname

camilamacedo86:

Fixes # [sc-68468]

```release-note

```

### DIFF
--- a/addons/ekco/template/base/reboot/shutdown.sh
+++ b/addons/ekco/template/base/reboot/shutdown.sh
@@ -1,8 +1,9 @@
 #!/bin/bash
 
 export KUBECONFIG=/etc/kubernetes/kubelet.conf
+KURL_INSTALL_DIRECTORY=/var/lib/kurl
 
-kubectl cordon "$(hostname | tr '[:upper:]' '[:lower:]')"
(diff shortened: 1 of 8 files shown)


addons/ekco/template/base/reboot/shutdown.sh
```diff
@@ -1,8 +1,9 @@
 #!/bin/bash
 
 export KUBECONFIG=/etc/kubernetes/kubelet.conf
+KURL_INSTALL_DIRECTORY=/var/lib/kurl
```
emosbaugh, line 4: this is technically configurable

see https://kurl.sh/docs/install-with-kurl/advanced-options `kurl-install-directory`

emosbaugh (changes requested):

addons/ekco/template/base/reboot/shutdown.sh
camilamacedo86: The only way to ensure that is if we in the install script overwrite this value. 
Note that those scripts are shipped with the add on.

camilamacedo86:
Closes in favor of : https://github.com/replicatedhq/kURL/pull/4116

PR #4108, opened 2023-02-15
testgrid: Update reference to have the commit

camilamacedo86:
testgrid: Update reference to have the commit


laverya:
But the reference you're deleting already had the commit?

If that was the wrong commit, why not update this to use the right one (and keep the date/prefix/etc)?

laverya (approved):

PR #4110, opened 2023-02-16
doc: proposal enhancement template for major changes

camilamacedo86:

For we have a template to follow up and began to the better discuss and granulate the proposals for major changes.

### DIFF
--- a/docs/enhancement/template.md
+++ b/docs/enhancement/template.md
@@ -1,0 +1,145 @@
+| Authors       | Creation Date | Status      | Extra |
+|---------------|---------------|-------------|---|
+| @name | date | Implementeble | - |
+
+Title of the Design/Proposal
+===================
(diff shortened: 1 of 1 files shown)


chris-sanders (changes requested):
This is a very heavy process and needs a lot more discussion than just a Pull Request.

We already have the ADR process, and our product teams have two different proposal processes to choose from. It is not clear to me which problem this solves that isn't already addressed.

I think this would be better handled as a team discussion on stand ups or our future of kurl meetings. Either way you should sync first with our Product Manager as this is very much a product driven process and Kyle can help you once he understand the issues you are trying to resolve.

chris-sanders:
Due to the issues I called out above, primarily this isn't the right way to have this discussion I'm going to close this PR.

PR #4128, opened 2023-02-17
fix(ekco): fix error flags --detach and --rm cannot be specified together

emosbaugh:

Fixes error when using containerd 1.6.18

```
Kubernetes bootstrap token: qrl3ym.ra209pop9td8k4e9
This token will expire in 24 hours
Updating the Kubernetes apiserver load balancer
ctr: container "bootstrap-lb" in namespace "k8s.io": not found
time="2023-02-17T19:33:20Z" level=error msg="failed to delete container \"bootstrap-lb\"" error="container \"bootstrap-lb\" in namespace \"k8s.io\": not found"
ctr: container "bootstrap-lb" in namespace "k8s.io": not found
ctr: flags --detach and --rm cannot be specified together
An error occurred on line 348
```

Fixes NONE

```release-note
Fixes an error when using Containerd 1.6.18 and the EKCO Internal Load Balancer that causes the install script to fail with error "ctr: flags --detach and --rm cannot be specified together".
```

NONE


laverya:
Wouldn't the solution be to get rid of --detach, so that we don't leave the container laying around?

laverya (dismissed):
I suppose we already delete the container so it doesn't matter if we leave it around

emosbaugh:
> Wouldn't the solution be to get rid of --detach, so that we don't leave the container laying around?

i am just copying the docker logic

https://github.com/replicatedhq/kURL/blob/8a977243933fa2dc992a628ed10b275b50210b3c/addons/ekco/template/base/install.sh#L325

```
        docker run -d -p "6444:6444" -v /etc/haproxy:/usr/local/etc/haproxy --name bootstrap-lb "${EKCO_HAPROXY_IMAGE}"
```

laverya (dismissed):

rrpolanco (approved):

laverya (approved):

PR #4102, opened 2023-02-15
Automated Flannel version update 0.21.2

replicated-ci-kurl:
Automated changes by the [cron-flannel-update](https://github.com/replicatedhq/kURL/blob/main/.github/workflows/update-flannel.yaml) GitHub action

```release-note
Adds [Flannel add-on](https://kurl.sh/docs/add-ons/flannel) version 0.21.2.
```

PR #4133, opened 2023-02-20
Add addon v0.21.2 for flannel and address a fix to its script to generate the new versions

camilamacedo86:

To fix the automated generation of the flannel addon to ensure that the flannel image used will be the same version of its release. And to add the new version v0.21.2.

Fixes # [sc-51232]

Closes: https://github.com/replicatedhq/kURL/pull/4102/files

```release-note
Adds [Flannel add-on](https://kurl.sh/docs/add-ons/flannel) version 0.21.2.
```

NONE

### DIFF
--- a/addons/flannel/template/generate.sh
+++ b/addons/flannel/template/generate.sh
@@ -10,8 +10,12 @@ function generate() {
     cp -r ./base/* "$dir"
 
     download_yaml
-
+    replace_flannel_image_with_version
     find_images_in_yaml
+}
+
+function replace_flannel_image_with_version() {
+    sed -E -i "s/docker\.io\/flannel\/flannel:.*/docker\.io\/flannel\/flannel:v$version/g" "$dir/yaml/kube-flannel.yml"
 }
 
 function download_yaml() {
--- a/addons/flannel/0.21.2/Manifest
+++ b/addons/flannel/0.21.2/Manifest
@@ -1,0 +1,2 @@
+image flannel-flannel docker.io/flannel/flannel:v0.20.2
+image flannel-flannel-cni-plugin docker.io/flannel/flannel-cni-plugin:v1.1.2
--- a/addons/flannel/0.21.2/Manifest
+++ b/addons/flannel/0.21.2/Manifest
@@ -1,2 +1,2 @@
-image flannel-flannel docker.io/flannel/flannel:v0.20.2
+image flannel-flannel docker.io/flannel/flannel:v0.21.2
 image flannel-flannel-cni-plugin docker.io/flannel/flannel-cni-plugin:v1.1.2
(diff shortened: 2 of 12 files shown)


addons/flannel/template/generate.sh
```diff
@@ -10,10 +10,14 @@ function generate() {
     cp -r ./base/* "$dir"
 
     download_yaml
-
+    replace_flannel_image_with_version
     find_images_in_yaml
 }
 
+function replace_flannel_image_with_version() {
+    sed -i -e "s/flannel:v0.20.2/flannel:v$version/" "$dir/yaml/kube-flannel.yml"
```
laverya, line 18: Unfortunately this one always worked because they will be updating the version in the docs soon

We could probably use a regex instead

camilamacedo86: @laverya , thank you for catching that. 

OH, that is a mistake I pushed my test. 
It is within the regex of course. 
Hold on.

camilamacedo86, line 18:
```suggestion
    sed -e "s/flannel:v?(\d+(?:\.\d+)*)/flannel:v$version/g" "$dir/yaml/kube-flannel.yml"
```

addons/flannel/template/generate.sh
```diff
@@ -10,10 +10,14 @@ function generate() {
     cp -r ./base/* "$dir"
 
     download_yaml
-
+    replace_flannel_image_with_version
     find_images_in_yaml
 }
 
+function replace_flannel_image_with_version() {
+    sed -e "s/flannel:v?(\d+(?:\.\d+)*)/flannel:v$version/g" "$dir/yaml/kube-flannel.yml"
```
laverya, line 18: Why not

```suggestion
    sed -e "s/flannel:v?(\d+(?:\.\d+)(?:\.\d+))/flannel:v$version/g" "$dir/yaml/kube-flannel.yml"
```

That way it only matches x.y.z and not also x.y or x.y.z.a

emosbaugh: what about the Manifest?

emosbaugh (changes requested):

camilamacedo86:
Hi @emosbaugh,

> what about the Manifest?

What do you mean with? This implementation works in this way:
- We get the resources (yaml) from the docs
- We read all images from the resources (yaml) and then we generated the Manifests

What we are doing here is before we generate the manifest ensure that the flannel images tag version will be the same version used for the tag release. In this way, the manifest file contains the image and versions accordingly. Is that make sense?

emosbaugh:
> Hi @emosbaugh,
> 
> > what about the Manifest?
> 
> What do you mean with? This implementation works in this way:
> 
> * We get the resources (yaml) from the docs
> * We read all images from the resources (yaml) and then we generated the Manifests
> 
> What we are doing here is before we generate the manifest ensure that the flannel images tag version will be the same version used for the tag release. In this way, the manifest file contains the image and versions accordingly. Is that make sense?

Thanks I missed that

emosbaugh (approved):
This lgtm for now. I worry though that if the flannel image is not updated then its possible the cni plugin image needs an update as well. We can address that in the future if necessary.

PR #4137, opened 2023-02-21
DO NOT MERGE - I AM USING IT FOR TESTs

camilamacedo86:
DO NOT MERGE

### DIFF
--- a/scripts/common/rook.sh
+++ b/scripts/common/rook.sh
@@ -64,34 +64,33 @@ function remove_rook_ceph() {
     # make sure there aren't any PVs using rook before deleting it
     all_pv_drivers="$(kubectl get pv -o=jsonpath='{.items[*].spec.csi.driver}')"
     if echo "$all_pv_drivers" | grep "rook" &>/dev/null ; then
-        # do stuff
-        printf "%b" "$RED"
-        printf "ERROR: \n"
-        printf "There are still PVs using rook-ceph.\n"
-        printf "Remove these PVs before continuing.\n"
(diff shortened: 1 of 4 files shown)


scripts/common/rook.sh
```diff
@@ -198,10 +198,27 @@ function rook_ceph_to_sc_migration() {
...
+                logWarn "Rook object store was successfully migrate but not PVCs. Therefore, rook can not be removed"
+            fi
+            if [ "$DID_MIGRATE_ROOK_PVCS" == "1" ] && [ "$DID_MIGRATE_ROOK_OBJECT_STORE" == "1" ]; then
+               logStep "Removing Rook"
+               report_addon_start "rook-ceph-removal" "v1"
+               remove_rook_ceph
+               report_addon_success "rook-ceph-removal" "v1"
+            fi
+        fi
+        if [ -z "$MINIO_VERSION" ]; then
```
emosbaugh, line 215: I dont think we should remove the object store data when it was not migrated. losing data is a worst case scenario

emosbaugh (changes requested):
I am not sure what this pull request is solving

scripts/common/rook.sh
camilamacedo86: Have MINIO is not mandatory. What we are doing here is calling `remove_rook_ceph` when Minio is not installed.

camilamacedo86: Minio is not mandatory. What should we do when MINIO is not selected?

PR #4141, opened 2023-02-22
chore: add test to check migration from rook without minio

camilamacedo86:

Perform the test with upper versions to ensure that the same behaviour can properly work with them

Motivated By # [sc-69454]

It is falling because rook-ceph namespace is not deleted. The test is for we start to cover this scenario.

### DIFF
--- a/addons/openebs/template/testgrid/k8s-docker-localpv.yaml
+++ b/addons/openebs/template/testgrid/k8s-docker-localpv.yaml
@@ -230,6 +230,47 @@
     # pulls the image we pushed before the migration.
     test_pull_image_from_registry
 
+- name: localpv migrate from rook and without MINIO
+  flags: "yes"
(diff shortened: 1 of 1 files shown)


addons/openebs/template/testgrid/k8s-docker-localpv.yaml
```diff
@@ -230,6 +230,47 @@
...
+    containerd:
+      version: "latest"
+    openebs:
+      isLocalPVEnabled: true
+      localPVStorageClassName: openebs
+      namespace: openebs
+      version: "__testver__"
+      s3Override: "__testdist__"
+    kotsadm:
+      version: "latest"
```
camilamacedo86, line 258:
```suggestion
```

camilamacedo86:
I will close this one because will need to discuss this scenario first.

PR #4143, opened 2023-02-22
chore: ensure timeout and log when ecko pods needs to be scaled down

camilamacedo86:

We add waiting but we do not check if is or not possible to scale down the ekco and properly warning it. 

Motivated By # [sc-69454]

My intention is to address some improvements horizontally.
Therefore, I am breaking it in many PRs so that can make easier the review.


addons/rook/1.10.6/install.sh
```diff
@@ -499,7 +499,9 @@ function rook_scale_down_ekco() {
     kubectl -n kurl scale deployment ekc-operator --replicas=0
     rook_did_scale_down_ekco=1 # local to caller
     log "Waiting for ekco pods to be removed"
-    spinner_until 120 ekco_pods_gone
+    if ! spinner_until 120 ekco_pods_gone; then
```
emosbaugh, line 502: i worry that this effectively changes the logic as spinner_until returns 1 if the command fails until the timeout

emosbaugh (changes requested):

addons/rook/1.10.6/install.sh
camilamacedo86: HI @emosbaugh,

Thank you for your input. Could you please clarify your concern for me?

Today, we face the timeout and we have not the WARN to let us know in this case.
So, the changes addressed here should help us to know that was unable to scale by looking at the logs.

Therefore, what are worried about? We began to log the info to know that happens?

emosbaugh: are you able to reproduce the issue where you received a timeout here? can you send me the output of the kurl installer around these lines of code?

camilamacedo86: I am just trying to ensure consistence. I am not facing the issue.

emosbaugh: ok, but as i see it you are changing the functionality. the function will no longer return 1 and instead return 0 if ekco_pods_gone does not succeed in 120 seconds

emosbaugh: here is a demo
Screen Shot 2023-02-22 at 1 02 25 PM
Screen Shot 2023-02-22 at 1 02 33 PM

camilamacedo86: HI @emosbaugh,

The only change here is : If the condition not be attend, if we face a timeout to check the ekco pods goes down then we will Log the info. We do NOT change the behaviour. 

Your demo above seems to confirmed that it will ONLY log the WARNING and continue. To make that clear I copy and past the same functions from kURL and I make the following test.

```shell
$ cat test.sh 
#!/bin/bash

set -e
set -E # cause 'trap funcname ERR' to be inherited by child commands, see https://stackoverflow.com/questions/35800082/how-to-trap-err-when-using-set-e-in-bash


# Run a test every second with a spinner until it succeeds
function spinner_until() {
    local timeoutSeconds="$1"
    local cmd="$2"
    local args=${@:3}

    if [ -z "$timeoutSeconds" ]; then
        timeoutSeconds=-1
    fi

    local delay=1
    local elapsed=0
    local spinstr='|/-\'

    while ! $cmd $args; do
        elapsed=$((elapsed + delay))
        if [ "$timeoutSeconds" -ge 0 ] && [ "$elapsed" -gt "$timeoutSeconds" ]; then
            return 1
        fi
        local temp=${spinstr#?}
        printf " [%c]  " "$spinstr"
        local spinstr=$temp${spinstr%"$temp"}
        sleep $delay
        printf "\b\b\b\b\b\b"
    done
}

function logWarn() {
    printf "${YELLOW}$1${NC}\n" 1>&2
}

function test(){
   if ! spinner_until 10 ekco_pods_gone; then
      logWarn "Unable to scale down ekco operator"
   fi
}

function ekco_pods_gone() {
    pods_gone_by_selector kurl app=ekc-operator
}

# pods_gone_by_selector returns true if there are no pods matching the given selector
function pods_gone_by_selector() {
    local namespace=$1
    local selector=$2
    [ "$(pod_count_by_selector "$namespace" "$selector")" = "0" ]
}

# pod_count_by_selector returns the number of pods matching the given selector or -1 if the command fails
function pod_count_by_selector() {
    local namespace=$1
    local selector=$2

    local pods=
    if ! pods="$(kubectl -n "$namespace" get pods --no-headers -l "$selector" 2>/dev/null)" ; then
        echo -1
    fi

    echo -n "$pods" | wc -l
}

echo "lets to do the test and it should return WARNING because the ekco pods are running in this host"
test
echo "After return the WARNING the script continue because it does not change anything we are ONLY logging the WARNING after the timeout ..."	
```

**RESULT:  See that we ONLY began to output the info. Just that.** 


Screenshot 2023-02-23 at 09 05 48

camilamacedo86: I see what do you mean. I tested it out either.
To ensure the same behaviour we need to bail.
Currently if we call `spinner_until 120 ekco_pods_gone` and it fails then the script finishes without ANY info.

addons/rook/1.10.6/install.sh
```diff
@@ -499,7 +499,9 @@ function rook_scale_down_ekco() {
...
+       bail "Unable to scale down ekco operator"
```
emosbaugh, line 503: the issue now is that bail will exit 1 which is not the same as return 1 which will allow the caller to choose how to handle the return value

```suggestion
       logFail "Unable to scale down ekco operator"
       return 1
```

emosbaugh (changes requested):

scripts/common/longhorn.sh
```diff
@@ -175,7 +178,9 @@ function remove_longhorn() {
     if kubernetes_resource_exists kurl deployment ekc-operator; then
         kubectl -n kurl scale deploy ekc-operator --replicas=0
         log "Waiting for ekco pods to be removed"
-        spinner_until 120 ekco_pods_gone
+        if ! spinner_until 120 ekco_pods_gone; then
+           logWarn "Unable to scale down ekco operator"
```
emosbaugh, line 182: this is missing return

emosbaugh (changes requested):

scripts/common/longhorn.sh
camilamacedo86, line 182:
```suggestion
           logWarn "Unable to scale down ekco operator"
           return 1
```

scripts/common/longhorn.sh
```diff
@@ -175,7 +178,10 @@ function remove_longhorn() {
...
```
camilamacedo86, line 182:
```suggestion
           logFail "Unable to scale down ekco operator"
```

emosbaugh (changes requested):
you are missing rook 1.10.11

emosbaugh (approved):

PR #4159, opened 2023-02-28
improve 'tasks.sh reset' resilience and clean up more things

laverya:

Fixes #

```release-note
The [reset script](https://kurl.sh/docs/install-with-kurl/managing-nodes#reset-a-node) will now clean up running haproxy processes and handle partial installations with more grace
```

### DIFF
--- a/scripts/tasks.sh
+++ b/scripts/tasks.sh
@@ -210,23 +210,28 @@ function reset() {
 
     if commandExists "kubeadm"; then
         printf "Resetting kubeadm\n"
-        kubeadm_reset
+        kubeadm_reset || true
     fi 
 
     printf "Removing kubernetes packages\n"
     case "$LSB_DIST" in
         ubuntu)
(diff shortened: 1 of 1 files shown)


scripts/tasks.sh
```diff
@@ -210,24 +210,29 @@ function reset() {
...
+            yum remove -y kubernetes-cni kubelet kubectl || true
         ;;
 
         *)
-            echo "Could not uninstall kuberentes host packages on ${LSB_DIST} ${DIST_MAJOR}"
+            echo "Could not uninstall kubernetes host packages on ${LSB_DIST} ${DIST_MAJOR}"
         ;;
     esac
 
+    printf "Potentially cleaning up kubelet service\n"
```
rrpolanco, line 231: why "Potentially"?

emosbaugh (approved):
approved. one question.

why not also remove /var/lib/containerd? do we need to also remove containers? this is not a blocker.

PR #4163, opened 2023-03-01
chore: add better info to no block devices found in Rook preflights

camilamacedo86:

Currently when no block devices are found we only show: `[FAIL] Block Devices: No available block device`
Therefore, this PR improves the info provided and add the link to help the users have further information. 

```release-note
Adds better information to block devices check for Rook versions >= 1.4.3
```

NONE

### DIFF
--- a/addons/rook/1.10.11/host-preflight.yaml
+++ b/addons/rook/1.10.11/host-preflight.yaml
@@ -24,7 +24,7 @@ spec:
           when: "{{kurl if (and .Installer.Spec.Rook.Version .Installer.Spec.Rook.BlockDeviceFilter) }}{{kurl .Installer.Spec.Rook.BlockDeviceFilter }}{{kurl else }}.*{{kurl end }} > 1"
           message: Multiple available block devices
       - fail:
-          message: No available block devices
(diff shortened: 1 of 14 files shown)


addons/rook/1.10.11/host-preflight.yaml
```diff
@@ -24,7 +24,7 @@ spec:
           when: "{{kurl if (and .Installer.Spec.Rook.Version .Installer.Spec.Rook.BlockDeviceFilter) }}{{kurl .Installer.Spec.Rook.BlockDeviceFilter }}{{kurl else }}.*{{kurl end }} > 1"
           message: Multiple available block devices
       - fail:
-          message: No available block devices
+          message: "No available block devices unformatted and dedicated to be use by Rook only were found attached to each node in the cluster. For further information see: https://kurl.sh/docs/add-ons/rook#block-storage"
```
laverya, line 27: >No available block devices unformatted and dedicated to be use by Rook only were found attached to each node in the cluster.

This is IMO not clear for a few reasons

1. "each node in the cluster" - this check only runs on the first node, and so this message is redundant
2. "dedicated to be use by Rook only" - this seems like it might be a source of confusion. How does one designate this? (they don't, but that's not clear)

I would instead say:
```suggestion
          message: "No available unformatted block devices were found, and Rook requires one. For further information see: https://kurl.sh/docs/add-ons/rook#block-storage"
```

emosbaugh (approved):

PR #4165, opened 2023-03-01
fix: print colours

camilamacedo86:

Currently we are not print the warnings, failures and info with colours which make harder to identify what are the scenarios. Therefore, this PR has the purpose to sort it out. 

Fixes #

```release-note

```

### DIFF
--- a/pkg/cli/host_preflight.go
+++ b/pkg/cli/host_preflight.go
@@ -9,6 +9,7 @@ import (
 	"time"
 
 	"github.com/briandowns/spinner"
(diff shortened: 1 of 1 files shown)


pkg/cli/host_preflight.go
```diff
@@ -9,7 +9,6 @@ import (
 	"time"
 
 	"github.com/briandowns/spinner"
```
camilamacedo86, line 11:
```suggestion
	"github.com/briandowns/spinner"
	"github.com/mattn/go-isatty"
```

PR #4173, opened 2023-03-06
fix: host preflight checks results output colored

camilamacedo86:

Currently, because of the usage of the tee to generate the logs we are unable to check the host preflights results with colours which makes harder to identify what is an error, warning and etc. 

Fixes #

See:

Screenshot 2023-03-06 at 12 30 12

## Steps to reproduce
Just run the install command and check the host preflights checks result. 

```release-note
fix: host preflight checks result by outputting them with colours in order to improve the user experience
```

NONE

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -10,7 +10,6 @@ require (
 	github.com/chzyer/readline v1.5.1
 	github.com/denisbrodbeck/machineid v1.0.1
 	github.com/docker/distribution v2.8.1+incompatible
-	github.com/fatih/color v1.14.1
 	github.com/foomo/htpasswd v0.0.0-20200116085101-e3a90e78da9c
 	github.com/golang/mock v1.6.0
 	github.com/google/go-cmp v0.5.9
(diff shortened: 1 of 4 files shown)


pkg/cli/host_preflight.go
```diff
@@ -137,7 +137,7 @@ func newHostPreflightCmd(cli CLI) *cobra.Command {
 				return errors.Wrap(err, "run host preflight")
 			}
 
-			printPreflightResults(cmd.OutOrStdout(), results)
```
laverya, line 140: why the change to no longer use outorstdout?

I feel like this is the change that's actually making a difference

camilamacedo86: Yep, I agree I reverted that.

pkg/cli/host_preflight_test.go
```diff
@@ -45,7 +45,7 @@ func TestNewHostPreflightCmd(t *testing.T) {
 					IsPass:  true,
 				},
 			},
-			stdout: green("[PASS]") + " Number of CPUs: At least 4 CPU cores are required\n",
+			stdout: OutputPassGreen() + " Number of CPUs: At least 4 CPU cores are required\n",
```
laverya, line 48: I would GREATLY prefer for this not to call `OutputPassGreen` etc

laverya: A() = A() isn't a good unit test IMO

camilamacedo86: That is how that was before. I think has no point to change the tests now.

laverya (approved):

go.mod
```diff
@@ -319,6 +319,7 @@ replace (
 	k8s.io/controller-manager => k8s.io/controller-manager v0.26.1
 	k8s.io/cri-api => k8s.io/cri-api v0.26.1
 	k8s.io/csi-translation-lib => k8s.io/csi-translation-lib v0.26.1
+	k8s.io/dynamic-resource-allocation => k8s.io/dynamic-resource-allocation v0.26.1
```
camilamacedo86, line 322:
```suggestion
```

PR #4177, opened 2023-03-07
fix colour to output failures for host packages checks

camilamacedo86:

The failures has been outputted with yellow and not red.  This PR solves that.

Screenshot 2023-03-07 at 09 08 27

Fixes # [sc-70924]


emosbaugh (approved):

camilamacedo86:
FOSSA check is facing an network issue.
However, we did not add any new dependency in this PR so has none new dependency license to be checked. 
In this way, for we not get blocked and be able to do the release within I am moving forward since the FOSSA check is getting stuck.

PR #4180, opened 2023-03-07
fix: only allow users to move from Rook to OpenEBS >= 3.3.0

camilamacedo86:

Only OpenEBS versions >= 3.3.0 can support the migrating path from Rook. 
Therefore, this PR adds a preflight check that will bail in case users try to move with this upgrade path

Fixes # [sc-70256]

See: 

Screenshot 2023-03-08 at 00 19 29

```release-note
Fixes unsupported migration path from Rook to OpenEBS with versions < 3.3.0 by adding a preflight check to prevent the scenario. 
```

NONE

### DIFF
--- a/scripts/common/preflights.sh
+++ b/scripts/common/preflights.sh
@@ -640,7 +640,9 @@ allow_remove_docker_new_install() {
      fi
 }
 
-allow_migrate_from_rook_to_openebs() {
+# bail_if_unsupported_openebs_to_rook_version will bail if the rook is being removed in favor of
+# openebs and the openebs version does not support migrations from rook.
+function bail_if_unsupported_openebs_to_rook_version() {
     if [ -z "$ROOK_VERSION" ] && [ -n "$OPENEBS_VERSION" ]; then
         if commandExists kubectl; then
             if kubectl get ns | grep -q rook-ceph; then
--- a/scripts/common/preflights.sh
+++ b/scripts/common/preflights.sh
@@ -644,7 +644,9 @@ allow_migrate_from_rook_to_openebs() {
     if [ -z "$ROOK_VERSION" ] && [ -n "$OPENEBS_VERSION" ]; then
         if commandExists kubectl; then
             if kubectl get ns | grep -q rook-ceph; then
-                if [ $(echo "$OPENEBS_VERSION" | cut -d. -f1) == 2 ] || [ $(echo "$OPENEBS_VERSION" | cut -d. -f1) == 1 ] || [ "$OPENEBS_VERSION" == "3.2.0" ]; then
+                semverParse "$OPENEBS_VERSION"
+                # if $OPENEBS_VERSION is less than 3.3.0
+                if [ "$major" -lt "3" ] || { [ "$major" = "3" ] && [ "$minor" -lt "3" ] ; }; then
                    logFail "The OpenEBS version $OPENEBS_VERSION cannot be installed"
                    bail "OpenEBS versions <= 3.2.0 does not support migration from Rook"
                 fi
--- a/scripts/common/preflights.sh
+++ b/scripts/common/preflights.sh
@@ -644,7 +644,7 @@ allow_migrate_from_rook_to_openebs() {
     if [ -z "$ROOK_VERSION" ] && [ -n "$OPENEBS_VERSION" ]; then
         if commandExists kubectl; then
             if kubectl get ns | grep -q rook-ceph; then
-                if  [ "$(echo "$OPENEBS_VERSION}" | wc -l)" -le "3" ] || [ "$OPENEBS_VERSION" == "3.2.0"]; then
+                if [ $(echo "$OPENEBS_VERSION" | cut -d. -f1) == 2 ] || [ $(echo "$OPENEBS_VERSION" | cut -d. -f1) == 1 ] || [ "$OPENEBS_VERSION" == "3.2.0" ]; then
                    logFail "The OpenEBS version $OPENEBS_VERSION cannot be installed"
                    bail "OpenEBS versions <= 3.2.0 does not support migration from Rook"
                 fi
--- a/scripts/common/preflights.sh
+++ b/scripts/common/preflights.sh
@@ -13,6 +13,7 @@ function preflights() {
     kotsadm_prerelease
     host_nameservers_reachable
     allow_remove_docker_new_install
+    allow_migrate_from_rook_to_openebs
     return 0
 }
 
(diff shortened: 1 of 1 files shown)


scripts/common/preflights.sh
```diff
@@ -638,3 +639,13 @@ allow_remove_docker_new_install() {
           logWarn "conflicts, please uninstall Docker and re-run the install script."
      fi
 }
+
+allow_migrate_from_rook_to_openebs() {
```
emosbaugh, line 643: following existing patterns, this function belongs in the openebs_pre_init functions of each openebs add-on version. do you have a reason for putting it here instead?

scripts/common/preflights.sh
```diff
@@ -638,3 +639,13 @@ allow_remove_docker_new_install() {
...
+    if [ -z "$ROOK_VERSION" ] && [ -n "$OPENEBS_VERSION" ]; then
+        if kubectl get ns | grep -q rook-ceph; then
+            if  [ "$(echo "$OPENEBS_VERSION}" | wc -l)" -le "3" ] || [ "$OPENEBS_VERSION" == "3.2.0"]; then
```
emosbaugh, line 646: what about 1.x and 2.x versions?

scripts/common/preflights.sh
camilamacedo86: If we add to each addon version that does not mean that it will ONLY fail when we try to add the addon?
So, should we allow apply the previous addons and etc prior fail?

scripts/common/preflights.sh
```diff
@@ -638,3 +639,13 @@ allow_remove_docker_new_install() {
...
+                  bail "OpenEBS versions <= 3.2.0 does not support migration from Rook"
```
camilamacedo86, line 647:
```suggestion
                  logFail "The OpenEBS version $OPENEBS_VERSION cannot be installed"
                  bail "OpenEBS versions <= 3.2.0 does not support migration from Rook"
```

scripts/common/preflights.sh
laverya: not having to change every old addon version, maybe?

scripts/common/preflights.sh
camilamacedo86: This code will bail when OpenEBS version is == 3.2.0 and lower.
I improve the log and just did another test to confirm:

Screenshot 2023-03-07 at 22 01 27

So, what do you mean with 1.x and 2.x? If they choose those then the installer will also fail.

emosbaugh: this looks like a typo with an extra `}` `"$(echo "$OPENEBS_VERSION}" | wc -l)" -le "3"`. I have a feeling its not bailing for the reason you think it is.

scripts/common/preflights.sh
camilamacedo86: But I found a problem we need to check if we have kubectl first as well I am adding it.

scripts/common/preflights.sh
```diff
@@ -638,3 +639,17 @@ allow_remove_docker_new_install() {
           logWarn "conflicts, please uninstall Docker and re-run the install script."
      fi
 }
+
+allow_migrate_from_rook_to_openebs() {
+    if [ -z "$ROOK_VERSION" ] && [ -n "$OPENEBS_VERSION" ]; then
+        if commandExists kubectl; then
+            if kubectl get ns | grep -q rook-ceph; then
+                if  [ "$(echo "$OPENEBS_VERSION}" | wc -l)" -le "3" ] || [ "$OPENEBS_VERSION" == "3.2.0"]; then
```
emosbaugh, line 647:
```
[ethan@ethanm-rhel9-1 ~]$ OPENEBS_VERSION=3.3.0
[ethan@ethanm-rhel9-1 ~]$ echo "$OPENEBS_VERSION}" | wc -l
1
```

emosbaugh: this needs to be fixed

emosbaugh (changes requested):

scripts/common/preflights.sh
camilamacedo86: You are shell master. Yep. You are right. I did all tests and I fixed it.

scripts/common/preflights.sh
```diff
@@ -638,3 +639,17 @@ allow_remove_docker_new_install() {
           logWarn "conflicts, please uninstall Docker and re-run the install script."
      fi
 }
+
+allow_migrate_from_rook_to_openebs() {
+    if [ -z "$ROOK_VERSION" ] && [ -n "$OPENEBS_VERSION" ]; then
+        if commandExists kubectl; then
+            if kubectl get ns | grep -q rook-ceph; then
+                if [ $(echo "$OPENEBS_VERSION" | cut -d. -f1) == 2 ] || [ $(echo "$OPENEBS_VERSION" | cut -d. -f1) == 1 ] || [ "$OPENEBS_VERSION" == "3.2.0" ]; then
```
emosbaugh, line 647: could you use the built in semverParse function? this seems more readable

```suggestion
                semverParse "$OPENEBS_VERSION"
                # if $OPENEBS_VERSION is less than 3.3.0
                if [ "$major" -lt "3" ] || { [ "$major" = "3" ] && [ "$minor" -lt "3" ] ; }; then
```

emosbaugh: another option is semverCompare

```suggestion
                semverCompare "$OPENEBS_VERSION" "3.3.0"
                # if $OPENEBS_VERSION is less than 3.3.0
                if [ "$SEMVER_COMPARE_RESULT" = "-1" ]; then
```

scripts/common/preflights.sh
```diff
@@ -638,3 +639,17 @@ allow_remove_docker_new_install() {
           logWarn "conflicts, please uninstall Docker and re-run the install script."
      fi
 }
+
+allow_migrate_from_rook_to_openebs() {
```
emosbaugh, line 643:
```suggestion
# bail_if_unsupported_openebs_to_rook_version will bail if the rook is being removed in favor of
# openebs and the openebs version does not support migrations from rook.
function bail_if_unsupported_openebs_to_rook_version() {
```

scripts/common/preflights.sh
```diff
@@ -13,6 +13,7 @@ function preflights() {
     kotsadm_prerelease
     host_nameservers_reachable
     allow_remove_docker_new_install
+    allow_migrate_from_rook_to_openebs
```
emosbaugh, line 16:
```suggestion
    bail_if_unsupported_openebs_to_rook_version
```

scripts/common/preflights.sh
```diff
@@ -638,3 +639,17 @@ allow_remove_docker_new_install() {
...
      fi
 }
+
+allow_migrate_from_rook_to_openebs() {
+    if [ -z "$ROOK_VERSION" ] && [ -n "$OPENEBS_VERSION" ]; then
+        if commandExists kubectl; then
+            if kubectl get ns | grep -q rook-ceph; then
+                if [ $(echo "$OPENEBS_VERSION" | cut -d. -f1) == 2 ] || [ $(echo "$OPENEBS_VERSION" | cut -d. -f1) == 1 ] || [ "$OPENEBS_VERSION" == "3.2.0" ]; then
+                   logFail "The OpenEBS version $OPENEBS_VERSION cannot be installed"
+                   bail "OpenEBS versions <= 3.2.0 does not support migration from Rook"
```
emosbaugh, line 649:
```suggestion
                   logFail "The OpenEBS version $OPENEBS_VERSION cannot be installed."
                   bail "OpenEBS versions less than 3.3.0 do not support migrations from Rook"
```

emosbaugh (changes requested):

camilamacedo86:
All your suggestions are accepted @emosbaugh.
Also, I re-tested it out and the result is the description.

It seems good to fly now.

emosbaugh (approved):

PR #4181, opened 2023-03-08
Fixes issue `failed to find plugin /opt/cni/bin/weave-net`

camilamacedo86:

Because in some scenario the weave-net cni binary is not found:
223588364-4751434a-4aad-4f7e-9a3c-242f55932e36

Therefore, we could check that by removing the pod and let be re-created we can fix it 

Screenshot 2023-03-08 at 09 54 19

Fixes # [sc-71175]

I am unable to reproduce the issue but I tried to make it happens to verify this solution by
removing the removing the weave-net binary prior check if the nodes are ready. See:

Screenshot 2023-03-08 at 10 59 57

Then, we can check that the installation succeed as well:

Screenshot 2023-03-08 at 10 59 27

Installer used : https://kurl.sh/9f773aa
SO: ubuntu 20.04 - GCP
## Steps to reproduce

In the script prior check the network we are removing the bin to simulate the scenario. 

```release-note
Fixes issue `failed to find plugin /opt/cni/bin/weave-net` when the installer is checking cluster networking by deleting the weave-net pod when the binary is not found to let it be re-created successfully  
```

NONE

### DIFF
--- a/scripts/common/kubernetes.sh
+++ b/scripts/common/kubernetes.sh
@@ -784,7 +784,7 @@ function kubeadm_cluster_status() {
 }
 
 function check_network() {
-	  logStep "Checking cluster networking"
+    logStep "Checking cluster networking"
 
     if [ -n "$WEAVE_VERSION" ]; then
         log "Checking if weave-net binary can be found in the path /opt/cni/bin/"
--- a/scripts/common/kubernetes.sh
+++ b/scripts/common/kubernetes.sh
@@ -784,7 +784,15 @@ function kubeadm_cluster_status() {
 }
 
 function check_network() {
-	logStep "Checking cluster networking"
+	  logStep "Checking cluster networking"
+
+    if [ -n "$WEAVE_VERSION" ]; then
+        log "Checking if weave-net binary can be found in the path /opt/cni/bin/"
+        if ! ls -la /opt/cni/bin/ | grep weave-net; then
+            logWarn "Unable to find weave-net binary. Deleting weave-net pod"
+            kubectl delete pods --selector=name=weave-net -n kube-system --ignore-not-found=true
+        fi
+    fi
 
     if ! kubernetes_any_node_ready; then
         echo "Waiting for node to report Ready"
--- a/scripts/common/kubernetes.sh
+++ b/scripts/common/kubernetes.sh
@@ -790,7 +790,7 @@ function check_network() {
         log "Checking if weave-net binary can be found in the path /opt/cni/bin/"
         if ! ls -la /opt/cni/bin/ | grep weave-net; then
             logWarn "Unable to find weave-net binary, deleting weave-net pod so that the binary will be recreated"
-            kubectl delete pods --selector=name=weave-net -n kube-system --ignore-not-found=true
+            kubectl delete pods --selector=name=weave-net --field-selector=spec.nodeName=$(hostname) -n kube-system --ignore-not-found=true
         fi
     fi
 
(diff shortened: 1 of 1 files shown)


scripts/common/kubernetes.sh
```diff
@@ -784,7 +784,15 @@ function kubeadm_cluster_status() {
...
 
 function check_network() {
-	logStep "Checking cluster networking"
+	  logStep "Checking cluster networking"
+
+    if [ -n "$WEAVE_VERSION" ]; then
+        log "Checking if weave-net binary can be found in the path /opt/cni/bin/"
+        if ! ls -la /opt/cni/bin/ | grep weave-net; then
+            logWarn "Unable to find weave-net binary. Deleting weave-net pod"
+            kubectl delete pods --selector=name=weave-net -n kube-system --ignore-not-found=true
```
laverya, line 793: Would it be possible to find the pod on this node in particular and delete that one?

`kubectl get pods -o wide -n kube-system -l name=weave-net | grep $(hostname)` or similar

scripts/common/kubernetes.sh
```diff
@@ -784,7 +784,15 @@ function kubeadm_cluster_status() {
 }
 
 function check_network() {
-	logStep "Checking cluster networking"
+	  logStep "Checking cluster networking"
+
+    if [ -n "$WEAVE_VERSION" ]; then
+        log "Checking if weave-net binary can be found in the path /opt/cni/bin/"
+        if ! ls -la /opt/cni/bin/ | grep weave-net; then
+            logWarn "Unable to find weave-net binary. Deleting weave-net pod"
```
laverya, line 792:
```suggestion
            logWarn "Unable to find weave-net binary, deleting weave-net pod so that the binary will be recreated"
```

scripts/common/kubernetes.sh
```diff
@@ -784,7 +784,15 @@ function kubeadm_cluster_status() {
 }
 
 function check_network() {
-	logStep "Checking cluster networking"
+	  logStep "Checking cluster networking"
```
laverya, line 787:
```suggestion
    logStep "Checking cluster networking"
```

scripts/common/kubernetes.sh
camilamacedo86: Are you suggesting it because of the `--ignore-not-found=true` usage? 
If so, I just it in case the pod (not created yet) be not found we still have a chance by re-trying to validate the network again. However, we could remove it. It has not a real reason to be there.

laverya: No, I'm saying it because you're deleting all of the weave pods on all of the nodes when only the current node has a problem.

camilamacedo86: Hun, by default this scenario should not occur and we will not delete the pods at all. If we check that the binaries in some edge case is not present I think the best approach is to delete the pods in all nodes since it can ensure that they will re-init, as we can check that adds the binaries in the required path and the config should be ok.

laverya: I would disagree - I think we should only delete it for the current node. This way, we won't break a currently-running cluster - only one node at a time.

scripts/common/kubernetes.sh
```diff
@@ -784,7 +784,15 @@ function kubeadm_cluster_status() {
...
 
 function check_network() {
-	logStep "Checking cluster networking"
+    logStep "Checking cluster networking"
+
+    if [ -n "$WEAVE_VERSION" ]; then
+        log "Checking if weave-net binary can be found in the path /opt/cni/bin/"
+        if ! ls -la /opt/cni/bin/ | grep weave-net; then
+            logWarn "Unable to find weave-net binary, deleting weave-net pod so that the binary will be recreated"
+            kubectl delete pods --selector=name=weave-net -n kube-system --ignore-not-found=true
```
camilamacedo86, line 793:
```suggestion
            kubectl delete pods --selector=name=weave-net --field-selector=spec.nodeName=$(hostname) -n kube-system --ignore-not-found=true
```

laverya (approved):

PR #4189, opened 2023-03-10
fix: Check for correct flannel image when migrating from Weave

rrpolanco:

The `weave-to-flannel-primary`/`weave-to-flannel-secondary` tasks are failing due to the flannel images being "missing":
```
⚙  Running tasks with the argument(s): weave-to-flannel-primary airgap cert-key=0e2bfbb240ce8436822ca27a545b635d12c979531b9c606f6380436f927f82a0
Flannel images not present on rafael-weave-to-flannel-airgap-cp-3, please ensure the 'load-images' task has been run successfully
```
However, the flannel images are present (and loaded) its just that the logic for checking whether the flannel image was present was searching for the incorrect image: `docker.io/rancher/mirrored-flannelcni-flannel`. The Flannel project changed the registry and repository for the image in this [commit](https://github.com/flannel-io/flannel/commit/3b242e9e7d72ff444c8d56566e3f2e75eb64d1e8).

Fixes #

[sc-70618](https://app.shortcut.com/replicated/story/70618/weave-to-flannel-primary-task-on-airgap-node-fails-with-missing-images)

```release-note
Fixes an issue where the weave-to-flannel-{primary,secondary} tasks fail with "Flannel images not present..."
```

### DIFF
--- a/scripts/tasks.sh
+++ b/scripts/tasks.sh
@@ -855,9 +855,9 @@ function task_requires_root() {
     fi
 }
 
-# check if containerd on the current node has the `docker.io/rancher/mirrored-flannelcni-flannel:v<version>` image
+# check if containerd on the current node has the `docker.io/flannel/flannel` image
 function flannel_images_present() {
-    if ! ctr -n=k8s.io images ls | grep -q "docker.io/rancher/mirrored-flannelcni-flannel" ; then
+    if ! ctr -n=k8s.io images ls | grep -q "docker.io/flannel/flannel" ; then
         logFail "Flannel images not present on $(get_local_node_name), please ensure the 'load-images' task has been run successfully"
         exit 1
     fi


scripts/tasks.sh
```diff
@@ -855,9 +855,9 @@ function task_requires_root() {
     fi
 }
 
-# check if containerd on the current node has the `docker.io/rancher/mirrored-flannelcni-flannel:v<version>` image
+# check if containerd on the current node has the `docker.io/flannel/flannel` image
 function flannel_images_present() {
-    if ! ctr -n=k8s.io images ls | grep -q "docker.io/rancher/mirrored-flannelcni-flannel" ; then
+    if ! ctr -n=k8s.io images ls | grep -q "docker.io/flannel/flannel" ; then
```
laverya, line 860: can we check for either of the two images instead?

laverya: that way we don't break this migration for previous flannel versions

rrpolanco: Good point. If you're migrating to flannel versions `0.20.0`, `0.20.1`,  and `0.20.2` then those use the previous image name.

scripts/tasks.sh
```diff
@@ -784,7 +784,7 @@ function weave_to_flannel_primary() {
     rm /tmp/kubeadm-token
 
     cat > /tmp/kubeadm-join.conf <<- EOM
-apiVersion: kubeadm.k8s.io/v1beta2
+apiVersion: kubeadm.k8s.io/v1beta3
```
emosbaugh, line 787: i dont think we should change this as its not backwards compatible with older kubeadm versions.

scripts/tasks.sh
```diff
@@ -855,9 +855,9 @@ function task_requires_root() {
     fi
 }
 
-# check if containerd on the current node has the `docker.io/rancher/mirrored-flannelcni-flannel:v<version>` image
+# check if containerd on the current node has the `docker.io/flannel/flannel` image
 function flannel_images_present() {
-    if ! ctr -n=k8s.io images ls | grep -q "docker.io/rancher/mirrored-flannelcni-flannel" ; then
+    if ! ctr -n=k8s.io images ls | grep -Eq "docker.io/flannel/flannel|docker.io/rancher/mirrored-flannelcni-flannel" ; then
```
emosbaugh, line 860:
```suggestion
    if ! ctr -n=k8s.io images ls | grep -Eq "docker\.io/flannel/flannel|docker\.io/rancher/mirrored-flannelcni-flannel" ; then
```

emosbaugh (changes requested):

scripts/tasks.sh
rrpolanco: What's the minimum supported version for kubeadm? `v1beta3` was introduced in [Kubernetes 1.22](https://kubernetes.io/blog/2021/08/04/kubernetes-1-22-release-announcement/#more-secure-control-plane-with-kubeadm).

emosbaugh: Regardless this seems unrelated to your change and I'm not sure the kustomize patches will apply if we don't use the same api version consistently

emosbaugh (approved):

scripts/tasks.sh
rrpolanco: Ok. I still think we should consider cleaning up the deprecation warnings as part of separate effort

PR #4190, opened 2023-03-11
chore: began to store and retrieve DID_MIGRATE_ROOK_PVCS and DID_MIGRATE_ROOK_OBJECT_STORE in kurl-last-config configmap

camilamacedo86:

Because we have migrations operations where is required to check if DID_MIGRATE_ROOK_PVCS and DID_MIGRATE_ROOK_OBJECT_STORE are ==1. Therefore, if some error occurs and the script is re-executed the value is lost.

This is the first step to sort out the problems:
- We should not migrate more than once the the object store [sc-70240]
- We should be able to delete rook in all scenarios (such as when a timeout is faced and we must to re-run the script)

Also, it will be helpful for troobolshoting and support case scenarios since we will be able to check in the kurl-last-config configmap what was or not migrated.  

Related # [sc-70239]

In order to make easier the review this PR has the purpose of ONLY start to use the configmap instead of envvar s to store and retrieve this info. 

```release-note
Adds status of Rook PVC and Object Store migration to the kurl-last-config configmap
```

NONE


emosbaugh (commented):
Shouldnt you patch the configmap and remove DID_MIGRATE_ROOK_PVCS and DID_MIGRATE_ROOK_OBJECT_STORE once the script finishes?

camilamacedo86:
Hi @emosbaugh,

Thank you for looking on that.

> Shouldnt you patch the configmap and set DID_MIGRATE_ROOK_PVCS and DID_MIGRATE_ROOK_OBJECT_STORE to 0 once the script finishes?

If we change the status to 0 and we need to check if the migration was not completed succefully according to the script in a support case or in troubleshooting, how do we will know that? 

In this way, I do not think so. If we migration in the latest config then it should be 1.

emosbaugh:
> Hi @emosbaugh,
> 
> Thank you for looking on that.
> 
> > Shouldnt you patch the configmap and set DID_MIGRATE_ROOK_PVCS and DID_MIGRATE_ROOK_OBJECT_STORE to 0 once the script finishes?
> 
> If we change the status to 0 and we need to check if the migration was not completed succefully according to the script in a support case or in troubleshooting, how will we know that?
> 
> In this way, I do not think so. If we migrate in the latest config then it should be 1.
> 
> Also, the goal of this one is ONLY to replace the ENVAR by have the data storage in the config map. So that we should ensure that the testgrids runs fine after it and then, continue to work via follow up us to address the required fixed.
> 
> Can we move forward here? Have we your LGTM in this one?

But could you toggle the values prior to maybe_cleanup_rook? This would address "We should not migrate more than once the the object store [sc-70240]".

You could then clean up rook if there is a namespace and ROOK_VERSION is empty.

camilamacedo86:
Hi @emosbaugh 

> But could you toggle the values prior to maybe_cleanup_rook? This would address "We should not migrate more than once the the object store [sc-70240]".
> You could then clean up rook if there is a namespace and ROOK_VERSION is empty.

This PR does not address [sc-70240]. It is a pre-requirement to achieve the goal of [sc-70240]
The PR that will address that is https://github.com/replicatedhq/kURL/pull/4197 (DRAFT)

To not let the object store be migrated twice we must have the info stored persistently (in the config map first) which is the goal of this PR. 

Sorry, but I can not follow your comment suggestion on this one.

scripts/common/rook.sh
```diff
@@ -207,11 +207,23 @@ function rook_ceph_to_sc_migration() {
 # if PVCs and object store data have both been migrated from rook-ceph and rook-ceph is no longer specified in the kURL spec, remove rook-ceph
 function maybe_cleanup_rook() {
     if [ -z "$ROOK_VERSION" ]; then
+        if ! kubectl get ns | grep -q rook-ceph; then
+           return
+        fi
+        logStep "Removing Rook"
```
emosbaugh, line 213: can you move this into the conditional below line 216?

addons/ekco/template/base/install.sh
```diff
@@ -120,6 +120,7 @@ function ekco_already_applied() {
     local dst="$DIR/kustomize/ekco"
 
     # if rook-ceph has been removed, ekco should be redeployed to not attempt to manage it
+    DID_MIGRATE_ROOK_PVCS=$(kubectl -n kurl get configmap kurl-migration-from-rook -o yaml | grep DID_MIGRATE_ROOK_PVCS | head -1 | awk '{print $2}' | sed 's/\"//g')
```
emosbaugh, line 123:
```suggestion
    DID_MIGRATE_ROOK_PVCS=$(kubectl -n kurl get configmap kurl-migration-from-rook -o jsonpath='{ .data.DID_MIGRATE_ROOK_PVCS }')
```

emosbaugh (changes requested):

scripts/common/rook.sh
```diff
@@ -246,3 +258,21 @@ function rook_operator_ready() {
...
+}
+
+# Check if the kurl-migration-from-rook exists then, if not creates it
+# To add DID_MIGRATE_ROOK_PVCS = "1" in order to track that the PVCs were migrated
+function add_rook_store_object_migration_status() {
+    if ! kubectl -n kurl get configmap kurl-migration-status; then
+       kubectl create configmap kurl-migration-from-rook -n kurl
+    fi
+    kubectl patch configmap kurl-migration-from-rook -n kurl --type merge -p '{"data":{"DID_MIGRATE_ROOK_OBJECT_STORE":"1"}}'
+}
```
laverya, line 278: Should we make these two still export the environment variables so that we don't need to change ekco everywhere?

scripts/common/rook.sh
```diff
@@ -207,11 +207,24 @@ function rook_ceph_to_sc_migration() {
...
             report_addon_start "rook-ceph-removal" "v1"
+            kubectl patch configmap kurl-migration-from-rook -n kurl --type merge -p '{"data":{"DID_MIGRATE_ROOK_OBJECT_STORE":"completed"}}'
+            kubectl patch configmap kurl-migration-from-rook -n kurl --type merge -p '{"data":{"DID_MIGRATE_ROOK_PVCS":"completed"}}'
             remove_rook_ceph
+            kubectl patch configmap kurl-migration-from-rook -n kurl --type merge -p '{"data":{"DID_REMOVE_ROOK":"completed"}}'
             report_addon_success "rook-ceph-removal" "v1"
+            logSuccess "Rook removed from cluster"
+            return
         fi
+        logWarn "Unable to remove Rook."
```
camilamacedo86, line 227:
```suggestion
```

scripts/common/rook.sh
```diff
@@ -207,10 +207,22 @@ function rook_ceph_to_sc_migration() {
 # if PVCs and object store data have both been migrated from rook-ceph and rook-ceph is no longer specified in the kURL spec, remove rook-ceph
 function maybe_cleanup_rook() {
     if [ -z "$ROOK_VERSION" ]; then
+        if ! kubectl get ns | grep -q rook-ceph; then
+           return
+        fi
+        logStep "Removing Rook"
```
camilamacedo86, line 213:
```suggestion
```

scripts/common/rook.sh
```diff
@@ -207,10 +207,18 @@ function rook_ceph_to_sc_migration() {
 # if PVCs and object store data have both been migrated from rook-ceph and rook-ceph is no longer specified in the kURL spec, remove rook-ceph
 function maybe_cleanup_rook() {
     if [ -z "$ROOK_VERSION" ]; then
+        ```suggestion
```
camilamacedo86, line 210:
```suggestion
```

scripts/common/rook.sh
```diff
@@ -207,10 +207,17 @@ function rook_ceph_to_sc_migration() {
...
+        DID_MIGRATE_ROOK_OBJECT_STORE=$(kubectl -n kurl get configmap kurl-migration-from-rook -o jsonpath='{ .data.DID_MIGRATE_ROOK_OBJECT_STORE }')
         if [ "$DID_MIGRATE_ROOK_PVCS" == "1" ] && [ "$DID_MIGRATE_ROOK_OBJECT_STORE" == "1" ]; then
             report_addon_start "rook-ceph-removal" "v1"
+            kubectl patch configmap kurl-migration-from-rook -n kurl --type merge -p '{"data":{"DID_MIGRATE_ROOK_OBJECT_STORE":"completed"}}'
+            kubectl patch configmap kurl-migration-from-rook -n kurl --type merge -p '{"data":{"DID_MIGRATE_ROOK_PVCS":"completed"}}'
             remove_rook_ceph
+            kubectl patch configmap kurl-migration-from-rook -n kurl --type merge -p '{"data":{"DID_REMOVE_ROOK":"completed"}}'
             report_addon_success "rook-ceph-removal" "v1"
+            logSuccess "Rook removed from cluster"
+            return
```
camilamacedo86, line 220:
```suggestion
```

scripts/common/rook.sh
```diff
@@ -207,10 +207,17 @@ function rook_ceph_to_sc_migration() {
...
+        DID_MIGRATE_ROOK_PVCS=$(kubectl -n kurl get configmap kurl-migration-from-rook -o jsonpath='{ .data.DID_MIGRATE_ROOK_PVCS }')
+        DID_MIGRATE_ROOK_OBJECT_STORE=$(kubectl -n kurl get configmap kurl-migration-from-rook -o jsonpath='{ .data.DID_MIGRATE_ROOK_OBJECT_STORE }')
         if [ "$DID_MIGRATE_ROOK_PVCS" == "1" ] && [ "$DID_MIGRATE_ROOK_OBJECT_STORE" == "1" ]; then
             report_addon_start "rook-ceph-removal" "v1"
+            kubectl patch configmap kurl-migration-from-rook -n kurl --type merge -p '{"data":{"DID_MIGRATE_ROOK_OBJECT_STORE":"completed"}}'
+            kubectl patch configmap kurl-migration-from-rook -n kurl --type merge -p '{"data":{"DID_MIGRATE_ROOK_PVCS":"completed"}}'
             remove_rook_ceph
+            kubectl patch configmap kurl-migration-from-rook -n kurl --type merge -p '{"data":{"DID_REMOVE_ROOK":"completed"}}'
             report_addon_success "rook-ceph-removal" "v1"
+            logSuccess "Rook removed from cluster"
```
camilamacedo86, line 219:
```suggestion
```

scripts/common/rook.sh
camilamacedo86: Interesting point. Are we using the ENVVARs there I will give a look. THANKS

emosbaugh: great call

camilamacedo86: I checked and ecko itself does not use DID_MIGRATE_ROOK_PVCS. 
So, the changes here are fine.

scripts/common/rook.sh
```diff
@@ -207,9 +207,14 @@ function rook_ceph_to_sc_migration() {
 # if PVCs and object store data have both been migrated from rook-ceph and rook-ceph is no longer specified in the kURL spec, remove rook-ceph
 function maybe_cleanup_rook() {
     if [ -z "$ROOK_VERSION" ]; then
+        DID_MIGRATE_ROOK_PVCS=$(kubectl -n kurl get configmap kurl-migration-from-rook -o jsonpath='{ .data.DID_MIGRATE_ROOK_PVCS }')
+        DID_MIGRATE_ROOK_OBJECT_STORE=$(kubectl -n kurl get configmap kurl-migration-from-rook -o jsonpath='{ .data.DID_MIGRATE_ROOK_OBJECT_STORE }')
         if [ "$DID_MIGRATE_ROOK_PVCS" == "1" ] && [ "$DID_MIGRATE_ROOK_OBJECT_STORE" == "1" ]; then
             report_addon_start "rook-ceph-removal" "v1"
+            kubectl patch configmap kurl-migration-from-rook -n kurl --type merge -p '{"data":{"DID_MIGRATE_ROOK_OBJECT_STORE":"completed"}}'
+            kubectl patch configmap kurl-migration-from-rook -n kurl --type merge -p '{"data":{"DID_MIGRATE_ROOK_PVCS":"completed"}}'
```
camilamacedo86, line 215:
```suggestion
```

scripts/common/rook.sh
```diff
@@ -207,9 +207,14 @@ function rook_ceph_to_sc_migration() {
...
             remove_rook_ceph
+            kubectl patch configmap kurl-migration-from-rook -n kurl --type merge -p '{"data":{"DID_REMOVE_ROOK":"completed"}}'
```
camilamacedo86, line 217:
```suggestion
            kubectl delete configmap kurl-last-config -n kurl
```

scripts/common/rook.sh
emosbaugh: As i understand it, andrew is suggesting that you do not have to add [this line](https://github.com/replicatedhq/kURL/pull/4190/files#diff-f569000b511cf099e7e945359058ff85ba2054cc009017373d7e27218054caa3R123) to each version of ekco if you set the env var which is being used [here](https://github.com/replicatedhq/kURL/pull/4190/files#diff-f569000b511cf099e7e945359058ff85ba2054cc009017373d7e27218054caa3R124). i have suggested the change in my additional comments in this review. you will have to backport this to other versions of ekco.

scripts/common/rook.sh
```diff
@@ -278,6 +281,24 @@ function rook_operator_ready() {
...
+}
+
+# Check if the kurl-migration-from-rook exists then, if not creates it
+# To add DID_MIGRATE_ROOK_PVCS = "1" in order to track that the PVCs were migrated
+function add_rook_store_object_migration_status() {
+    if ! kubectl -n kurl get configmap kurl-migration-status; then
+       kubectl create configmap kurl-migration-from-rook -n kurl
+    fi
+    kubectl patch configmap kurl-migration-from-rook -n kurl --type merge -p '{"data":{"DID_MIGRATE_ROOK_OBJECT_STORE":"1"}}'
+}
```
emosbaugh, line 300:
```suggestion
    export DID_MIGRATE_ROOK_OBJECT_STORE=1
}
```

addons/ekco/template/base/install.sh
```diff
@@ -120,6 +120,7 @@ function ekco_already_applied() {
     local dst="$DIR/kustomize/ekco"
 
     # if rook-ceph has been removed, ekco should be redeployed to not attempt to manage it
+    DID_MIGRATE_ROOK_PVCS=$(kubectl -n kurl get configmap kurl-migration-from-rook -o jsonpath='{ .data.DID_MIGRATE_ROOK_PVCS }')
```
emosbaugh, line 123:
```suggestion
```

scripts/common/rook.sh
```diff
@@ -278,6 +281,24 @@ function rook_operator_ready() {
...
 }
 
+# Check if the kurl-migration-from-rook exists then, if not creates it
+# To add DID_MIGRATE_ROOK_PVCS = "1" in order to track that the PVCs were migrated
+function add_rook_pvc_migration_status() {
+    if ! kubectl -n kurl get configmap kurl-migration-status; then
+       kubectl create configmap kurl-migration-from-rook -n kurl
+    fi
+    kubectl patch configmap kurl-migration-from-rook -n kurl --type merge -p '{"data":{"DID_MIGRATE_ROOK_PVCS":"1"}}'
+}
```
emosbaugh, line 291:
```suggestion
    export DID_MIGRATE_ROOK_PVCS=1
}
```

scripts/common/rook.sh
camilamacedo86: I see what do you mean. Ok, I will split the change. 👍

PR #4196, opened 2023-03-13
create storage migration adr

laverya:

Fixes #

```release-note

```

### DIFF
--- a/docs/arch/adr-009-storage-migration.md
+++ b/docs/arch/adr-009-storage-migration.md
@@ -35,9 +35,12 @@ spec:
 When installed on a single node, two storageclasses will be available - "scaling" and "local". 
 The default storage class will be "scaling", and both storage classes will be backed by openebs localpv.
 
-When the third node is added, a ceph cluster will be created by the EKCO operator.
-When that ceph cluster becomes healthy (with at least 3 replicas), the "distributed" storageclass will be created using rook-ceph.
-pvmigrate will then be used to migrate all data from the "scaling" storageclass to the "distributed" storageclass, and the default storageclass will be changed to "distributed".
+When the third node is added, a Ceph cluster will be created by the EKCO operator.
+When that Ceph cluster becomes healthy (with at least 3 replicas), the "distributed" storageclass will be created using rook-ceph.
+If MinIO is present, Kotsadm will be scaled down, and the existing `sync-object-store` command will be used to migrate data from MinIO to Rook.
+Kotsadm, Registry, and Velero will then be updated to use the Rook object store, and Kotsadm scaled back up.
+After MinIO data is migrated and its consumers updated, the MinIO statefulset and namespace will be deleted.
+`pvmigrate` will then be used to migrate all data from the "scaling" storageclass to the "distributed" storageclass, and the default storageclass will be changed to "distributed".
 This process does involve stopping pods using "scaling" storage.
 
 In this way, applications can specifically request storage that will always be local to a node (with the "local" storageclass), or storage that will be distributed across the cluster (with the "distributed" storageclass).
@@ -48,10 +51,6 @@ YET TO BE DECIDED:
    1. Allow the user to trigger the migration with a 'tasks.sh migrate-storage' command.
    2. Provide a prompt in kotsadm that will trigger the migration.
    3. Trigger the migration automatically when the third node is added.
-2. What to do with object storage. (should data be migrated from MinIO to Rook, and MinIO deleted?)
-   1. Migrate data from MinIO to Rook, and delete MinIO. This requires some downtime to move between the object stores, and might interact poorly with any user applications that used the kurl object store.
-   2. Keep MinIO running, and use it for object storage. Disable (or otherwise do not use) Rook's object storage. This is easy to do, but rook will likely have better performance.
-   3. Forbid the use of MinIO in automatic migration configs. This would require migrating the registry from non-object-store (and thus non-ha) to object-store.
 
 ## Status
 
@@ -59,3 +58,10 @@ Proposed
 
 ## Consequences
 
+Vendors will be able to specify a single spec for single and multi-node installs, with a smooth migration path between the two.
+
+There will be short application downtime migrating between the two storage systems.
+This could be mitigated by changing pvmigrate to not stop all pods at once, but it is as yet unknown how this would be accomplished.
+
+Vendor applications will not be able to use the kurl-provided object store with this configuration, as this object store will change endpoints during the migration process.
+It is believed this will not impact any vendors.
--- a/docs/arch/adr-009-storage-migration.md
+++ b/docs/arch/adr-009-storage-migration.md
@@ -44,6 +44,7 @@ However, a migration will not begin until one of three things occurs:
 3. The user runs the `migrate-multinode-storage` command in `tasks.sh` from a primary node
 
 The migration process will NOT be triggered automatically.
+While the migration process is available to be triggered, a banner will be shown in Kotsadm indicating that the migration is available, and that the cluster is currently in an undesirable state.
 
 The migration process will be as follows:
 A trigger configmap will be created in the `kurl` namespace, to be observed by EKCO, prompting it to carry out the following steps.
@@ -51,7 +52,7 @@ If MinIO is present, Kotsadm will be scaled down, and the existing `sync-object-
 Kotsadm, Registry, and Velero will then be updated to use the Rook object store, and Kotsadm scaled back up.
 After MinIO data is migrated and its consumers updated, the MinIO statefulset and namespace will be deleted.
 `pvmigrate` will then be used to migrate all data from the "scaling" storageclass to the "distributed" storageclass, and the default storageclass will be changed to "distributed".
-This process does involve stopping pods using "scaling" storage.
+This process does involve downtime, caused by stopping pods using "scaling" storage, and is why we require the migration to be manually initiated.
 
 In this way, applications can specifically request storage that will always be local to a node (with the "local" storageclass), or storage that will be distributed across the cluster (with the "distributed" storageclass).
 Using the "scaling" storageclass directly (instead of merely using the default storageclass) would be an application linting error.
--- a/docs/arch/adr-009-storage-migration.md
+++ b/docs/arch/adr-009-storage-migration.md
@@ -39,17 +39,17 @@ When the third node is added, a Ceph cluster will be created by the EKCO operato
 When that Ceph cluster becomes healthy (with at least 3 replicas), the "distributed" storageclass will be created using rook-ceph.
 However, a migration will not begin until one of three things occurs:
 
-1. The user approves the migration in kotsadm
+1. The user approves the migration in KOTS
 2. The user runs `install.sh` on a primary node, and accepts a prompt to migrate storage
 3. The user runs the `migrate-multinode-storage` command in `tasks.sh` from a primary node
 
 The migration process will NOT be triggered automatically.
-While the migration process is available to be triggered, a banner will be shown in Kotsadm indicating that the migration is available, and that the cluster is currently in an undesirable state.
+While the migration process is available to be triggered, a banner will be shown in the KOTS browser UI indicating that the migration is available, and that the cluster is currently in an undesirable state.
 
 The migration process will be as follows:
 A trigger configmap will be created in the `kurl` namespace, to be observed by EKCO, prompting it to carry out the following steps.
-If MinIO is present, Kotsadm will be scaled down, and the existing `sync-object-store` command will be used to migrate data from MinIO to Rook.
-Kotsadm, Registry, and Velero will then be updated to use the Rook object store, and Kotsadm scaled back up.
+If MinIO is present, KOTS will be scaled down, and the existing `sync-object-store` Go command will be used to migrate data from MinIO to Rook.
+KOTS, Registry, and Velero will then be updated to use the Rook object store, and KOTS scaled back up.
 After MinIO data is migrated and its consumers updated, the MinIO statefulset and namespace will be deleted.
 `pvmigrate` will then be used to migrate all data from the "scaling" storageclass to the "distributed" storageclass, and the default storageclass will be changed to "distributed".
 This process does involve downtime, caused by stopping pods using "scaling" storage, and is why we require the migration to be manually initiated.
@@ -68,7 +68,7 @@ Vendors will be able to specify a single spec for single and multi-node installs
 There will be short application downtime migrating between the two storage systems.
 This could be mitigated by changing pvmigrate to not stop all pods at once, but it is as yet unknown how this would be accomplished.
 
-Vendor applications will not be able to use the kurl-provided object store with this configuration, as this object store will change endpoints during the migration process.
+Vendor applications will not be able to use the kurl-provided object store with this configuration, as this object store will change endpoints during the migration process and not preserve user credentials.
 It is believed this will not impact any vendors.
 
 Clusters may run with 3+ nodes and not use Rook for some time if the user does not approve the migration.
--- a/docs/arch/adr-009-storage-migration.md
+++ b/docs/arch/adr-009-storage-migration.md
@@ -1,0 +1,17 @@
+# ADR 9: OpenEBS to OpenEBS + Rook Migration
+
+## Context
+
+
+## Decision
+
+
+## Solution
+
+
+## Status
+
+Proposed
+
+## Consequences
+
--- a/docs/arch/adr-009-storage-migration.md
+++ b/docs/arch/adr-009-storage-migration.md
@@ -49,9 +49,9 @@ YET TO BE DECIDED:
    2. Provide a prompt in kotsadm that will trigger the migration.
    3. Trigger the migration automatically when the third node is added.
 2. What to do with object storage. (should data be migrated from MinIO to Rook, and MinIO deleted?)
-   1. Migrate data from MinIO to Rook, and delete MinIO.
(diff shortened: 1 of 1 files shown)


docs/arch/adr-009-storage-migration.md
```diff
@@ -0,0 +1,61 @@
...
+    localPVStorageClassName: "local"
+```
+
+When installed on a single node, two storageclasses will be available - "scaling" and "local". 
+The default storage class will be "scaling", and both storage classes will be backed by openebs localpv.
+
+When the third node is added, a ceph cluster will be created by the EKCO operator.
+When that ceph cluster becomes healthy (with at least 3 replicas), the "distributed" storageclass will be created using rook-ceph.
+pvmigrate will then be used to migrate all data from the "scaling" storageclass to the "distributed" storageclass, and the default storageclass will be changed to "distributed".
+This process does involve stopping pods using "scaling" storage.
```
emosbaugh, line 41:
```suggestion
This process does involve downtime, stopping pods using "scaling" storage.
```

docs/arch/adr-009-storage-migration.md
```diff
@@ -0,0 +1,61 @@
...
+
+In this way, applications can specifically request storage that will always be local to a node (with the "local" storageclass), or storage that will be distributed across the cluster (with the "distributed" storageclass).
+Using the "scaling" storageclass directly (instead of merely using the default storageclass) would be an application linting error.
+
+YET TO BE DECIDED:
+1. Should the migration be triggered immediately upon adding a third node, or should it be triggered by the user? It can involve application downtime!
+   1. Allow the user to trigger the migration with a 'tasks.sh migrate-storage' command.
+   2. Provide a prompt in kotsadm that will trigger the migration.
```
emosbaugh, line 49: should you mention something about warning the user with some sort of banner in the kots ui that the cluster is running in an unstable state prior to a storage migration?

docs/arch/adr-009-storage-migration.md
```diff
@@ -0,0 +1,61 @@
...
+   3. Trigger the migration automatically when the third node is added.
+2. What to do with object storage. (should data be migrated from MinIO to Rook, and MinIO deleted?)
```
emosbaugh, line 51: I believe the solution must also take into account vendors who do not require distributed storage such as swimlane. Can you go into a little more detail in the soution about how the cluster is configured in this use case? For example, is minio required to back the registry and velero?

docs/arch/adr-009-storage-migration.md
```diff
@@ -0,0 +1,61 @@
...
+  openebs:
+    version: "3.4.x"
+    isLocalPVEnabled: true
+    localPVStorageClassName: "local"
+```
+
+When installed on a single node, two storageclasses will be available - "scaling" and "local". 
+The default storage class will be "scaling", and both storage classes will be backed by openebs localpv.
+
+When the third node is added, a ceph cluster will be created by the EKCO operator.
```
ricardomaraschini, line 38: So Rook won't be deployed by the `installer` whenever `minimumNodeCount` is specified ?

ricardomaraschini: Note: installer still needs to load the images in all nodes.

docs/arch/adr-009-storage-migration.md
```diff
@@ -0,0 +1,61 @@
...
+    storageClassName: "distributed"
+    minimumNodeCount: 3
+  openebs:
+    version: "3.4.x"
+    isLocalPVEnabled: true
+    localPVStorageClassName: "local"
+```
+
+When installed on a single node, two storageclasses will be available - "scaling" and "local". 
+The default storage class will be "scaling", and both storage classes will be backed by openebs localpv.
```
ricardomaraschini, line 36: How is this going to work in currently installed cluster where the user has named OpenEBS storage class as `foo` and wants to have its data automatically migrated to `distributed` ?

docs/arch/adr-009-storage-migration.md
```diff
@@ -0,0 +1,61 @@
...
+When that ceph cluster becomes healthy (with at least 3 replicas), the "distributed" storageclass will be created using rook-ceph.
+pvmigrate will then be used to migrate all data from the "scaling" storageclass to the "distributed" storageclass, and the default storageclass will be changed to "distributed".
+This process does involve stopping pods using "scaling" storage.
+
+In this way, applications can specifically request storage that will always be local to a node (with the "local" storageclass), or storage that will be distributed across the cluster (with the "distributed" storageclass).
+Using the "scaling" storageclass directly (instead of merely using the default storageclass) would be an application linting error.
+
+YET TO BE DECIDED:
+1. Should the migration be triggered immediately upon adding a third node, or should it be triggered by the user? It can involve application downtime!
+   1. Allow the user to trigger the migration with a 'tasks.sh migrate-storage' command.
```
ricardomaraschini, line 48: I prefer this approach.

docs/arch/adr-009-storage-migration.md
laverya: it'll be based on whatever openebs storageclass name you have specified. The only one that isn't configurable is `scaling`, and that will be created upon updating to this kurl version with the appropriate config.

If someone for some reason has an existing storageclass named `scaling`, I'm happy to say I'll burn that bridge when I get to it.

docs/arch/adr-009-storage-migration.md
laverya: Rook will be deployed, but just the operator - there won't be a cephcluster or storageclass to begin with (and thus no OSD pods etc)

docs/arch/adr-009-storage-migration.md
laverya: yeah, will do

docs/arch/adr-009-storage-migration.md
laverya: I would say that this proposal is almost completely irrelevant to vendors that don't require distributed storage!

That being said, I've added automatic minio migration to this doc.

docs/arch/adr-009-storage-migration.md
laverya: that's what I'm going with (along with kotsadm trigger)

docs/arch/adr-009-storage-migration.md
```diff
@@ -0,0 +1,74 @@
...
+When installed on a single node, two storageclasses will be available - "scaling" and "local". 
+The default storage class will be "scaling", and both storage classes will be backed by openebs localpv.
+
+When the third node is added, a Ceph cluster will be created by the EKCO operator.
+When that Ceph cluster becomes healthy (with at least 3 replicas), the "distributed" storageclass will be created using rook-ceph.
+However, a migration will not begin until one of three things occurs:
+
+1. The user approves the migration in kotsadm
+2. The user runs `install.sh` on a primary node, and accepts a prompt to migrate storage
+3. The user runs the `migrate-multinode-storage` command in `tasks.sh` from a primary node
```
emosbaugh, line 44: I'd prefer that this be a go command directly so that it is reusable. If the only interface we have for the end user is tasks.sh then the shell should only invoke a go command and nothing else.

docs/arch/adr-009-storage-migration.md
```diff
@@ -0,0 +1,74 @@
...
+
+The migration process will NOT be triggered automatically.
+While the migration process is available to be triggered, a banner will be shown in Kotsadm indicating that the migration is available, and that the cluster is currently in an undesirable state.
```
emosbaugh, line 47:
```suggestion
While the migration process is available to be triggered, a banner will be shown in the KOTS browser UI indicating that the migration is available, and that the cluster is currently in an undesirable state.
```

docs/arch/adr-009-storage-migration.md
```diff
@@ -0,0 +1,74 @@
...
+
+The migration process will be as follows:
+A trigger configmap will be created in the `kurl` namespace, to be observed by EKCO, prompting it to carry out the following steps.
```
emosbaugh, line 50: Can we make use of a CRD? This may be new to the project but it is about time.

docs/arch/adr-009-storage-migration.md
```diff
@@ -0,0 +1,74 @@
...
+If MinIO is present, Kotsadm will be scaled down, and the existing `sync-object-store` command will be used to migrate data from MinIO to Rook.
```
emosbaugh, line 51:
```suggestion
If MinIO is present, KOTS will be scaled down, and the existing `sync-object-store` Go command will be used to migrate data from MinIO to Rook.
```

docs/arch/adr-009-storage-migration.md
```diff
@@ -0,0 +1,74 @@
...
+Kotsadm, Registry, and Velero will then be updated to use the Rook object store, and Kotsadm scaled back up.
```
emosbaugh, line 52:
```suggestion
KOTS, Registry, and Velero will then be updated to use the Rook object store, and KOTS scaled back up.
```

docs/arch/adr-009-storage-migration.md
```diff
@@ -0,0 +1,74 @@
...
+```
+
+When installed on a single node, two storageclasses will be available - "scaling" and "local". 
+The default storage class will be "scaling", and both storage classes will be backed by openebs localpv.
+
+When the third node is added, a Ceph cluster will be created by the EKCO operator.
+When that Ceph cluster becomes healthy (with at least 3 replicas), the "distributed" storageclass will be created using rook-ceph.
+However, a migration will not begin until one of three things occurs:
+
+1. The user approves the migration in kotsadm
```
emosbaugh, line 42:
```suggestion
1. The user approves the migration in KOTS
```

docs/arch/adr-009-storage-migration.md
```diff
@@ -0,0 +1,74 @@
...
+Proposed
+
+## Consequences
+
+Vendors will be able to specify a single spec for single and multi-node installs, with a smooth migration path between the two.
+
+There will be short application downtime migrating between the two storage systems.
+This could be mitigated by changing pvmigrate to not stop all pods at once, but it is as yet unknown how this would be accomplished.
+
+Vendor applications will not be able to use the kurl-provided object store with this configuration, as this object store will change endpoints during the migration process.
```
emosbaugh, line 71:
```suggestion
Vendor applications will not be able to use the kurl-provided object store with this configuration, as this object store will change endpoints during the migration process and not preserve user credentials.
```

docs/arch/adr-009-storage-migration.md
```diff
@@ -0,0 +1,74 @@
...
+
+When installed on a single node, two storageclasses will be available - "scaling" and "local". 
+The default storage class will be "scaling", and both storage classes will be backed by openebs localpv.
+
+When the third node is added, a Ceph cluster will be created by the EKCO operator.
+When that Ceph cluster becomes healthy (with at least 3 replicas), the "distributed" storageclass will be created using rook-ceph.
+However, a migration will not begin until one of three things occurs:
+
+1. The user approves the migration in kotsadm
+2. The user runs `install.sh` on a primary node, and accepts a prompt to migrate storage
```
emosbaugh, line 43: Can we do any trickery to prompt and migrate the storage on a secondary node? This seems particularly useful for the use case where the user inits a cluster and immediately joins two additional worker nodes before installing a KOTS app.

emosbaugh (changes requested):

docs/arch/adr-009-storage-migration.md
```diff
@@ -0,0 +1,74 @@
...
+
+The migration process will be as follows:
+A trigger configmap will be created in the `kurl` namespace, to be observed by EKCO, prompting it to carry out the following steps.
+If MinIO is present, Kotsadm will be scaled down, and the existing `sync-object-store` command will be used to migrate data from MinIO to Rook.
+Kotsadm, Registry, and Velero will then be updated to use the Rook object store, and Kotsadm scaled back up.
+After MinIO data is migrated and its consumers updated, the MinIO statefulset and namespace will be deleted.
+`pvmigrate` will then be used to migrate all data from the "scaling" storageclass to the "distributed" storageclass, and the default storageclass will be changed to "distributed".
+This process does involve downtime, caused by stopping pods using "scaling" storage, and is why we require the migration to be manually initiated.
+
+In this way, applications can specifically request storage that will always be local to a node (with the "local" storageclass), or storage that will be distributed across the cluster (with the "distributed" storageclass).
```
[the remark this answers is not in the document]
laverya: The idea is that 'distributed' will always be on Rook, and 'scaling' will always be on openebs-to-be-moved-to-rook. 

If we rename back to 'scaling', there's no way for a vendor to say "I don't want this to run until it's distributed storage".

docs/arch/adr-009-storage-migration.md
laverya: Agreed that the logic in tasks.sh should be minimal-to-nonexistent

docs/arch/adr-009-storage-migration.md
```diff
@@ -0,0 +1,74 @@
...
+
+## Consequences
+
+Vendors will be able to specify a single spec for single and multi-node installs, with a smooth migration path between the two.
+
+There will be short application downtime migrating between the two storage systems.
+This could be mitigated by changing pvmigrate to not stop all pods at once, but it is as yet unknown how this would be accomplished.
+
+Vendor applications will not be able to use the kurl-provided object store with this configuration, as this object store will change endpoints during the migration process.
+It is believed this will not impact any vendors.
```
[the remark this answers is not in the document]
laverya: Theoretically they could use rook-ceph (when it becomes available)

But if they depend on it, then why go through the trouble of migrating from openebs when they could just require Rook from the beginning?

docs/arch/adr-009-storage-migration.md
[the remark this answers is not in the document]
laverya: For both Ethan and Kyle - I would love a way to trigger this upon adding the third node, even if it's not a primary node.

I just don't see a way to do this besides either:
1. adding a http endpoint, and having the script curl that after the node has been joined
2. having a daemonset watching files in a directory on the host

and both of those seem a little irritating, but in retrospect the only real irritation from the http endpoint would be authentication and choosing an appropriate port. (And we could include a token in the join script + use a dynamic port, so that's doable)

docs/arch/adr-009-storage-migration.md
laverya: it might actually be better to make it a http endpoint, so that join.sh could hit it too

docs/arch/adr-009-storage-migration.md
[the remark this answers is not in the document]
emosbaugh: i think we need to answer this question. i cannot get onboard with this proposal without some way to bootstrap a 3 node cluster and not have to go through a storage migration.

docs/arch/adr-009-storage-migration.md
emosbaugh: when you come to a decision will you change this?

docs/arch/adr-009-storage-migration.md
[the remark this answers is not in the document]
laverya: I think an easy solution would be to do an automatic migration if no app has been installed yet, or if the app doesn't use any PVCs in the scaling storageclass - that way we only require customer approval for things that will take the app down. (things that will take kotsadm down are different, IMO)

docs/arch/adr-009-storage-migration.md
laverya: Yeah, I'll change this to the HTTP endpoint option

emosbaugh (approved):

docs/arch/adr-009-storage-migration.md
```diff
@@ -0,0 +1,94 @@
...
+    version: "1.10.x"
+    storageClassName: "distributed"
+    minimumNodeCount: 3
+  openebs:
+    version: "3.4.x"
+    isLocalPVEnabled: true
+    localPVStorageClassName: "local"
+```
+
+When installed on a single node, two storageclasses will be available - "scaling" and "local".
```
rrpolanco, line 35: Calling it _scaling_ doesn't seem to convey it's purpose clear enough for me. Perhaps we can call it _centralized_ which is the opposite of _distributed_? _Centralized_ makes sense to me as it convey cluster storage that is currently designated for this node aside from calling it something like 'local 2`/ 'second local`.

emosbaugh: how about "dynamic"?

emosbaugh: "elastic"?

laverya: 'dynamic' would be fine with me

laverya: I'm going to merge this now though, and we can make changes to it afterwards

rrpolanco: `dynamic` is better but I still think looking at the SC list, after a third node is added, that it better conveys the state of the cluster (i.e. that we went from non-distributed to distributed storage).
```
kubectl get sc
NAME                        PROVISIONER                                                RECLAIMPOLICY   VOLUMEBINDINGMODE      ALLOWVOLUMEEXPANSION   AGE
local                       openebs.io/local                                           Delete          WaitForFirstConsumer   false                  16d
centralized                 openebs.io/local                                           Delete          WaitForFirstConsumer   false                  16d
distributed (default)       rook-ceph.rbd.csi.ceph.com                                          
```

rrpolanco:
@laverya You're also going to have to update the status to _Accepted_

PR #4195, opened 2023-03-13
fix: Adds check and info to allow users be aware that Rook can not be uninstalled programmatically when the migration from Rook to OpenEBS is done without Minio be installed or when was not possible successfully migrate the Object Store or PVCs

camilamacedo86:
#### What this PR does / why we need it:

Currently when the condition is not satisfied we do not log that we were checking if Rook could or not be removed and that was not done or way. Also, we do not provide enough info to let the users know how then can manually removed when it is required. 

Therefore this PR:
- If rook is not selected in the spec but is in the cluster we add the log "Removing rook" so that we know that we will try to do it 
- If MINIO is not selected and because of it the Object Store was not migrated then, we let the users be aware of why Rook is not delete and we provide the manual steps to achieve the goal if they wish
- If from a migration from Rook to OpenEBS + Minio Rook was not deleted then, we log with warning if PVC or Object store was not migrated with success.

#### Which issue(s) this PR fixes:

Partial (it does not close the short cut) # [sc-70239]

#### Special notes for your reviewer:

Screenshot 2023-03-16 at 08 22 29

#### Does this PR introduce a user-facing change?

```release-note
Fixes issue when migrate from Rook and it is not uninstalled by adding a check and better log information to allow users be aware of and the reasons for Rook not be able to be uninstalled programmatically.
```

PR #4199, opened 2023-03-14
chore: creating job to test upgrades and migration paths.

ricardomaraschini:

This new job is composed by some tests copied from the `full.yaml` file (all tests involving cluster upgrades) plus 14 new storage provisioner migration tests. These new tests are going to run daily but can also be dispatched manually.


emosbaugh (commented):
without looking too deeply, why did you choose to move all upgrades to the new daily? my understanding is that is was for migrations only. i will look continue to review

ricardomaraschini:
@emosbaugh When you say migrations do you mean "storage provisioner migrations" or any kind of migration ? I certainly misunderstood this part.

emosbaugh (commented):
do we need to validate that the namespace was successfully removed in cases where rook and longhorn are removed?

ricardomaraschini:
> do we need to validate that the namespace was successfully removed in cases where rook and longhorn are removed?

Yes, that is an excellent idea. Done.

ricardomaraschini:
> > @emosbaugh When you say migrations do you mean "storage provisioner migrations" or any kind of migration ? I certainly misunderstood this part.
> 
> the storage provisioner migrations defined in the spreadsheet. you did a good job adding them all, i double checked. i was only asking that you remove duplicated tests from the current daily if there are any as part of the task.

It looks like everything is in order. I have returned the original `full.yaml` and transferred only one test to the new `storage-migration.yaml` file. I searched for duplicate tests but couldn't find any others. I have also removed a test that validated the migration from Rook to Longhorn. Please let me know if you believe that test should be kept.

Regarding the duplicate tests in Rook and OpenEBS add-ons, I think it would be beneficial to keep them there to aid in identifying issues before merging new add-on versions. What are your thoughts on this?

emosbaugh (approved):
this looks good. i think it makes sense to leave the duplicate tests in the add-ons

testgrid/specs/storage-migration.yaml
```diff
@@ -0,0 +1,767 @@
+- name: Migrate from Rook 1.0.4 to OpenEBS + Minio
```
camilamacedo86, line 1: We have none tests with docker and unfortunately it still used.

Can we add a test with the migration that we have been facing issues
(Migration from Rook with Docker config and k8s 1.19 to 1.21)

install:
```
spec:
  kubernetes:
    version: 1.19.3
  docker:
    daemonConfig: |
      {
        \"exec-opts\": [\"native.cgroupdriver=systemd\"],
        \"log-opts\": {
            \"max-size\": \"100m\",
            \"max-file\": \"3\"
            }
      }
    version: 19.03.10
  weave:
    version: 2.6.5
  rook:
    version: 1.0.4
  openebs:
    isCstorEnabled: false
    isLocalPVEnabled: true
    localPVStorageClassName: openebs
    namespace: openebs
    version: 1.6.0
  contour:
    tlsMinimumProtocolVersion: '1.2'
    version: 1.22.0
  registry:
    version: 2.7.1
  prometheus:
    version: 0.58.0-39.4.0
  kotsadm:
    version: 1.57.0
  velero:
    version: 1.9.0
  ekco:
    minReadyMasterNodeCount: 2
    minReadyWorkerNodeCount: 0
    nodeUnreachableToleration: 5m
    rookShouldUseAllNodes: false
    shouldDisableClearNodes: false
    shouldDisableRebootServices: false
    shouldEnablePurgeNodes: false
    version: 0.16.0
  goldpinger:
    version: 3.2.0-4.1.1

```

upgrade 


```
spec:
  kubernetes:
    version: 1.21.14
  docker:
    daemonConfig: |
      {
        \"exec-opts\": [\"native.cgroupdriver=systemd\"],
        \"log-opts\": {
          \"max-size\": \"100m\",
          \"max-file\": \"3\"
        }
      }
    version: 19.03.10
  weave:
    version: 2.6.5-20221025
  openebs:
    isCstorEnabled: false
    isLocalPVEnabled: true
    localPVStorageClassName: openebs
    namespace: openebs
    version: 3.3.0
  minio:
    version: 2022-10-20T00-55-09Z
  contour:
    tlsMinimumProtocolVersion: '1.2'
    version: 1.23.0
  registry:
    version: 2.8.1
  kotsadm:
    version: 1.90.0
    s3Override: https://s3.kurl.sh/external/kotsadm-1.90.0.tar.gz
  velero:
    resticTimeout: 24h0m0s
    version: 1.9.2
  ekco:
    minReadyMasterNodeCount: 2
    minReadyWorkerNodeCount: 0
    nodeUnreachableToleration: 5m
    shouldDisableClearNodes: false
    shouldDisableRebootServices: false
    shouldEnablePurgeNodes: false
    version: latest
  goldpinger:
    version: 3.7.0-5.5.0
```

testgrid/specs/storage-migration.yaml
```diff
@@ -0,0 +1,767 @@
...
+    create_random_file_and_upload_to_deployment "migration-test" "default" "./test.data" "/data/test.data"
+    test_push_image_to_registry
+  postUpgradeScript: |
+    source /opt/kurl-testgrid/testhelpers.sh
+    sleep 120
+    download_file_from_deployment_and_compare "migration-test" "default" "./test.data" "/data/test.data"
+    pvc_uses_provisioner "migration-test" "default" "rook"
+    test_pull_image_from_registry
+    if kubectl get namespace/rook-ceph ; then
+      echo Rook namespace found after the upgrade.
```
camilamacedo86, line 255: This one should not check  by kubectl get namespace/rook-ceph.
The Minio is not selected so we cannot migrate the object store which means that we cannot delete Rook.

testgrid/specs/storage-migration.yaml
```diff
@@ -0,0 +1,767 @@
...
+    create_deployment_with_mounted_volume "migration-test" "default" "/data"
+    create_random_file_and_upload_to_deployment "migration-test" "default" "./test.data" "/data/test.data"
+    test_push_image_to_registry
+  postUpgradeScript: |
+    source /opt/kurl-testgrid/testhelpers.sh
+    sleep 120
+    download_file_from_deployment_and_compare "migration-test" "default" "./test.data" "/data/test.data"
+    pvc_uses_provisioner "migration-test" "default" "rook"
+    test_pull_image_from_registry
+    if kubectl get namespace/rook-ceph ; then
```
camilamacedo86, line 304: same here.

testgrid/specs/storage-migration.yaml
```diff
@@ -0,0 +1,767 @@
...
+      echo Rook namespace found after the upgrade.
+      exit 1
+    fi
+- name: Migrate from Rook 1.4.x to OpenEBS (S3 disabled)
```
camilamacedo86, line 258: This tile is == the other one. Could we have a name that let us know the diff?

testgrid/specs/storage-migration.yaml
```diff
@@ -0,0 +1,767 @@
...
+    source /opt/kurl-testgrid/testhelpers.sh
+    sleep 120
+    download_file_from_deployment_and_compare "migration-test" "default" "./test.data" "/data/test.data"
+    pvc_uses_provisioner "migration-test" "default" "rook"
+    test_pull_image_from_registry
+    if kubectl get namespace/longhorn-system ; then
+      echo Longhorn namespace found after the upgrade.
+      exit 1
+    fi
+- name: Migrate from Rook 1.0.4 to OpenEBS + Minio
```
camilamacedo86, line 559: Is not this one duplicated? 
I found twice `Migrate from Rook 1.0.4 to OpenEBS + Minio`

testgrid/specs/storage-migration.yaml
```diff
@@ -0,0 +1,767 @@
...
+    source /opt/kurl-testgrid/testhelpers.sh
+    sleep 120
+    download_file_from_deployment_and_compare "migration-test" "default" "./test.data" "/data/test.data"
+    pvc_uses_provisioner "migration-test" "default" "rook"
+    test_pull_image_from_registry
+    if kubectl get namespace/rook-ceph ; then
+      echo Rook namespace found after the upgrade.
+      exit 1
+    fi
+- name: Migrate from Rook 1.4.x to OpenEBS + Minio
```
camilamacedo86, line 613: Should we not have migrate from OpenEBS and Minio to Rook

camilamacedo86 (changes requested):
I added some comments.
Did you check it by pushing manually a testgrid with those tests?
Can you add the link for the test made so that we can verify if has no syntax errors and etc and the tests will be executed?

testgrid/specs/storage-migration.yaml
ricardomaraschini: Great idea. I have added a test case using Docker instead of containerd.

testgrid/specs/storage-migration.yaml
ricardomaraschini: What happens on this scenario ? Does the install script starts and then crashes or does it bails out before starting to upgrade the cluster ?

testgrid/specs/storage-migration.yaml
ricardomaraschini: Kinda confusing but this one has `S3 disabled` while the other one does not.

testgrid/specs/storage-migration.yaml
ricardomaraschini: Yeah, copy and pasta mistake. Good catch. Removed.

ricardomaraschini:
> I added some comments. Did you check it by pushing manually a testgrid with those tests? Can you add the link for the test made so that we can verify if has no syntax errors and etc and the tests will be executed?

https://testgrid.kurl.sh/run/STAGING-daily-storage-migration-5496608-2023-03-15T14:04:59Z

testgrid/specs/storage-migration.yaml
camilamacedo86: Currently nothing and no logs at all. We indeed does not know that we checked if we could or not remove rook
See: https://github.com/replicatedhq/kURL/pull/4195

testgrid/specs/storage-migration.yaml
emosbaugh: I do not think this is supported. We do not support rook on single node in the future. On multi node I dont know of any vendors that are going from openebs to rook.

camilamacedo86 (approved):
I think has something wrong : https://github.com/replicatedhq/kURL/pull/4199#issuecomment-1470074290

However, I am ok with:
- we get it merged 
- then we checked it in testing (testgrid) 
- and if we need we do a follow up.

testgrid/specs/storage-migration.yaml
```diff
@@ -831,3 +825,53 @@
...
+    source /opt/kurl-testgrid/testhelpers.sh
+    create_deployment_with_mounted_volume "migration-test" "default" "/data" "registry:2.8.1"
+    create_random_file_and_upload_to_deployment "migration-test" "default" "./test.data" "/data/test.data"
+    test_push_image_to_registry
+  postUpgradeScript: |
+    source /opt/kurl-testgrid/testhelpers.sh
+    sleep 120
+    download_file_from_deployment_and_compare "migration-test" "default" "./test.data" "/data/test.data"
+    pvc_uses_provisioner "migration-test" "default" "openebs"
+    test_pull_image_from_registry
```
camilamacedo86, line 877: Can we add here the check to ensure that the rook-ceph ns was removed?
Only that then it shows fine.

camilamacedo86 (approved):

testgrid/specs/storage-migration.yaml
ricardomaraschini: Rook is not installed on this test so I don't think we actually need that.

ricardomaraschini:
This PR depends on https://github.com/replicatedhq/kURL/pull/4199 to be merged and deployed first.

camilamacedo86 (changes requested):
We should also verify if we cleanup the Rook data:

```sh
    echo "Verify if Rook data was removed"
    if [ -d "/var/lib/rook" ] || [ -d "/opt/replicated/rook" ]; then
            echo  "Unable to clean the Rook Data"
            exit 1
    fi
```

ricardomaraschini:
> We should also verify if we cleanup the Rook data:
> 
> ```shell
>     echo "Verify if Rook data was removed"
>     if [ -d "/var/lib/rook" ] || [ -d "/opt/replicated/rook" ]; then
>             echo  "Unable to clean the Rook Data"
>             exit 1
>     fi
> ```

Added.

emosbaugh (approved):

testgrid/specs/storage-migration.yaml
```diff
@@ -51,6 +51,10 @@
       echo Rook namespace found after the upgrade.
       exit 1
     fi
+    if [ -d "/var/lib/rook" ] || [ -d "/opt/replicated/rook" ]; then
```
emosbaugh, line 54: I assume we expect this to fail at this point @ricardomaraschini  @camilamacedo86 ?

https://github.com/replicatedhq/kURL/blob/61580229c2758349b8ac3efb39e22bdff19298dc/scripts/common/rook.sh#L121

rrpolanco (dismissed):

testgrid/specs/storage-migration.yaml
ricardomaraschini: Ouch, thanks for the heads-up @emosbaugh . Seems like we never delete those directories so all tests moving the cluster from Rook to OpenEBS will fail. @camilamacedo86 when you proposed this change what did you expect ? The directories to be removed or that we will eventually patch the migration code to remove them ?

ricardomaraschini:
@emosbaugh @rrpolanco I had to rebase once more this PR, can I have your review again ?

testgrid/specs/storage-migration.yaml
ricardomaraschini: Oh, nice. Thanks.

camilamacedo86 (approved):

PR #4204, opened 2023-03-16
Fixes issue when migrate from Longhorn and it is not uninstalled by adding a check and better log information to allow users be aware of and the reasons for Longhorn not be able to be uninstalled programmatically.

camilamacedo86:

Currently, by installing a new install where we are migrating from Longhorn users are unable to check that Longhorn was or not delete. Also, no information about why it was not possible to delete it is provided to them when failures occurs. 

Fixes # [sc-71884]

Just adding the namespace to check out the error;

Screenshot 2023-03-17 at 13 05 45

Just forcing DID_MIGRATE_LONGHORN_PVCS=1 to check out its results:

Screenshot 2023-03-17 at 13 08 54

(script fails because it is a mock longhorn does not exist)

PS: Tested without the namespace to check that would not print the step

```release-note
Fixes issue when migrate from Longhorn and it is not uninstalled by adding a check and better log information to allow users be aware of and the reasons for Longhorn not be able to be uninstalled programmatically.
```

### DIFF
--- a/scripts/common/longhorn.sh
+++ b/scripts/common/longhorn.sh
@@ -155,7 +155,7 @@ function maybe_cleanup_longhorn() {
 
         logWarn ""
         logWarn "ATTENTION: Unable to remove Longhorn."
-        if [ "$DID_MIGRATE_LONGHORN_PVCS" == "1" ]; then
+        if [ "$DID_MIGRATE_LONGHORN_PVCS" != "1" ]; then
             logFail "Storage class migration did not succeed"
         fi
     fi
--- a/scripts/common/longhorn.sh
+++ b/scripts/common/longhorn.sh
@@ -154,7 +154,7 @@ function maybe_cleanup_longhorn() {
         fi
 
         logWarn ""
-        logWarn "ATTENTION: Unable to remove Longhorn."
+        logFail "Unable to remove Longhorn."
         if [ "$DID_MIGRATE_LONGHORN_PVCS" != "1" ]; then
             logFail "Storage class migration did not succeed"
         fi
--- a/scripts/common/longhorn.sh
+++ b/scripts/common/longhorn.sh
@@ -145,7 +145,7 @@ function maybe_cleanup_longhorn() {
         if ! kubectl get ns | grep -q longhorn-system; then
             return
         fi 
-            logStep "Removing Longhorn"
+        logStep "Removing Longhorn"
         if [ "$DID_MIGRATE_LONGHORN_PVCS" == "1" ]; then
             report_addon_start "longhorn-removal" "v1"
             remove_longhorn
--- a/scripts/common/longhorn.sh
+++ b/scripts/common/longhorn.sh
@@ -141,9 +141,10 @@ function longhorn_to_sc_migration() {
 # if PVCs and object store data have both been migrated from longhorn and longhorn is no longer specified in the kURL spec, remove longhorn
 function maybe_cleanup_longhorn() {
     if [ -z "$LONGHORN_VERSION" ]; then
-        if kubectl get ns | grep -q longhorn-system; then
-            logStep "Removing Longhorn ..."
-        fi
+        # Just continue if longhorn is installed. 
+        if ! kubectl get ns | grep -q longhorn-system; then
+            return
+        fi 
 
         if [ "$DID_MIGRATE_LONGHORN_PVCS" == "1" ]; then
             report_addon_start "longhorn-removal" "v1"
--- a/scripts/common/longhorn.sh
+++ b/scripts/common/longhorn.sh
@@ -141,10 +141,21 @@ function longhorn_to_sc_migration() {
 # if PVCs and object store data have both been migrated from longhorn and longhorn is no longer specified in the kURL spec, remove longhorn
 function maybe_cleanup_longhorn() {
     if [ -z "$LONGHORN_VERSION" ]; then
+        if kubectl get ns | grep -q longhorn-system; then
+            logStep "Removing Longhorn ..."
+        fi
+
         if [ "$DID_MIGRATE_LONGHORN_PVCS" == "1" ]; then
             report_addon_start "longhorn-removal" "v1"
             remove_longhorn
             report_addon_success "longhorn-removal" "v1"
+            return
+        fi
+
+        logWarn ""
+        logWarn "ATTENTION: Unable to remove Rook."
+        if [ "$DID_MIGRATE_LONGHORN_PVCS" == "1" ]; then
+            logFail "Storage class migration did not succeed"
         fi
     fi
 }
--- a/scripts/common/longhorn.sh
+++ b/scripts/common/longhorn.sh
@@ -145,7 +145,7 @@ function maybe_cleanup_longhorn() {
         if ! kubectl get ns | grep -q longhorn-system; then
             return
         fi 
-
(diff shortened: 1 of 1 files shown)


scripts/common/longhorn.sh
```diff
@@ -141,10 +141,21 @@ function longhorn_to_sc_migration() {
...
+
         if [ "$DID_MIGRATE_LONGHORN_PVCS" == "1" ]; then
             report_addon_start "longhorn-removal" "v1"
             remove_longhorn
             report_addon_success "longhorn-removal" "v1"
+            return
+        fi
+
+        logWarn ""
+        logWarn "ATTENTION: Unable to remove Rook."
```
ricardomaraschini, line 156: Wasn't this supposed to read "Unable to remove longhorn" ?

scripts/common/longhorn.sh
```diff
@@ -141,10 +141,21 @@ function longhorn_to_sc_migration() {
...
+        if [ "$DID_MIGRATE_LONGHORN_PVCS" == "1" ]; then
+            logFail "Storage class migration did not succeed"
         fi
     fi
```
laverya, line 160: This logic... Has issues.

For one, we shouldn't log that we're removing longhorn unless `DID_MIGRATE_LONGHORN_PVCS` is `1`. Not even if the namespace is present.

For another, the 
```
logWarn ""
logWarn "ATTENTION: Unable to remove Rook."
```
lines WILL ALWAYS RUN if `LONGHORN_VERSION` isn't set! And it mentions Rook, not Longhorn!

Finally, 
```
        if [ "$DID_MIGRATE_LONGHORN_PVCS" == "1" ]; then
            logFail "Storage class migration did not succeed"
        fi
```
will NEVER run, as it has the same condition as the `remove_longhorn` call above!

laverya:
```suggestion
    if [ -z "$LONGHORN_VERSION" ]; then
        if [ "$DID_MIGRATE_LONGHORN_PVCS" == "1" ]; then
            logStep "Removing Longhorn"
            report_addon_start "longhorn-removal" "v1"
            remove_longhorn
            report_addon_success "longhorn-removal" "v1"
            return
        fi
    fi
```
At most, you can do this - but I would recommend making sure that `remove_longhorn` does not itself have a logstep.

scripts/common/longhorn.sh
```diff
@@ -141,10 +141,21 @@ function longhorn_to_sc_migration() {
 # if PVCs and object store data have both been migrated from longhorn and longhorn is no longer specified in the kURL spec, remove longhorn
 function maybe_cleanup_longhorn() {
     if [ -z "$LONGHORN_VERSION" ]; then
+        if kubectl get ns | grep -q longhorn-system; then
+            logStep "Removing Longhorn ..."
+        fi
```
camilamacedo86, line 146:
```suggestion
        # Just continue if longhorn is installed. 
        if ! kubectl get ns | grep -q longhorn-system; then
            return
        fi 
```

scripts/common/longhorn.sh
```diff
@@ -141,10 +141,22 @@ function longhorn_to_sc_migration() {
...
+
         if [ "$DID_MIGRATE_LONGHORN_PVCS" == "1" ]; then
             report_addon_start "longhorn-removal" "v1"
             remove_longhorn
             report_addon_success "longhorn-removal" "v1"
+            return
+        fi
+
+        logWarn ""
+        logWarn "ATTENTION: Unable to remove Rook."
```
camilamacedo86, line 157:
```suggestion
        logWarn "ATTENTION: Unable to remove Longhorn."
```

scripts/common/longhorn.sh
```diff
@@ -141,10 +141,22 @@ function longhorn_to_sc_migration() {
...
         if [ "$DID_MIGRATE_LONGHORN_PVCS" == "1" ]; then
             report_addon_start "longhorn-removal" "v1"
             remove_longhorn
             report_addon_success "longhorn-removal" "v1"
+            return
+        fi
+
+        logWarn ""
+        logWarn "ATTENTION: Unable to remove Longhorn."
+        if [ "$DID_MIGRATE_LONGHORN_PVCS" == "1" ]; then
```
camilamacedo86, line 158:
```suggestion
        if [ "$DID_MIGRATE_LONGHORN_PVCS" != "1" ]; then
```

scripts/common/longhorn.sh
camilamacedo86: No that does not solve the problem.
We must have the logs and know that;
1- We are trying to remove longhorn
2- We did not remove because did not pass and etc. 


See the code now.

scripts/common/longhorn.sh
```diff
@@ -141,10 +141,22 @@ function longhorn_to_sc_migration() {
 # if PVCs and object store data have both been migrated from longhorn and longhorn is no longer specified in the kURL spec, remove longhorn
 function maybe_cleanup_longhorn() {
     if [ -z "$LONGHORN_VERSION" ]; then
+        # Just continue if longhorn is installed. 
+        if ! kubectl get ns | grep -q longhorn-system; then
+            return
+        fi 
+
```
camilamacedo86, line 148:
```suggestion
            logStep "Removing Longhorn"
```

scripts/common/longhorn.sh
```diff
@@ -141,10 +141,22 @@ function longhorn_to_sc_migration() {
 # if PVCs and object store data have both been migrated from longhorn and longhorn is no longer specified in the kURL spec, remove longhorn
 function maybe_cleanup_longhorn() {
     if [ -z "$LONGHORN_VERSION" ]; then
+        # Just continue if longhorn is installed. 
+        if ! kubectl get ns | grep -q longhorn-system; then
+            return
+        fi 
+            logStep "Removing Longhorn"
```
camilamacedo86, line 148:
```suggestion
        logStep "Removing Longhorn"
```

scripts/common/longhorn.sh
```diff
@@ -141,10 +141,22 @@ function longhorn_to_sc_migration() {
...
+        logStep "Removing Longhorn"
         if [ "$DID_MIGRATE_LONGHORN_PVCS" == "1" ]; then
             report_addon_start "longhorn-removal" "v1"
             remove_longhorn
             report_addon_success "longhorn-removal" "v1"
+            return
+        fi
+
+        logWarn ""
+        logWarn "ATTENTION: Unable to remove Longhorn."
```
camilamacedo86, line 157:
```suggestion
        logFail "Unable to remove Longhorn."
```

scripts/common/longhorn.sh
```diff
@@ -141,10 +141,22 @@ function longhorn_to_sc_migration() {
...
+        fi 
+        logStep "Removing Longhorn"
         if [ "$DID_MIGRATE_LONGHORN_PVCS" == "1" ]; then
             report_addon_start "longhorn-removal" "v1"
             remove_longhorn
             report_addon_success "longhorn-removal" "v1"
+            return
+        fi
+
+        logWarn ""
```
camilamacedo86, line 156:
```suggestion
```

camilamacedo86 (commented):
@laverya all changes are addressed now. It seems good to fly.
WDYT?

laverya (approved):

PR #4205, opened 2023-03-16
fix: just ensure that we have the step so we can have clear logs

camilamacedo86:

By looking the logs we will see something like:

Screenshot 2023-03-16 at 08 57 14

See that is not clear from where the config pacthes came from or in what step it is made.
That is confusing (mainly to troubleshot) since we use the stepLog to make clear about each part of the workflows.

So, this PR fix that by adding a logStep prior config kurl. 

Fixes #

```release-note
Fixes missguided logs by adding clear output to the step to configure the installer.  
```

NONE

### DIFF
--- a/scripts/install.sh
+++ b/scripts/install.sh
@@ -365,7 +365,7 @@ function kubernetes_maybe_generate_bootstrap_token() {
 }
 
 function kurl_config() {
-    logStep "Configuring the installer ..."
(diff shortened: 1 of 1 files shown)


camilamacedo86 (commented):
@laverya all changes are addressed now. It seems good to fly.
WDYT?

rrpolanco (approved):

PR #4207, opened 2023-03-17
Fix invalid upgrade path with Cust1 Upgrade - no HA - Migrate from Rook to OpenEBS is only supported with versions >= 3.3.0 

camilamacedo86:

- Old wrong upgrade configuration is invalid and it is failing now that we added the preflight check.
- Also update the name of the test to make clear 
- Use latest 1.9.x to 1.21.x k8s version instead of pinned ones
- Improve the name of the test to be more meaningful
- Add postgrid test checks to validate migration 

More info: https://testgrid.kurl.sh/run/STAGING-daily-5f9db1a-2023-03-17T01:28:13Z?kurlLogsInstanceId=ewjsdujyuxvwatfy&nodeId=ewjsdujyuxvwatfy-initialprimary

### DIFF
--- a/testgrid/specs/full.yaml
+++ b/testgrid/specs/full.yaml
@@ -1468,7 +1468,7 @@
     echo "Kubeconfig was $KUBECONFIG"
     unset KUBECONFIG
     kubectl get namespaces
-- name: "Cust1 Upgrade - no HA"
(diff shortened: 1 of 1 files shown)


emosbaugh (approved):

camilamacedo86:
@ricardomaraschini I am get this one merged.
But  I understand that it will also be moved from this file with: https://github.com/replicatedhq/kURL/pull/4199

PR #4220, opened 2023-03-20
Always re-apply addons

rrpolanco:

Part of the kURL addon lifecyle includes an `addon_already_applied` phase which attempts to avoid deploying the addon again. For certain addons there is a complexity overhead to keep to use of this logic and in certain scenarios there are issues with simply just re-configuring an existing addon when the kURL installer is re-run which leads to an inconsistent state of the cluster.

Moreover, the `--force-reapply-addon` flag will no longer be supported as this will now be the default behavior:
```sh
$ cat install.sh | sudo bash -s force-reapply-addons
⚙  Running install with the argument(s): force-reapply-addons
Error: unknown parameter "force-reapply-addons"
```

Fixes  [sc-70122](https://app.shortcut.com/replicated/story/70122/addon-already-applied-has-potential-to-break-things-especially-when-upgrading-some-but-not-all-add-ons)

```release-note
The kURL installer will now force reapply addons by default deprecating the usage of the `force-reapply-addons` options. 
```

https://github.com/replicatedhq/kurl.sh/pull/950

### DIFF
--- a/addons/containerd/template/base/install.sh
+++ b/addons/containerd/template/base/install.sh
@@ -11,7 +11,11 @@ function containerd_pre_init() {
         cp "$src/kubeadm-init-config-v1beta2.yaml" "$DIR/kustomize/kubeadm/init-patches/containerd-kubeadm-init-config-v1beta2.yml"
     fi
 
-    containerd_host_init
+    # preserve containerd config
+    if addon_has_been_applied "containerd"; then
+        CONTAINERD_PRESERVE_CONFIG=1
+        log "Preserving containerd $CONTAINERD_VERSION config"
+    fi
 }
 
 function containerd_join() {
@@ -34,7 +38,18 @@ function containerd_install() {
 
     containerd_migrate_from_docker
 
-    install_host_packages "$src" containerd.io
+    # only install if package is not present
+    if [ "$CONTAINERD_PRESERVE_CONFIG" = "1" ]; then
+        log "Skipping host package installation for containerd: $CONTAINERD_VERSION already intalled"
+    else
+        install_host_packages "$src" containerd.io
+    fi
+
+    case "$LSB_DIST" in
+        centos|rhel|amzn|ol)
+            yum_install_host_archives "$src" libzstd
+            ;;
+    esac
 
     chmod +x ${DIR}/addons/containerd/${CONTAINERD_VERSION}/assets/runc
     # If the runc binary is executing the cp command will fail with "text file busy" error.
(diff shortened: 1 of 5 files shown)


scripts/common/addon.sh
```diff
@@ -15,28 +15,17 @@ function addon_install() {
...
-        export REPORTING_CONTEXT_INFO="addon $name $version"
-        # shellcheck disable=SC1090
-        addon_source "$name" "$version"
-        # containerd is a special case because there is also a binary named containerd on the host
-        if [ "$name" = "containerd" ]; then
-            containerd_install
-        else
-            $name
-        fi
-        export REPORTING_CONTEXT_INFO=""
```
emosbaugh, line 39: i think you lost this line

addons/containerd/template/base/install.sh
```diff
@@ -30,7 +36,12 @@ function containerd_install() {
 
     containerd_migrate_from_docker
 
-    install_host_packages "$src" containerd.io
+    # only install if package is not present
```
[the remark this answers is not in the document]
rrpolanco: I see. I'll change this so that `preserveConfig` continues to behave as before.

kurl_util/cmd/bashmerge/main.go
```diff
@@ -264,8 +264,6 @@ func parseBashFlags(installer *kurlv1beta1.Installer, bashFlags string) error {
 			continue
 		case "secondary-host":
 			continue
-		case "force-reapply-addons":
```
emosbaugh, line 267: can you leave this for backwards compatibility and add a comment?

scripts/common/utilbinaries.sh
```diff
@@ -175,9 +175,6 @@ function get_patch_yaml() {
                     SECONDARY_HOST="$SECONDARY_HOST,$_value"
                 fi
                 ;;
-            force-reapply-addons)
```
emosbaugh, line 178: can you leave this for backwards compatibility and add a comment?

emosbaugh (changes requested):
what happened to that CONTAINERD_PRESERVE_CONFIG logic? i quite liked that. i just dont think you can reappropriate the variable because our vendors are already using it.

emosbaugh (changes requested):
can you remove remove get_force_reapply_addons_flag again? appending that flag is only perpetuating the problem. i only asked that you add back the switch statements.

emosbaugh (approved):

PR #4222, opened 2023-03-20
fix: does to stop upgrade because was impossible to remove rook

camilamacedo86:

If in the upgrade we be unable to remove Rook we can still be able to do that manually afterwords or by re-calling the script. However, does not seems make sense we stop the cluster upgrade because did when we could check that we migrated the PVC(s), the ObjectStore successfully. 

Fixes # [sc-70239]

```release-note
Fixes script stop when the migration from Rook were done successfully but only was not possible to remove Rook. Note that it still possible to call the task `curl <installer>/task.sh | sudo bash -s remove_rook_ceph` to remove Rook manually or just re-run the script to re-try.
```

### DIFF
--- a/scripts/common/rook.sh
+++ b/scripts/common/rook.sh
@@ -216,7 +216,10 @@ function maybe_cleanup_rook() {
 
         if [ "$DID_MIGRATE_ROOK_PVCS" == "1" ] && [ "$DID_MIGRATE_ROOK_OBJECT_STORE" == "1" ]; then
             report_addon_start "rook-ceph-removal" "v1"
-            remove_rook_ceph
+            if ! remove_rook_ceph; then
+                logFail "Unable to remove Rook."
+                return
+            fi
             report_addon_success "rook-ceph-removal" "v1"
             return
         fi
--- a/scripts/common/rook.sh
+++ b/scripts/common/rook.sh
@@ -218,6 +218,7 @@ function maybe_cleanup_rook() {
             report_addon_start "rook-ceph-removal" "v1"
             if ! remove_rook_ceph; then
                 logFail "Unable to remove Rook."
+                report_addon_fail "rook-ceph-removal" "v1"
                 return
             fi
             report_addon_success "rook-ceph-removal" "v1"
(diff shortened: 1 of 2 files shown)


scripts/common/rook.sh
```diff
@@ -216,7 +216,10 @@ function maybe_cleanup_rook() {
 
         if [ "$DID_MIGRATE_ROOK_PVCS" == "1" ] && [ "$DID_MIGRATE_ROOK_OBJECT_STORE" == "1" ]; then
             report_addon_start "rook-ceph-removal" "v1"
-            remove_rook_ceph
+            if ! remove_rook_ceph; then
+                logFail "Unable to remove Rook."
+                return
+            fi
             report_addon_success "rook-ceph-removal" "v1"
```
emosbaugh, line 223: it does not seem like we should report successful here but there does not see any way to report a failure. @laverya what do you recommend?

camilamacedo86: That is a good point.

laverya: I swear we had a report_addon_failure command

laverya: congratulations, you can be the first user
https://github.com/replicatedhq/kURL/blob/4d146f4e2ac810f92376100041e2c6388e016203/scripts/common/reporting.sh#L144-L166
just make sure to get rid of the 'unused' comment 😆

laverya: actually that's probably not quite what you want
maybe make a copy of the https://github.com/replicatedhq/kURL/blob/4d146f4e2ac810f92376100041e2c6388e016203/scripts/common/reporting.sh#L104 function but with the fail_addon endpoint instead

scripts/common/rook.sh
```diff
@@ -216,7 +216,11 @@ function maybe_cleanup_rook() {
 
         if [ "$DID_MIGRATE_ROOK_PVCS" == "1" ] && [ "$DID_MIGRATE_ROOK_OBJECT_STORE" == "1" ]; then
             report_addon_start "rook-ceph-removal" "v1"
-            remove_rook_ceph
+            if ! remove_rook_ceph; then
+                logFail "Unable to remove Rook."
+                report_addon_fail "rook-ceph-removal" "v1"
```
emosbaugh, line 221: does this function exist?

emosbaugh (changes requested):

scripts/common/rook.sh
laverya: I like the new function

laverya (approved):

PR #4209, opened 2023-03-18
Fixes stuck issues scenarios when is trying to delete Rook and provide better information about the steps performed.  

camilamacedo86:
#### What this PR does / why we need it:

Currently we have the following problems that are addressed by this PR:
- The script can get stuck because it is missing timeout in the operations
- The script does not provide info/logs about what operation was or not possible to do to remove rook
- The script has not been checking and reporting the error properly when the Rook StorageClasses and Namespace is not removed

#### Which issue(s) this PR fixes:

Fixes # [sc-70239]

#### Special notes for your reviewer:

**(Test A) When is possible to remove Rook (Rook 1.0.4->To OpenEBS 3.3.0+Minio)**

- Install with:

```ymal
INSTALLER_YAML="apiVersion: cluster.kurl.sh/v1beta1
kind: Installer
metadata:
  name: 68f22f4
spec:
  kubernetes:
    version: 1.19.16
  weave: 
    version: 2.8.1-20230222
  rook:
    version: 1.0.4
  contour:
    version: 1.24.1
  registry:
    version: 2.8.1
  prometheus:
    version: 0.63.0-45.7.1
  ekco:
    version: 0.26.4
  kurl:
    installerVersion: v2023.03.13-0
    additionalNoProxyAddresses: []
  docker: 
    version: 20.10.17  
"
```

- Then, upgrade with the following spec and the changes made in this PR:

```yaml
INSTALLER_YAML="apiVersion: cluster.kurl.sh/v1beta1
kind: Installer
metadata:
  name: 68f22f4
spec:
  kubernetes:
    version: 1.19.16
  weave: 
    version: 2.8.1-20230222
  minio: 
    version: 2023-03-09T23-16-13Z
  openebs: 
    version: 3.3.0
    isLocalPVEnabled: true
    namespace: openebs
  contour:
    version: 1.24.1
  registry:
    version: 2.8.1
  prometheus:
    version: 0.63.0-45.7.1
  ekco:
    version: 0.26.4
  kurl:
    installerVersion: v2023.03.13-0
    additionalNoProxyAddresses: []
  docker: 
    version: 20.10.17  
"

``` 

Result: 

> See that now we have the data to be able to troubleshooting when failures occurs and this script will no longer get stuck or stop to be executed when is not possible to remove Rook so that is possible to just retry it via script OR remove Rook manually.

```sh
Waiting for ekco pods to be removed
Removing rook-ceph custom resource objects - this may take some time:\n
[...]

PR #4223, opened 2023-03-20
fix: ensure that Rook Data will be cleanup when rook is removed

camilamacedo86:

Currently, when we are removing Rook we are not removing the data. 
That can cause problems mainly if Rook be re-installed on the cluster. 
Also, uses space in disk which is no longer required. 

Fixes # [sc-70238]

See that we are also adding the check to the testgrid

```release-note
Fixes Rook data not been removed when Rook Ceph is removed from the cluster.
```

### DIFF
--- a/scripts/common/rook.sh
+++ b/scripts/common/rook.sh
@@ -86,6 +86,10 @@ function remove_rook_ceph() {
         fi
     fi
 
+    # More info: https://rook.io/docs/rook/v1.10/Getting-Started/ceph-teardown/#delete-the-cephcluster-crd
+    log "Allow RookCeph delete the data"
+    kubectl -n rook-ceph patch cephcluster rook-ceph --type merge -p '{"spec":{"cleanupPolicy":{"confirmation":"yes-really-destroy-data"}}}'
+
     # remove all rook-ceph CR objects
     printf "Removing rook-ceph custom resource objects - this may take some time:\n"
     kubectl delete cephcluster -n rook-ceph rook-ceph # deleting this first frees up resources
@@ -114,6 +118,13 @@ function remove_rook_ceph() {
             sed --expression='s/maintain_rook_storage_nodes:[ ]*true/maintain_rook_storage_nodes: false/g' | \
             kubectl -n kurl apply -f - 
         kubectl -n kurl scale deploy ekc-operator --replicas=1
+    fi
+
+    rm -rf /var/lib/rook || true
+    rm -rf /opt/replicated/rook || true
+
+    if [ -d "/var/lib/rook" ] || [ -d "/opt/replicated/rook" ]; then
+        logWarn  "Data within /var/lib/rook, /opt/replicated/rook and any bound disks has not been freed."
     fi
 
     # print success message
--- a/scripts/common/rook.sh
+++ b/scripts/common/rook.sh
@@ -129,7 +129,6 @@ function remove_rook_ceph() {
 
     # print success message
     printf "%bRemoved rook-ceph successfully!\n%b" "$GREEN" "$NC"
-    printf "Data within /var/lib/rook, /opt/replicated/rook and any bound disks has not been freed.\n"
 }
 
 # scale down prometheus, move all 'rook-ceph' PVCs to provided storage class, scale up prometheus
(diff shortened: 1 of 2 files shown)


testgrid/specs/full.yaml
```diff
@@ -1841,4 +1845,9 @@
...
     else
        echo "Namespace rook-ceph was removed"
-    fi
\ No newline at end of file
+    fi
+    echo "Verify if Rook data was removed"
+    if [ -d "/var/lib/rook" ] || [ -d "/opt/replicated/rook" ]; then
+        logWarn  "Data within /var/lib/rook, /opt/replicated/rook and any bound disks has not been freed."
+    fi
+
```
camilamacedo86, line 1853:
```suggestion

```

emosbaugh (changes requested):

scripts/common/rook.sh
```diff
@@ -116,6 +120,13 @@ function remove_rook_ceph() {
...
+    rm -rf /var/lib/rook || true
+    rm -rf /opt/replicated/rook || true
+
+    if [ -d "/var/lib/rook" ] || [ -d "/opt/replicated/rook" ]; then
+        logWarn  "Data within /var/lib/rook, /opt/replicated/rook and any bound disks has not been freed."
+    fi
+
     # print success message
     printf "%bRemoved rook-ceph successfully!\n%b" "$GREEN" "$NC"
     printf "Data within /var/lib/rook, /opt/replicated/rook and any bound disks has not been freed.\n"
```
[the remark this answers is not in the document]
camilamacedo86: Yes, we should.

emosbaugh:
I will wait for [this pr](https://github.com/replicatedhq/kURL/pull/4209) to be merged before reviewing.

scripts/common/rook.sh
```diff
@@ -133,9 +138,15 @@ function remove_rook_ceph() {
         kubectl -n kurl scale deploy ekc-operator --replicas=1
     fi
 
+    rm -rf /var/lib/rook || true
+    rm -rf /opt/replicated/rook || true
```
emosbaugh, line 142: I think this looks good but I'd like @laverya 's opinion as to whether he feels it is safe to remove the data here if potentially other operations failed above with only a logWarn

camilamacedo86: From the docs; https://rook.io/docs/rook/v1.10/Getting-Started/ceph-teardown/#delete-the-operator-and-related-resources

Screenshot 2023-03-20 at 21 58 05

Also, @laverya comment that would make sense we remove here: https://app.shortcut.com/replicated/story/70238/migration-from-rook-does-not-clean-the-data-stored-and-are-not-following-the-rook-teardown-docs#:~:text=Mar%2010%202023%2C%205%3A11%20pm

emosbaugh: yes. im just worried that there are plenty of conditionals above here that do not return when a command fails and instead log warn and continue. is it safe to delete the directories if we have not yet successfully removed rook and ceph?

laverya: I think to get here the namespace has to have been successfully deleted, so it's fine to delete the hostpaths

camilamacedo86: We only are in the place where we delete the data after:
a) we migrate 
b) we delete the namespace

The path : https://github.com/replicatedhq/kURL/pull/4223/files#diff-d1a3d4d46c348ae45d17405ff5db5f2a57f6de8264635200a462fb8a7ac9b7bdR91-R93

Will also allow the Operator remove the data. 
That seems sorted out the scenarios to remove Rook for upper versions either.

scripts/common/rook.sh
```diff
@@ -81,11 +81,17 @@ function remove_rook_ceph() {
...
     fi
+
     log "Removing rook-ceph Storage Classes"
     if ! kubectl get storageclass | grep rook | awk '{ print $1 }' | xargs -I'{}' kubectl delete storageclass '{}' --timeout=60s; then
         logFail "Unable to delete rook-ceph StorageClasses"
         return 1
     fi
+
+    # More info: https://rook.io/docs/rook/v1.10/Getting-Started/ceph-teardown/#delete-the-cephcluster-crd
+    log "Allow RookCeph delete the data"
```
laverya, line 92:
```suggestion
    log "Patch Ceph cluster to allow deletion"
```

scripts/common/rook.sh
emosbaugh: sounds good. approving

emosbaugh (approved):

camilamacedo86:
Moving forward since we got via the slack (private)  the OK from @laverya as well.

PR #4226, opened 2023-03-20
chore: Use v1.1 instead of v1 to update the report metric when rook delation fails

camilamacedo86:


scripts/common/rook.sh
```diff
@@ -252,7 +252,7 @@ function maybe_cleanup_rook() {
                 report_addon_fail "rook-ceph-removal" "v1"
                 return
             fi
-            report_addon_success "rook-ceph-removal" "v1"
+            report_addon_success "rook-ceph-removal" "v1.1"
```
laverya, line 255: All three of these need to be changed - lines 249 and 252

camilamacedo86: Thanks now I got.
We will have a new version of the report after it get merged ,

laverya (approved):

laverya (approved):

scripts/common/rook.sh
rrpolanco: why the version bump?

PR #4237, opened 2023-03-23
Adds preflight checks to prevent migrating from Rook to OpenEBS with Registry without MinIO

camilamacedo86:

To prevent migrating from Rook to OpenEBS with Registry without MinIO

Fixes # [sc-70234]

```release-note
Adds preflight checks to prevent migrating from Rook to OpenEBS with Registry without MinIO
```


ricardomaraschini (commented):
I thought that the installer would make the Registry a PVC if no Object Store is available. Isn't that true ?

rrpolanco (approved):

ricardomaraschini (approved):

PR #4243, opened 2023-03-23
skip removing docker if it is not running

laverya:
Currently, if docker is installed on the system but not running, the uninstallation process will fail as the `docker system prune` command will fail.

This PR avoids the failure by checking to see if the docker service is running. If the service is not running, it ensures the service is disabled and skips the rest of the docker uninstallation.

Fixes #

```release-note

```


scripts/common/docker.sh
```diff
@@ -98,10 +98,17 @@ function uninstall_docker() {
     if [ "$(docker ps -aq | wc -l)" != "0" ] ; then
         docker ps -aq | xargs docker rm -f || true
     fi
+
+    # if docker is not running, skip this process
+    if ! systemctl is-active --quiet docker; then
+        echo "Docker not running, not uninstalling."
+        systemctl disable docker.service
+        return
```
emosbaugh, line 106: why return here and not remove docker and uncordon the node etc...

emosbaugh (changes requested):

scripts/common/docker.sh
laverya: yeah probably a better solution

emosbaugh (approved):

camilamacedo86 (approved):

PR #4247, opened 2023-03-24
Replace troubleshoot.io with troubleshoot.sh

xavpaice:
Relates to: https://github.com/replicatedhq/troubleshoot/issues/1070

This changes references to `troubleshoot.io` in Support Bundle types to `troubleshoot.sh`.  The reason for this is that the Troubleshoot project does not own troubleshoot.io.

Fixes [Troubleshoot #1070](https://github.com/replicatedhq/troubleshoot/issues/1070)

Note that this changes the label stored on secrets that contain support bundles.  This should be reviewed alongside https://github.com/replicatedhq/troubleshoot/pull/1075

```release-note
Support-bundle specs stored in-cluster prior to this change were stored with a label `"troubleshoot.io/kind=support-bundle"`.  The label was used as a default setting for the Troubleshoot command to search for support bundles.  As of release 0.60.0, Troubleshoot searches for `"troubleshoot.sh/kind=support-bundle"`.  The label has been updated in this change to match the new search value.
```

NONE


banjoh (commented):
This is going to be a tricky upgrade. What happens if an older installer is used, say a versioned one such as `https://kurl.sh/version/v2023.03.21-0/latest`? The installer points to bundled `support-bundle` and `preflight` binaries that end up installed as `kubectl` plugins.

Making this changes would mean discovery using older troubleshoot binaries will seize to work.

xavpaice:
I've updated the change to preserve the existing labels, and add a new one

xavpaice:
https://github.com/replicatedhq/troubleshoot/pull/1203 added the default support for the 'old' label as well as the corrected one, this PR is now safe to add to kURL.

PR #4250, opened 2023-03-24
fix: Fixes log message upgrading from Rook 1.0.4 to 1.4.9 that caused the Rook Ceph bug

camilamacedo86:
Fixes log message upgrading when upgrading from Rook 1.0.4 to 1.4.9 that caused the Rook Ceph preflight health check to incorrectly report that Ceph was unhealthy because Ceph version information could not be found. This issue was caused by a bug in Rook Ceph versions earlier than 1.4.8.

Same of : https://github.com/replicatedhq/kURL/pull/4211/files

Fixes #

### DIFF
--- a/pkg/rook/upgrade.go
+++ b/pkg/rook/upgrade.go
@@ -55,6 +55,10 @@ func hasRookOrCephVersion(deployments *appsv1.DeploymentList, desiredVersion str
 	for _, dep := range deployments.Items {
 		rookVer := normalizeRookVersion(dep.Labels[labelKey])
 		if rookVer != desiredVersion {
+			if strings.Contains(rookVer, "0.0.0") || len(rookVer) == 0 {
+				// Ignore this scenario because Rook versions < 1.4.8 has a bug where the version is not set.
+				continue
+			}
 			_, ok := oldVersions[rookVer]
 			if !ok {
 				oldVersions[rookVer] = []string{dep.Name}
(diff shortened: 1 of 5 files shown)


camilamacedo86:
Hi @emosbaugh,

 > Could you add this logic to wait-for-version go command rather than allowing that command to timeout?
In this case,I think we should not. The main reasons are:
- The goal of this code here is just output the result by moving to the command we will not able to do achieve it easily
- Also, we are doing with shell the same check across all project. So if we would change in POV that should be an horizontally change. 
- This pr actually just improve the log (very nit small silly thing )

Therefore, the effort vs the value that it would bring here does not seems to pay off.

emosbaugh:
> Hi @emosbaugh,
> 
> > Could you add this logic to wait-for-version go command rather than allowing that command to timeout?
> > In this case,I think we should not. The main reasons are:
> 
> * The goal of this code here is just output the result by moving to the command we will not able to do achieve it easily
> * Also, we are doing with shell the same check across all project. So if we would change in POV that should be an horizontally change.
> * This pr actually just improve the log (very nit small silly thing )
> 
> Therefore, the effort vs the value that it would bring here does not seems to pay off.

I have two issues with this.

first, you have essentially identified there is a bug in kurl (due to a bug in rook) and you are stating that rather than fix it you will log.

second, i am unclear why fixing the command is not a simple fix. from my perspective, it should presumably take less time than pushing this change through which i fundamentally disagree with.

pkg/rook/upgrade.go
```diff
@@ -55,6 +55,10 @@ func hasRookOrCephVersion(deployments *appsv1.DeploymentList, desiredVersion str
 	for _, dep := range deployments.Items {
 		rookVer := normalizeRookVersion(dep.Labels[labelKey])
 		if rookVer != desiredVersion {
+			if strings.Contains(rookVer, "0.0.0") || len(rookVer) == 0 {
```
emosbaugh, line 58: i think we should target the exact bug. what is the label? is it empty or v0.0.0-0?

emosbaugh (changes requested):

emosbaugh (approved):

PR #4260, opened 2023-03-24
(Testing) fixes log if previous version of kURL was not find and continue

camilamacedo86:

If be unable to obtain the previous version used then, we just log and continue. 

Related to # https://github.com/replicatedhq/kURL/issues

```release-note
Fixes problem faced to check the previous kURL version used to do the install. If the installer be unable to obtain this information then it will be outputted and the script will not fail because of it. 
```


camilamacedo86:
Done in : https://github.com/replicatedhq/kURL/pull/4263/files

Issue #4259, opened 2023-03-24
Kurl script fails on machine with kubectl pre-installed


namgk:
Pleas see https://github.com/replicatedhq/kURL/pull/4208#issuecomment-1483195165

To repro, run this in a machine with kubectl already installed: 

    curl https://kurl.sh/662cd72 | sudo bash

And here's the error:

```
⚙  Running install with the argument(s):
Package kurl-bin-utils-v2023.03.21-0.tar.gz already exists, not downloading
SELinux is not installed: no configuration will be applied
W0324 16:27:25.392983    3826 loader.go:221] Config not found: /etc/kubernetes/admin.conf
The connection to the server localhost:8080 was refused - did you specify the right host or port?
W0324 16:27:25.420258    3838 loader.go:221] Config not found: /etc/kubernetes/admin.conf
The connection to the server localhost:8080 was refused - did you specify the right host or port?
An error occurred on line 4555
```

PR #4261, opened 2023-03-24
Add new check to avoid an install that is not supported

camilamacedo86:

We check in many places if k8s is installed or not already to know if is a new install or an upgrade.
The installer cannot support upgrades in a pre-existent installation that was not made with and if kubectl be installed previously we cannot either ensure the workflow since kurl is design to be used in a clean envinroment. 

Motivation: https://github.com/replicatedhq/kURL/issues/4259

Fixes #

## Steps to reproduce

Screenshot 2023-03-24 at 23 12 10

```release-note
Adds check to warning users that the installer should not be used to upgrade an Kubernetes installation that was not made with kURL and it might not work well if kubectl be pre installed on the machine already. 
```

### DIFF
--- a/scripts/common/preflights.sh
+++ b/scripts/common/preflights.sh
@@ -16,7 +16,21 @@ function preflights() {
     bail_when_no_object_store_and_s3_enabled
     bail_if_unsupported_openebs_to_rook_version
     bail_if_kurl_version_is_lower_than_previous_config
+    warn_if_kubectl_is_installed_without_kurl
     return 0
+}
+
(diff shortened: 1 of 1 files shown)


emosbaugh (commented):
In non happy path scenarios or non standard use cases this could be disruptive. I'm not sure we should proceed with this change

camilamacedo86:
> In non happy path scenarios or non standard use cases this could be disruptive. I'm not sure we should proceed with this change
Closing this one based on the feedback

PR #4265, opened 2023-03-24
chore: fix testgrid docker specs are not running on rhel 8

emosbaugh:

```
2023-03-24 20:23:36+00:00 ⚙  Running install with the argument(s): airgap
2023-03-24 20:23:36+00:00 Docker 20.10.17 is not supported on centos 8.4.
2023-03-24 20:23:36+00:00 The containerd addon is recommended. https://kurl.sh/docs/add-ons/containerd
2023-03-24 20:23:36+00:00 Continue? (y/N) main: line 4824: /dev/tty: No such device or address
+ KURL_EXIT_STATUS=1
+ export KUBECONFIG=/etc/kubernetes/admin.conf
+ KUBECONFIG=/etc/kubernetes/admin.conf
+ export PATH=/sbin:/bin:/usr/sbin:/usr/bin:/usr/local/bin
+ PATH=/sbin:/bin:/usr/sbin:/usr/bin:/usr/local/bin
+ '[' 1 -eq 0 ']'
+ echo ''
+ echo 'failed kurl install with exit status 1'
```

also yaml `yes` is `true` not `"yes"`

### DIFF
--- a/addons/containerd/template/testgrid/k8s-ctrd.yaml
+++ b/addons/containerd/template/testgrid/k8s-ctrd.yaml
@@ -20,6 +20,7 @@
       version: latest
 
 - name: "Migrate from Docker to Containerd and Kubernetes from 1.23 to 1.25"
+  flags: "yes"
   installerSpec:
     kubernetes:
       version: 1.23.x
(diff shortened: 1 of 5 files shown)


rrpolanco:
>also yaml yes is true not "yes"

Good catch!

PR #4269, opened 2023-03-25
fix storage testgrid : Migrate from Rook 1.0.4 to OpenEBS without minio cannot have registry

camilamacedo86:
Migrate from ROOK to OpenEBS with Registry set requires Minio.
Therefore, those tests fails in the preflight host checks done with: https://github.com/replicatedhq/kURL/blob/530779278521ef51285d7e7374d97bffbb811813/scripts/common/preflights.sh#L668-L670

The upgrade will NOT be done at all. In this way we must to test the following workflows without registry because those are support:
- Migrate from Rook 1.0.4 to OpenEBS (S3 disabled)
- Migrate from Rook 1.10.x to OpenEBS (S3 disabled)

This PR remove the registry from the above specs to install and upgrade and ams to remove the specific registry checks done after the install and upgrade.


ricardomaraschini (dismissed):
Some of the tests you have updated have, as a post install and upgrade validations, an attempt to push and pull an image in the registry. These validations are going to fail.

testgrid/specs/storage-migration.yaml
```diff
@@ -252,9 +248,7 @@
       localPVStorageClassName: local
   postInstallScript: |
     source /opt/kurl-testgrid/testhelpers.sh
-    create_deployment_with_mounted_volume "migration-test" "default" "/data" "registry:2.8.1"
     create_random_file_and_upload_to_deployment "migration-test" "default" "./test.data" "/data/test.data"
-    test_push_image_to_registry
     install_and_customize_kurl_integration_test_application
   postUpgradeScript: |
     source /opt/kurl-testgrid/testhelpers.sh
```
laverya, line 254: you need to remove test_pull_image from this

testgrid/specs/storage-migration.yaml
```diff
@@ -336,8 +323,6 @@
       version: 1.6.x
     flannel:
       version: 0.21.x
-    registry:
-      version: 2.8.1
```
laverya, line 340: this test case absolutely should work and does not require S3, _even with registry_

I do not support removing registry from this test

If the preflight is catching this, the preflight is too restrictive

testgrid/specs/storage-migration.yaml
```diff
@@ -1044,7 +1010,7 @@
       shouldDisableRestartFailedEnvoyPods: false
       envoyPodsNotReadyDuration: 5m
   unsupportedOSIDs:
-  - ubuntu-2204 # containerd 1.4.x is not supported on ubuntu 22.04
+    - ubuntu-2204 # containerd 1.4.x is not supported on ubuntu 22.04
```
camilamacedo86, line 1013:
```suggestion
  - ubuntu-2204 # containerd 1.4.x is not supported on ubuntu 22.04
```

testgrid/specs/storage-migration.yaml
camilamacedo86: @laverya we discussed that. I proposed change the check and you and @emosbaugh do not agree with it.
See: https://replicated.slack.com/archives/C0499EGNPNY/p1679944354578159

testgrid/specs/storage-migration.yaml
camilamacedo86: Is not it done already? Are we removing ` test_push_image_to_registry` should we remove any other test/check done in this scenario?

testgrid/specs/storage-migration.yaml
laverya: The appropriate check here isn't "is kots running with s3override"

it's "is there currently an object store that we would be migrating Registry away from"

testgrid/specs/storage-migration.yaml
laverya: it's currently 4 lines below this comment, on line 258. (why github won't let you comment on lines outside the fold I will never understand)

If there's no registry, you can't pull from the registry.

camilamacedo86: done

laverya (approved):

PR #4271, opened 2023-03-27
change check to allow Registry without Minio without kots or with kots and disable33 = true

camilamacedo86:

We are not allowing move from Rook when has Registry and minio is not selected.
This PR changes the check to not allow it just if kots is select and disableS3 is not true. 

Fixes # [sc-70234]

```release-note

```


camilamacedo86:
It is only to show an option/test

PR #4299, opened 2023-03-29
fix(containerd): do not wait for k8s to report healthy if not yet installed

emosbaugh:

If the kubernetes binaries get installed by kurl but the script fails before we install kubernetes, the script will always fail on re-running as it will wait for the apiserver to report as healthy.

```
✔ Status checked successfully. No broken packages were found.
Restarting containerd...
Service containerd restarted.
unpacking registry.k8s.io/pause:3.6 (sha256:79b611631c0d19e9a975fb0a8511e5153789b4c26610d1842e9f735c57cc8b13)...done

error: stat /etc/kubernetes/kubelet.conf: no such file or directory
spent 5m attempting to run "kubectl --kubeconfig=/etc/kubernetes/kubelet.conf get nodes.v1." without success
```

Fixes NONE

```release-note
Fixed an issue that could cause re-running the install script to fail if the Kubernetes binaries are installed but the cluster was never installed or configured.
```

### DIFF
--- a/addons/containerd/1.6.19/install.sh
+++ b/addons/containerd/1.6.19/install.sh
@@ -74,13 +74,16 @@ function containerd_install() {
 
     load_images $src/images
 
-    if systemctl list-unit-files | grep -v disabled | grep -q kubelet.service; then
-        systemctl start kubelet
-        # If using the internal load balancer the Kubernetes API server will be unavailable until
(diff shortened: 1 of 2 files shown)


emosbaugh:
[sc-73334]

rrpolanco (approved):

ricardomaraschini (approved):

PR #4302, opened 2023-03-30
Revert "feat: support for iptables kube proxy mode"

emosbaugh:
Reverts replicatedhq/kURL#4298

This seems to be causing some trouble in testgrid

https://testgrid.kurl.sh/run/STAGING-daily-ef61bf7-2023-03-30T01:28:07Z


rrpolanco (approved):
Agree. We should hold off on this feature until we have confidence it works.

laverya (approved):

PR #4316, opened 2023-03-31
fix daily testgrid:  is missing minio

camilamacedo86:

The test `Upgrade to 1.26 minimal` fails with the error:

```
2023-03-31 09:32:22+00:00 This is a prerelease version of kotsadm and should not be run in production. Continuing because this is testgrid.
2023-03-31 09:32:22+00:00 KOTS with s3 enabled requires an object store.
2023-03-31 09:32:22+00:00 Please ensure that your installer also provides an object store with either the MinIO or Rook add-on.
+ KURL_EXIT_STATUS=1
+ '[' 1 -eq 0 ']'
+ echo ''
+ echo 'failed kurl upgrade with exit status 1'
+ collect_debug_info_after_kurl
+ '[' 1 -ne 0 ']'
+ echo 'kubelet status'
+ systemctl status kubelet
2023-03-31 09:32:22+00:00 
+ echo 'kubelet journalctl
```

Note that `Upgrade to 1.26 minimal aigap` either have minio in the upgrade.therefore, this PR adds minio to the upgrade spec which seems be missing.  For further info see: https://testgrid.kurl.sh/run/STAGING-daily-49d7bcf-2023-03-31T05:27:37Z

Screenshot 2023-03-31 at 11 46 50

### DIFF
--- a/testgrid/specs/full.yaml
+++ b/testgrid/specs/full.yaml
@@ -1295,6 +1295,8 @@
       version: latest
     ekco:
       version: latest
+    minio:
+      version: latest
 - name: "Upgrade to 1.26 minimal airgap"
   installerSpec:
     kubernetes:


ricardomaraschini (approved):

camilamacedo86 (commented):
Fossa is falling a timeout issue.
This PR does not add new deps and is a just a fix for the test so it could not introduce any related issue for what is checked by FOSSA.

PR #4318, opened 2023-03-31
migrate from longhorn with registry without object store should be blocked

camilamacedo86:

The same check introduced when we are migrating from rook should be made when we are migrating from longhorn since it is the same case scenario. 

https://github.com/replicatedhq/kURL/blob/530779278521ef51285d7e7374d97bffbb811813/scripts/common/preflights.sh#L668-L670

Fixes #

```release-note
Adds a preflight check to prevent unsupported migrations from Longhorn to OpenEBS versions earlier than 3.3.0 and without an object store when Registry is set.
```

### DIFF
--- a/scripts/common/preflights.sh
+++ b/scripts/common/preflights.sh
@@ -14,8 +14,9 @@ function preflights() {
     host_nameservers_reachable
     allow_remove_docker_new_install
     bail_when_no_object_store_and_s3_enabled
-    bail_if_unsupported_openebs_to_rook_version
     bail_if_kurl_pods_are_unhealthy
+    bail_if_unsupported_migration_from_rook_to_openebs
+    bail_if_unsupported_migration_from_longhorn_to_openebs
     bail_if_kurl_version_is_lower_than_previous_config
     return 0
 }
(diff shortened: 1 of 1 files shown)


scripts/common/preflights.sh
```diff
@@ -774,6 +775,28 @@ function bail_if_unsupported_openebs_to_rook_version() {
...
+            if kubectl get ns 2>/dev/null | grep -q longhorn-system; then
+                semverParse "$OPENEBS_VERSION"
+                # if $OPENEBS_VERSION is less than 3.3.0
+                if [ "$major" -lt "3" ] || { [ "$major" = "3" ] && [ "$minor" -lt "3" ] ; }; then
+                    logFail "The OpenEBS version $OPENEBS_VERSION cannot be installed."
+                    bail "OpenEBS versions less than 3.3.0 do not support migrations from Longhorn"
+                fi
+                # registry + openebs without rook requires minio
+                if [ -n "$REGISTRY_VERSION" ] && [ -z "$MINIO_VERSION" ]; then
+                    logFail "Migration from Longhorn with Registry required an object store."
```
rrpolanco, line 792:
```suggestion
                    logFail "Migration from Longhorn with Registry requires an object store."
```

ricardomaraschini (approved):

rrpolanco (approved):

camilamacedo86:
We are facing timeout issues with FOSSA. This PR does not add any dependency so should not be impacted by.
We are moving forward to be unblocked.

PR #4319, opened 2023-03-31
adds checks to ensure that longhorn and openebs are healthy prior migration

camilamacedo86:

Longhorn is deprecated. However, we should also check if it and OpenEBS are healthy prior the migrations. 

Fixes # [sc-71717]

```release-note
Adds checks to ensure that Longhorn and OpenEBS are healthy prior migration from Longhorn to OpenEBS.
```


rrpolanco (approved):

camilamacedo86:

PR #4320, opened 2023-03-31
fix storage testgrid : apply same rules migrate from longhorn

camilamacedo86:

The same rule applied to migrate from Rook without object store is now applied to Longhorn.
Therefore, we are doing here the same changes (removing registry and its specific checks ) from the tests where the upgrade has Registry but has not an Object Store.

Changed done for Migration from Rook
- Same remove Registry and checks : https://github.com/replicatedhq/kURL/pull/4269
- Follow up where we figure out that was missing an step: https://github.com/replicatedhq/kURL/pull/4317/files

Also, check the PR that adds the same check to Longhorn:
- https://github.com/replicatedhq/kURL/pull/4318

Fixes #

```release-note

```

### DIFF
--- a/testgrid/specs/storage-migration.yaml
+++ b/testgrid/specs/storage-migration.yaml
@@ -340,8 +340,6 @@
       version: 1.6.x
     flannel:
       version: 0.21.x
-    registry:
-      version: 2.8.1
     kotsadm:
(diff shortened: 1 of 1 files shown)


camilamacedo86:

PR #4321, opened 2023-03-31
adds health checks from longhorn to rook

camilamacedo86:

We need to check if longhorn and rook are healthy prior migrate.

Fixes # [sc-71717]

```release-note
Adds checks to ensure that Longhorn and Rook Ceph are healthy prior migration from Longhorn to Rook Ceph.
```


rrpolanco (approved):

camilamacedo86:

PR #4331, opened 2023-04-05
bug: downloading previous containerd minor version

ricardomaraschini:
If we don't download the installation fails. This PR also adjusts the function `containerd_patch_for_minor_version` to take the major version into account as well when calculating the latest patch for the previous minor version.

### DIFF
--- a/scripts/common/common.sh
+++ b/scripts/common/common.sh
@@ -434,6 +434,10 @@ function report_install_containerd() {
     containerd_evaluate_upgrade "$current_containerd_version" "$CONTAINERD_VERSION"
     for version in "${CONTAINERD_INSTALL_VERSIONS[@]}"; do
         logStep "Moving containerd to version v$version."
+        if [ "$version" != "$CONTAINERD_VERSION" ] && [ "$AIRGAP" != "1" ] ; then
+            log "Downloading containerd v$version."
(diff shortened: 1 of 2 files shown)


laverya (approved):

ricardomaraschini:
Remarks on the the failing check in this PR: FOSSA scan failure is not because of a dependency issue in this PR but due to a backend problem on FOSSA server side: https://github.com/replicatedhq/kURL/actions/runs/4620529557/jobs/8170757868?pr=4331

```
Error:  ----------
  An issue occurred

  >>> Relevant errors

    Error

      A connection to the FOSSA endpoint was established, but the service took too long to respond.

          Request: GET https://app.fossa.com/api/cli/organization

      This is a networking error.

      Networking errors are typically caused by actual network failure or a network appliance
      (e.g. a firewall) between the FOSSA CLI and the FOSSA backend.
      This means that often such errors are transient, or are caused by local network configuration.

      Trying again in a few minutes may resolve this issue.
      If this issue persists, please report a bug to FOSSA support at https://support.fossa.com/

      In your bug report, please include FOSSA's debug bundle file: fossa.debug.json.gz.

      You can generate debug bundle by using `--debug` flag, for example:
        fossa analyze --debug

      Traceback:
        (none)


Error: Process completed with exit code 1.
```

PR #4341, opened 2023-04-06
bug: allow installation of any containerd version in case of docker

ricardomaraschini:
if the node we are running the installer is running with docker we can skip the containerd upgrade restrictions check, we can simply install it.

### DIFF
--- a/scripts/common/containerd.sh
+++ b/scripts/common/containerd.sh
@@ -14,6 +14,12 @@ function containerd_patch_for_minor_version() {
     echo ""
 }
 
+# containerd_node_is_using_docker returns 0 if the current node is using docker as the container runtime.
+function containerd_node_is_using_docker() {
+    local node=$(hostname | tr '[:upper:]' '[:lower:]')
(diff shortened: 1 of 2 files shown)


ricardomaraschini:
I have quickly test this in a single and in a multi node cluster. Upgrades from Docker are back up.  FOSSA is still failing but no new dependency was added by this PR.

scripts/common/containerd.sh
```diff
@@ -14,6 +14,12 @@ function containerd_patch_for_minor_version() {
     echo ""
 }
 
+# containerd_node_is_using_docker returns 0 if the current node is using docker as the container runtime.
+function containerd_node_is_using_docker() {
+    local node=$(hostname | tr '[:upper:]' '[:lower:]')
```
emosbaugh, line 19: is it possible to use the utility function get_local_node_name?

emosbaugh (changes requested):

scripts/common/containerd.sh
ricardomaraschini: Sure, done.

emosbaugh (approved):

PR #4347, opened 2023-04-07
fix(preflights): do not run some preflights on join or upgrade

emosbaugh:

Separates init preflights from join and upgrade. Only run spec validation and pod health check preflights on init.

Fixes the following issue on upgrading secondary nodes due to pods being down as the node is cordoned and drained:

```
Awaiting 2 minutes to check kURL Pod(s) are Running
Kurl has unhealthy Pod(s). Check the namespace kurl. Restarting the pod may fix the issue.
```

Fixes the following issue on upgrading worker nodes due to 

Fixes NONE

This was never release so no release note is necessary.

```release-note
NONE
```

### DIFF
--- a/scripts/common/preflights.sh
+++ b/scripts/common/preflights.sh
@@ -1,4 +1,5 @@
 
+# preflights are run on all nodes for init.sh, join.sh, and upgrade.sh
 function preflights() {
     require64Bit
     bailIfUnsupportedOS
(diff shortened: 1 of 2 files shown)


laverya (approved):

emosbaugh:
No deps added. Ignoring failed license scan

PR #4350, opened 2023-04-07
delete previous ekco versions, and only keep one version in the future

laverya:

This removes all previous ekco versions, as they are no longer installable. 

Fixes #

```release-note

```


laverya:
I don't care if the PR license scan fails here because this is a _removal_ of code, not adding/changing dependencies

emosbaugh (approved):

PR #4362, opened 2023-04-11
handle dynamic storage storageclass creation

laverya:

This allows installing with two openebs storageclasses and no rook storageclass, while still installing the rook operator.

It will be used by the scaling from openebs to rook feature in the future.

(you could use the rook param with this change, but it would not be particularly useful until the ekco changes in the pipeline are completed)

Fixes #

```release-note

```

### DIFF
--- a/addons/ekco/template/base/install.sh
+++ b/addons/ekco/template/base/install.sh
@@ -71,6 +71,8 @@ function ekco() {
     local src="$DIR/addons/ekco/$EKCO_VERSION"
     local dst="$DIR/kustomize/ekco"
 
+    ekco_dynamicstorage
+
     ekco_create_deployment "$src" "$dst"
 
     if [ "$EKCO_SHOULD_INSTALL_REBOOT_SERVICE" = "1" ]; then
@@ -497,4 +499,15 @@ function ekco_load_images() {
 function ekco_generate_config_hash() {
     local dst="$1"
     md5sum "$dst/configmap.yaml" | awk '{ print $1 }'
+}
+
+function ekco_dynamicstorage() {
+    if [ "$ROOK_MINIMUM_NODE_COUNT" -gt "1" ]; then
+        # check if the rook storageclass name exists - if it does we've already migrated and should not recreate/update 'scaling'
+        # yes the env var for rook's storage class name is "STORAGE_CLASS" - this is not a typo
+        if kubectl get storageclasses.storage.k8s.io "$STORAGE_CLASS" >/dev/null 2>&1; then
+            return 0
+        fi
+        kubectl apply -f "$src/storageclass-scaling.yaml"
+    fi
 }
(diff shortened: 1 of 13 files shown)


addons/ekco/template/base/install.sh
```diff
@@ -464,3 +466,14 @@ function ekco_load_images() {
         find "$DIR/image-overrides" -type f | xargs -I {} bash -c "cat {} | ctr -a $(${K8S_DISTRO}_get_containerd_sock) -n=k8s.io images import -"
     fi
 }
+
+function ekco_dynamicstorage() {
+    if [ "$ROOK_MINIMUM_NODE_COUNT" -gt "1" ]; then
```
emosbaugh, line 471: what if ROOK_MINIMUM_NODE_COUNT is not set? will it be empty? in this case dont you need to check that there is a value?

```suggestion
    if [ -n "$ROOK_MINIMUM_NODE_COUNT" ] && [ "$ROOK_MINIMUM_NODE_COUNT" -gt "1" ]; then
```

addons/openebs/template/base/install.sh
```diff
@@ -291,6 +291,9 @@ function openebs_should_be_default_storageclass() {
     elif openebs_has_default_storageclass ; then
         # if there is already a default storage class that is not "$storage_class_name"
         return 1
+    elif [ "$ROOK_MINIMUM_NODE_COUNT" -gt "1" ]; then
```
emosbaugh, line 294:
```suggestion
    elif [ -n "$ROOK_MINIMUM_NODE_COUNT" ] && [ "$ROOK_MINIMUM_NODE_COUNT" -gt "1" ]; then
```

addons/rook/template/base/install.sh
```diff
@@ -86,6 +86,11 @@ function rook() {
     rook_operator_deploy
     rook_set_ceph_pool_replicas
     rook_ready_spinner # creating the cluster before the operator is ready fails
+
+    if [ "$ROOK_MINIMUM_NODE_COUNT" -gt "1" ]; then
```
emosbaugh, line 90:
```suggestion
    if [ -n "$ROOK_MINIMUM_NODE_COUNT" ] && [ "$ROOK_MINIMUM_NODE_COUNT" -gt "1" ]; then
```

addons/rook/template/base/install.sh
```diff
@@ -86,6 +86,11 @@ function rook() {
...
+        return 0 # do not create a ceph cluster if it should instead be managed by ekco
```
emosbaugh, line 91: below this is looks like there may be code that ekco will not manage. for example `rook_maybe_migrate_from_longhorn`

it looks like you took into account upgrades but im not entirely convinced upgrades will work in some scenarios.

laverya: agreed this really doesn't handle migrating from longhorn right now

emosbaugh (dismissed):

emosbaugh (approved):

emosbaugh (approved):

PR #4367, opened 2023-04-12
bump: upgrade replicatedhq/plumber to v2

ricardomaraschini:
Upgrades to the newer version of plumber as it provides other facilities that are going to be useful.


ricardomaraschini:
Closing this one. There is an even newer version for this package: v2.1.0.

PR #4370, opened 2023-04-12
allow automation to create k8s 1.27 versions

laverya:

This allows us to add k8s 1.27.

Fixes #

```release-note

```

### DIFF
--- a/packages/kubernetes/template/script.sh
+++ b/packages/kubernetes/template/script.sh
@@ -7,6 +7,12 @@ VERSIONS=()
 function find_available_versions() {
     docker build -t k8s - < Dockerfile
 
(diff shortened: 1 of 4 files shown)


laverya:
The updated automation was run and produced https://github.com/replicatedhq/kURL/pull/4371

packages/kubernetes/template/script.sh
rrpolanco, line 1: We should remove the 1.23 and 1.22 version discovery logic.

emosbaugh (changes requested):
can you separate testgrid full and deploy update into a different pr? im worried there will be some lag in supporting 1.27 and testgrid will fail until then

laverya:
@emosbaugh yeah can do - should I leave in the minimal-127 test though?

packages/kubernetes/template/script.sh
laverya: Probably! For now though, it doesn't harm anything to keep it in

emosbaugh (approved):

PR #4381, opened 2023-04-14
chore: standardize logging and example

emosbaugh:

This change proposes a standard for logging in the kurl cli. It tries to disambiguate between logging and command output. It does not change behavior, as log statements will continue to go to stderr with timestamps. Additionally it adds a debug logger for verbose output.

### DIFF
--- a/cmd/kurl/main.go
+++ b/cmd/kurl/main.go
@@ -15,7 +15,12 @@ func main() {
 	if err != nil {
 		log.Fatal(err)
 	}
-	err = cli.NewKurlCmd(kurlCLI).ExecuteContext(ctx)
+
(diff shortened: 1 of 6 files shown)


ricardomaraschini (approved):
I liked it a lot.

rrpolanco (approved):
This was worthwhile refactor 👍

laverya:
Could we see some before+after log line examples?
Or is this not intended to change output?

emosbaugh:
> Could we see some before+after log line examples? Or is this not intended to change output?

There should be no change. The logger is instantiated the same as the log.DefaultLogger and includes timestamps.

camilamacedo86 (approved):

camilamacedo86 (approved):

PR #4394, opened 2023-04-18
Revert "chore(deps): bump github.com/replicatedhq/troubleshoot from 0.59.0 to 0.61.0"

camilamacedo86:
Reverts replicatedhq/kURL#4356

This version has an error which make the Block Device check fail always.
We cannot bump the next release 0.62.1 within this fix because it is using k8s 1.27 and we need a new controller-runtime release using k8s 1.27 either to be able to move forward.

fix [sc-74865]


laverya:
Should be resolved by https://github.com/replicatedhq/kURL/pull/4395

PR #4396, opened 2023-04-18
fix: k8s upgrades to non-step versions unsupported

emosbaugh:

Currently the step logic always stops at the step version for a given minor upgrade. This change makes it stop at the specified minor.


camilamacedo86 (approved):

camilamacedo86:
It seems OK for me. 
I think we will need to check it out in the daily testgrid afterwords.

laverya (approved):

PR #4407, opened 2023-04-20
Support kubeadm.k8s.io/v1beta2 API Removal with Kubernetes 1.26+

rrpolanco:

Use `kubeadm.k8s.io/v1beta3` api version for all kubeadm configuration types starting with Kubernetes 1.26+.
This change is backwards compatible with `kubeadm.k8s.io/v1beta2` kubeadm API.

Fixes [SC-74454](https://app.shortcut.com/replicated/story/74454/add-support-for-k8s-1-27)

Start with looking at `setup_kubeadm_kustomize()` in `yaml.sh`. This is where the kubeadm config types are conditionally copied depending on the api version.

```release-note

```


scripts/common/yaml.sh
```diff
@@ -100,19 +100,53 @@ twospace_ path: $patch_file"       "$kustomization_file"
...
-    mkdir -p $DIR/kustomize/kubeadm/init-patches
-    rm -rf $DIR/kustomize/kubeadm/join-patches
-    mkdir -p $DIR/kustomize/kubeadm/join-patches
+    rm -rf "$DIR/kustomize/kubeadm/init"
+    rm -rf "$DIR/kustomize/kubeadm/join"
+    rm -rf "$DIR/kustomize/kubeadm/init-patches"
+    rm -rf "$DIR/kustomize/kubeadm/join-patches"
+
+    # copy kubeadm kustomize resources
+    eval rsync -avr "$rsync_exclude" "$kubeadm_init_src/" "$kubeadm_init_dst"
```
emosbaugh, line 141: i dont think you can rely on rsync being available on the server. we have to be very careful not to use new unused commands or utilities

scripts/distro/kubeadm/distro.sh
```diff
@@ -196,4 +196,28 @@ EOF
 function kubeadm_api_is_healthy() {
     curl --globoff --noproxy "*" --fail --silent --insecure "https://$(kubernetes_api_address)/healthz" >/dev/null
 }
+
+function kubeadm_conf_api_version() {
```
emosbaugh, line 200: Did you find that KUBERNETES_TARGET_VERSION_MINOR is ever unset or not correct to use for this function? What about the variable KUBERNETES_VERSION? In other code that is how we get the intended k8s version. I don't think we use kubeadm version.

addons/containerd/1.6.20/kubeadm-join-config.yaml
```diff
@@ -1,4 +1,4 @@
-apiVersion: kubeadm.k8s.io/v1beta2
+apiVersion: kubeadm.k8s.io/$(kubeadm_conf_api_version)
```
emosbaugh, line 1: can you explain to me how the function invocation in this file is evaluated? it is unclear to me.

scripts/distro/kubeadm/distro.sh
rrpolanco: `KUBERNETS_TARGET_VERSION_*` vars always get parsed from the `KUBERNETES_VERSION` variable from [parse_kubernetes_target_version](https://github.com/replicatedhq/kURL/blob/main/scripts/common/utilbinaries.sh#L44). This function is invoked from the main scripts, `install.sh`, `join.sh`, etc.

addons/containerd/1.6.20/kubeadm-join-config.yaml
rrpolanco: I'm actually thinking of undoing this change and instead replace the api version of the addon patches at runtime. None of the current kubeadm conf addon patches use `v1beta2` specific fields so I think this is safe to do.

rrpolanco: `$(kubeadm_conf_api_version)` will be evaluated when the final kubeadm config file rendered: https://github.com/replicatedhq/kURL/blob/main/scripts/install.sh#L195

scripts/common/yaml.sh
rrpolanco: Makes sense. I re-wrote this logic using a loop to _not_ depend on `rsync`.

scripts/distro/kubeadm/distro.sh
```diff
@@ -196,4 +196,28 @@ EOF
...
+    
+    # Get version from runtime
+    # Enforce the use of kubeadm.k8s.io/v1beta api version beginning with Kubernetes 1.26+
+    local kubeadm_v1beta3_min_version=
+    kubeadm_v1beta3_min_version="26"
+    if [ -n "$KUBERNETES_TARGET_VERSION_MINOR" ]; then
```
rrpolanco, line 206: `KUBERNETES_TARGET_VERSION_MINOR` is always set so there is no need to check if it is _set_.

scripts/common/yaml.sh
```diff
@@ -100,22 +100,88 @@ twospace_ path: $patch_file"       "$kustomization_file"
...
+    else
+        # only include kubeadm/v1beta2 resources
+        kubeadm_exclude=("$kubeadm_cluster_config_v1beta3_file" "$kubeadm_init_config_v1beta3_file" "$kubeadm_join_config_v1beta3_file")
+    fi
+
+    # copy kubeadm kustomize resources
+    copy_kustomize_kubeadm_resources "$kubeadm_init_src" "$kubeadm_init_dst" "${kubeadm_exclude[@]}"
+    copy_kustomize_kubeadm_resources "$kubeadm_join_src" "$kubeadm_join_dst" "${kubeadm_exclude[@]}"
+
+    # tell kustomize which resources to generate
```
rrpolanco, line 137: Explain `eval` usage here.

scripts/kustomize/kubeadm/init-orig/kubeadm-cluster-config-v1beta2.yml
rrpolanco, line 1: Note on why this files is suffixed with `v1beta2`:
This is a kubeadm `ClusterConfiguration` type which uses v1beta2 api fields:
- `userHyperKubeImage`
- `dns.type`

scripts/distro/kubeadm/distro.sh
rrpolanco: @emosbaugh It turns out I need this check. It's a predicate to determine whether I need to grab the version from the installer runtime or from an existing cluster (e.g. the [weave-to-flannel migration task](https://github.com/replicatedhq/kURL/pull/4407/files#diff-ee341a571b062fdbc0e72cd063cfaef84629998edace18b631db9efcf2963809R797)).

emosbaugh (approved):

PR #4425, opened 2023-04-25
doc: add adr on new versioning scheme.

ricardomaraschini:
Adds a new ADR regarding the new kURL versioning scheme.

### DIFF
--- a/docs/arch/adr-010-new-versioning-schema.md
+++ b/docs/arch/adr-010-new-versioning-schema.md
@@ -11,13 +11,13 @@ Under the new versioning scheme, our offering will be organised into Channels th
 
 Consequently, we can initiate the design of our new kURL installation YAML in the following manner:
 
-```yaml=
+```yaml
 apiVersion: "cluster.kurl.sh/v1"
 kind: "Installer"
 metadata:
-	name: "cluster"
+  name: "cluster"
 spec:
-	channel:  "1.27”
+  channel:  "1.27”
 ```
 
 With the Channel (Kubernetes) now established as the basis for our delivery model, we can enable users to select a Version within this channel, which will allow them to access a specific set of Kubernetes-related features and functionalities (add-ons) that have been thoroughly tested and vetted by our team. This Version serves to abstract each individual add-on version into our own standardised version, ensuring that users can reliably and confidently utilise our offering without worrying about compatibility or interoperability issues.
@@ -28,10 +28,10 @@ We will establish the practice of starting each Channel with a version v1.0.0 th
 apiVersion: "cluster.kurl.sh/v1"
 kind: "Installer"
 metadata:
-	name: "cluster"
+  name: "cluster"
 spec:
-	channel:  "1.27"
-	version:  "1.0.0"
+  channel:  "1.27"
+  version:  "1.0.0"
 ```
 While we will delve into the specifics of what the new kURL Version inside a Channel entails in more detail shortly, for the sake of simplicity, bear in mind that this Version essentially groups together a list of add-ons that are compatible with a given Kubernetes version. As such, users can easily upgrade to a newer version within the same Channel by updating the YAML version field and running the installer accordingly. For instance, the following example illustrates how to update the cluster to kURL version v1.8.0 within 1.27 Channel:
 
@@ -39,10 +39,10 @@ While we will delve into the specifics of what the new kURL Version inside a Cha
 apiVersion: "cluster.kurl.sh/v1"
 kind: "Installer"
 metadata:
-	name: "cluster"
+  name: "cluster"
 spec:
-	channel:  "1.27"
-	version:  "1.8.0"
+  channel:  "1.27"
+  version:  "1.8.0"
 ```
 As we've designed it, users are free to skip as many versions as they wish within a given Channel without any restrictions. This means that they can easily move between kURL versions within the same Channel at their own pace, depending on their individual needs and preferences, without any limitations or requirements.
 
@@ -115,15 +115,15 @@ Let's exemplify: to make an Object Storage API available inside the cluster, cus
 apiVersion: "cluster.kurl.sh/v1"
 kind: "Installer"
 metadata:
-	name: "cluster"
+  name: "cluster"
 spec:
-	channel:  "1.27"
-	version:  "1.8.0"
-	objectStorage:
-		enabled: true
-		credentialsSecrets:
-		- namespace: "default"
-		  name: "object-storage"
+  channel:  "1.27"
+  version:  "1.8.0"
+  objectStorage:
+    enabled: true
+    credentialsSecrets:
+    - namespace: "default"
+      name: "object-storage"
 ```
 Expanding on the previous statement, the reason why the configuration would deploy Minio in the cluster is because it has been selected as the official Object Storage API provider for kURL. The `credentialsSecrets` property allows the user to specify where they expect to see the credentials required to access this Object Storage API. However, it is important to note that this is only an example and each individual configuration and selected add-ons will require further discussion and customisation.
   
--- a/docs/arch/adr-010-new-versioning-schema.md
+++ b/docs/arch/adr-010-new-versioning-schema.md
@@ -1,4 +1,4 @@
-# ADR 10: kURL new versioning scheme
+# ADR 10: New kURL Versioning Scheme
 
 ## Context
 Based on our prior experience, offering customers the ability to select various add-on versions provides them with significant flexibility, but it also creates numerous challenges. Presently, we're struggling to ensure that each customer's chosen add-ons are compatible with one another and that migration paths are available for all potential combinations. Moreover our user feedback indicates that stability and predictability are their top priorities, rather than the specific technologies they are running behind the scenes.
--- a/docs/arch/adr-010-new-versioning-schema.md
+++ b/docs/arch/adr-010-new-versioning-schema.md
@@ -4,7 +4,7 @@
 Based on our prior experience, offering customers the ability to select various add-on versions provides them with significant flexibility, but it also creates numerous challenges. Presently, we're struggling to ensure that each customer's chosen add-ons are compatible with one another and that migration paths are available for all potential combinations. Moreover our user feedback indicates that stability and predictability are their top priorities, rather than the specific technologies they are running behind the scenes.
 
 ## Decision
-Based on users feedback and the struggling we are facing to maintain kURL upgrades predictability, we have decided to adopt a more opinionated approach to delivering our Kubernetes distribution. This will involve implementing a well-designed versioning scheme, which is outlined in detail in this Architecture Decision Record.
+Based on users feedback and the challenges we are facing to maintain kURL upgrades predictable, we have decided to adopt a more opinionated approach to delivering our Kubernetes distribution. This will involve implementing a well-designed versioning scheme, which is outlined in detail in this Architecture Decision Record.
 
 ## Solution
 Under the new versioning scheme, our offering will be organised into Channels that correspond to specific Kubernetes minor versions. As of now, we will begin with a single channel named v1.27, which serves as our official offering for Kubernetes versions v1.27.x. Put simply, users who wish to install or keep their installations on Kubernetes v1.27 will remain on this Channel.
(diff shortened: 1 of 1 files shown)


docs/arch/adr-010-new-versioning-schema.md
```diff
@@ -0,0 +1,145 @@
...
+  version:  "1.8.0"
+```
+As we've designed it, users are free to skip as many versions as they wish within a given Channel without any restrictions. This means that they can easily move between kURL versions within the same Channel at their own pace, depending on their individual needs and preferences, without any limitations or requirements.
+
+_From now on this ADR may refer to Channel and Version in its short version "vChannel+kURL Version", for example v1.26+1.1.0 is the equivalent of Channel 1.26 and kURL version 1.1.0_
+
+### About kURL version inside a Channel
+As we've discussed earlier, a kURL Version within a given Channel comprises a list of add-ons, each with their own specific version. With that in mind, we could begin by outlining the initial group of add-ons that would make up kURL Version 1.0.0 within a given Channel 1.26. This version could includes the following set of add-ons, each with their corresponding version (this set of add-ons mentioned is used solely as an example and is not the definitive list of add-ons):
+
+| Add-on | Version |
```
emosbaugh, line 54: Does this also pin a kurl version? How can a user pin the kurl version?

docs/arch/adr-010-new-versioning-schema.md
```diff
@@ -0,0 +1,145 @@
...
+| 1.26    | 1.2.0  |
+| 1.27    | 1.0.0  |
+
+Based on the above-mentioned scenarios, the following table outlines various upgrade attempts and their respective outcomes:
+
+| From       | To        | Outcome |
+|------------|-----------|---------|
+|v1.25+1.0.0.|v1.25+1.2.0|Success. Upgrades inside the same Channel are allowed. |
+|v1.25+1.0.0.|v1.26+1.0.0|Failure. Users can only upgrade to the next Channel when they are already running the latest version of the current Channel. |
+|v1.25+1.2.0.|v1.26+1.2.0|Success. Upgrades from the latest version of the current Channel to any version on the next Channel are allowed. |
```
emosbaugh, line 102: I question whether we should allow upgrades to anything but the latest version in a channel. One thing is for sure we should not allow downgrading the kurl version. The question above impacts this.

docs/arch/adr-010-new-versioning-schema.md
```diff
@@ -0,0 +1,145 @@
...
+```yaml
+apiVersion: "cluster.kurl.sh/v1"
+kind: "Installer"
+metadata:
+  name: "cluster"
+spec:
+  channel:  "1.27"
+  version:  "1.8.0"
+  objectStorage:
+    enabled: true
```
emosbaugh, line 123: i believe it would make more sense to enable minio by default and allow the user to opt out. does it then make sense to invert this?

docs/arch/adr-010-new-versioning-schema.md
```diff
@@ -0,0 +1,145 @@
...
+    credentialsSecrets:
```
emosbaugh, line 124: i am not familiar with this feature. is this an actual proposed addition to the spec?

docs/arch/adr-010-new-versioning-schema.md
ricardomaraschini: That is indeed valid of a note. I think it is a valid idea.

ricardomaraschini: They way I was seeing this is (and I might have just now realised what you mean by pinning the kURL version) :

1) the cluster has a version.
2) the binary as a version.

Binary version `v1.27+1.2.0` installs or upgrades the current cluster to the version 1.2.0 on for Channel 1.27. Binary and configuration are two distinct things. The configuration only indicates what binary should be downloaded and is used by the binary to assess what needs to be enabled/disabled.

docs/arch/adr-010-new-versioning-schema.md
ricardomaraschini: Yet another case where I should have restricted myself to the scope of the task at hand. Sorry for that. This is, as I stated a few lines after this, and "example" on how we could manage add-on configurations in the installer. This configuration only exists on my head and such a feature would, in my head, be implemented in our operator.

docs/arch/adr-010-new-versioning-schema.md
```diff
@@ -0,0 +1,145 @@
...
+## Status
+
+Proposed
+
+## Consequences
+
+1. In order to plan the migration path from the current kURL version to the new one, we must determine the necessary versions that the customer must be running to transition to the first version in the initial channel.
+2. The vendor portal will need to be modified to accommodate this new arrangement.
+3. An API must be designed and developed in order to provide information about the current Channels and Versions available.
+4. All non-essential add-ons would need to be manually enabled by the user. Non-essential add-ons being anything other than Kubernetes, CNI and CSI.
```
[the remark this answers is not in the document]
ricardomaraschini: Yes, agreed. Let me just stay focused on the task at hand. I will remove it.

docs/arch/adr-010-new-versioning-schema.md
ricardomaraschini: As you mentioned in a comment below I think this whole section of the ADR is out of scope. I will remove it if you are OK with it.

emosbaugh: Ya i think you can remove it so we dont get hung up on the details here and prevent us from making a decision on versioning

ricardomaraschini: Removed. I am resolving the conversations about the subject.

docs/arch/adr-010-new-versioning-schema.md
```diff
@@ -0,0 +1,145 @@
...
+
+### Upgrades between Channels
+
+Upgrades between different Channels can possibly involve longer cluster downtime due to data migration. This may become very complex as some add-ons may not be supported in updated Kubernetes versions. Therefore, it's crucial for us to keep track of the current state of the cluster before staring the upgrade.
+
+To manage upgrades between Channels, we have decided to allow only one Channel version upgrade at a time. This means that customers can upgrade from Channel 1.25 to 1.26, but upgrades from 1.25 to 1.27 would be blocked. Additionally, users must be in the **latest** version within one Channel before upgrading to any version in the next Channel. By requiring users to be in the latest Channel version before upgrading to the next one, we can ensure two things: Firstly, we have a clear understanding of the current state of the cluster, which is essential for successful upgrades. Secondly, it provides us with some flexibility in preparing the cluster for the Channel upgrade, as we can introduce a new version in the previous Channel to gradually prepare the cluster to the next Channel.
```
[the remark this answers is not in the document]
emosbaugh: I would argue that multi-upgrades should be a separate automation as mentioned [here](https://github.com/replicatedhq/kURL/pull/4425/files#diff-7d4a25f424638f4928154398d50578dd6ede21ddee2af98f65c989704a3253bfR143) as that starts getting us back into the "testing every combination under the sun" territory.

docs/arch/adr-010-new-versioning-schema.md
```diff
@@ -0,0 +1,119 @@
+# ADR 10: kURL new versioning scheme
```
rrpolanco, line 1:
```suggestion
# ADR 10: New kURL Versioning Scheme
```

docs/arch/adr-010-new-versioning-schema.md
```diff
@@ -0,0 +1,119 @@
...
+We will establish the practice of starting each Channel with a version v1.0.0 that groups together the initial set of add-ons (and respective versions) we support for that particular version. As a result, we will need to update the YAML file to reflect this approach, an initial Installer would look like this:
+
+```yaml
+apiVersion: "cluster.kurl.sh/v1"
+kind: "Installer"
+metadata:
+  name: "cluster"
+spec:
+  channel:  "1.27"
+  version:  "1.0.0"
```
emosbaugh, line 34: im trying to think about how much sense it makes to tie these two versions in the same spec. the way i see it the version determines the binary installer version or the scripts version that is downloaded and then the channel or any other config will be used as configuration to the install command. it doesnt make a whole lot of sense to add the "version" to the latter. i dont have an answer for you only something to think about. for example:

```
curl https://get.kurl.sh | sudo VERSION=1.0.1 sh # this downloads binary version 1.0.1
kurl install --config=kurl-config.yaml # specifies the k8s version to use and the add-on config
```

docs/arch/adr-010-new-versioning-schema.md
emosbaugh: im only thinking about reproducibility here for the end user. we have to ensure both the homogeneity of the versions as well as reproducibility by pinning the binary or scripts version used to install k8s and add-ons.

docs/arch/adr-010-new-versioning-schema.md
```diff
@@ -0,0 +1,119 @@
...
+kind: "Installer"
+metadata:
+  name: "cluster"
+spec:
+  channel:  "1.27”
+```
+
+With the Channel (Kubernetes) now established as the basis for our delivery model, we can enable users to select a Version within this channel, which will allow them to access a specific set of Kubernetes-related features and functionalities (add-ons) that have been thoroughly tested and vetted by our team. This Version serves to abstract each individual add-on version into our own standardised version, ensuring that users can reliably and confidently utilise our offering without worrying about compatibility or interoperability issues.
+
+We will establish the practice of starting each Channel with a version v1.0.0 that groups together the initial set of add-ons (and respective versions) we support for that particular version. As a result, we will need to update the YAML file to reflect this approach, an initial Installer would look like this:
```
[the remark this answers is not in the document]
ricardomaraschini: Just to make sure we are on the same page here, the version 1.0.0 on Channel 1.26 would not be the same as the version 1.0.0 on Channel 1.27 (add-ons may differ on version and type, the logic may be different in the installer as may in the operator).

IIUC you are saying that we should be using, inside the Channels, "dates as versions" instead of semantic versions, if that is the case I would like to hear from others what they think .

/cc @rrpolanco @emosbaugh @kyleprice @camilamacedo86

docs/arch/adr-010-new-versioning-schema.md
ricardomaraschini: I don't see why one would need to set an environment variable. The way I was looking at this was:
```shell
$ curl https://get.kurl.sh/<sha>
```
The service answering this request would discover what Version from what Channel to provide the installer based on reading the `<sha>`.  Assuming that `<sha>` holds the following Installer:
```yaml
apiVersion: "cluster.kurl.sh/v1"
kind: "Installer"
metadata:
  name: "cluster"
spec:
  channel:  "1.27"
  version:  "1.8.0"
```
The service, on this case, would render and answer the following script (just to ilustrate):
```shell
#!/bin/bash

installer="
apiVersion: cluster.kurl.sh/v1
kind: Installer
metadata:
  name: cluster
spec:
  channel:  1.27
  version:  1.8.0
"

echo $installer > config.yaml
curl -O https://get.kurl.sh/release/1.27/1.8.0/kurl.bin
./kurl.bin --config ./config.yaml
```

The Channel and Version parts of the URL path (see the curl command above) would be replaced by the service answering to the initial call (I used `kurl.bin` so it is easier to understand that we are downloading a binary). 

When running the installer (`./kurl.bin --config ./config.yaml`) both values for `channel` and `version` are obsolete but we still need to include the installer as it may also contain add-ons configuration.

Another possibility of installation: manually downloading the release binary and then running it. On this case both `version` and `channel` are useless as those were already chosen by the user when downloading the binary.

In essence, the Version and Channel are useful only when a given vendor wants to pin their Installer to an specific release (most likely in the vendor portal).

ricardomaraschini:
@laverya @rrpolanco @kyleprice @emosbaugh @camilamacedo86 

To make the current installer to support a more opinionated installation we could use something like this: https://gist.github.com/ricardomaraschini/01f4cadaced8ff9956b768faf434bd24

ricardomaraschini:
I am closing this one as this is generating too much confusion. I have based this on the future kURL v2 and not in the current implementation. I need to completely rewrite it, this time thinking on how we gonna introduce the new versioning scheme with the current implementation.

ricardomaraschini:
This has been replaced by https://github.com/replicatedhq/kURL/pull/4430.

PR #4430, opened 2023-04-26
adr: introduces adr on new kurl versioning scheme

ricardomaraschini:
This ADR attempts to introduce the new kURL versioning scheme to the current kURL installer.

### DIFF
--- a/docs/arch/adr-010-new-versioning-scheme.md
+++ b/docs/arch/adr-010-new-versioning-scheme.md
@@ -1,0 +1,193 @@
+# ADR 10: New kURL Versioning Scheme
+
+## Context
+
+Based on our prior experience, offering customers the ability to select various add-on versions provides them with significant flexibility, but it also creates numerous challenges. Presently, we're struggling to ensure that each customer's chosen add-ons are compatible with one another and that migration paths are available for all potential combinations. Moreover our user feedback indicates that stability and predictability are their top priorities, rather than the specific technologies they are running behind the scenes.
+
+## Decision
+
+Based on users feedback and the challenges we are facing to maintain kURL upgrades predictable, we have decided to adopt a more opinionated approach to delivering our Kubernetes distribution. This will involve implementing a well-designed versioning scheme, which is outlined in detail in this Architecture Decision Record.
+
+## Solution
+
+The solution involves three different entities named _Channels_, _Installers_ and _Bundles_, each one of them are described below.
+
+### Choosing the Channel
+
+Our revised versioning scheme entails organising our offering into Channels that align with specific Kubernetes minor versions. As a result, users will be required to select the Channel to which they want their installer assigned.
+
+Although we won't initially provide a Channel for every Kubernetes minor version, we have arbitrarily selected Kubernetes minor version 1.24 as our starting point. At the time of writing, here are the Channels that we will be offering to our users:
+
+| Channel  | Kubernetes version installed by this Channel    |
+|----------|-------------------------------------------------|
+| v1.24    | Most up-to-date Kubernetes version for v1.24.x. |
+| v1.25    | Most up-to-date Kubernetes version for v1.25.x. |
+| v1.26    | Most up-to-date Kubernetes version for v1.26.x. |
+| v1.27    | Most up-to-date Kubernetes version for v1.27.x. |
+
+By drafting an installation file similar to the model given below, users can link their setup with a particular kURL Channel (_therefore they will be adhering to Kubernetes versions 1.27_)
+```yaml
+apiVersion: "cluster.kurl.sh/v1beta2"
+kind: "Installer"
+metadata:
+  name: "cluster"
+spec:
+  channel:  "1.27"
+```
+_We have decided to transition to the newer **v1beta2** variant. Existing users who prefer to continue working with the outdated format may still do so at their own discretion_
+
+#### Pinning the kURL version
+
+Another feature we aim to offer is the ability for users to fix (pin) their installation to a particular kURL installer version. The organisation of kURL installer releases follows the pattern `YYYY.MM.DD-x`. A complete list of kURL versions can be seen [here](https://github.com/replicatedhq/kURL/releases).
+
+For those seeking to pin their setup to a designated kURL installer version, the `installerVersion` field must be appropriately configured as shown in the example below.
+
+```yaml
+apiVersion: "cluster.kurl.sh/v1beta2"
+kind: "Installer"
+metadata:
+  name: "cluster"
+spec:
+  channel:  "1.27"
+  installerVersion: "v2023.04.24-0"
+```
+
+While it's certainly possible to omit the `installerVersion` configuration, doing so means relying upon the current most up-to-dated kURL installer version.
+
+### Add-on versions
+
+In order to maintain greater control over the installation process, we have decided to modify our approach to add-on selection. Going forward, we will no longer allow users to freely select any add-on on any version of their choosing. Instead, we will provide users with specific bundles of add-ons (from here on this ADR will call them _Bundles_), each of which will be available in different versions within a designated Channel.
+
+This is an example of a Bundle on its v1.0.0 version:
+
+| Add-on     | Version              |
+|------------|----------------------|
+| Containerd | 1.6.20               |
+| Flannel    | 0.21.5               |
+| OpenEBS    | 3.5.0.               |
+| Minio      | 2023-03-13T19-46-17Z |
+
+
+To facilitate the selection of a Bundle within a Channel, we will implement a new feature that enables users to pin a specific version of a Bundle. This means that users can select and install the precise version of a Bundle that best aligns with their needs, providing them with reproducibility.
+
+To illustrate the concept of Bundle version within a Channel, we have provided an example of one that could be available within the Channel v1.24. Let's call this _Bundle v1.0.0_:
+
+| Add-on     | Version              |
+|------------|----------------------|
+| Containerd | 1.6.16               |
+| Flannel    | 0.21.4               |
+| OpenEBS    | 3.3.0.               |
+| Minio      | 2023-03-13T19-46-17Z |
+
+For users who wish to pin their Bundle version within a Channel, we will implemented a straightforward solution. By providing the `bundleVersion` parameter as follows, users can easily specify the exact version of the add-on Bundle that they wish to install:
+
+```yaml
+apiVersion: "cluster.kurl.sh/v1beta2"
+kind: "Installer"
+metadata:
+  name: "cluster"
+spec:
+  channel:  "1.24"
+  bundleVersion: "1.0.0"
+```
+
+_If no `bundleVersion` parameter is passed, the default behaviour is to install or update to the most up-to-dated version of the Channel's Bundle_
+
+For this example let's say we have just released a new version of our Channel 1.24 Bundle, which is now available at version 1.1.0. This release includes several updates and bug fixes on the add-on offered. This is how the Bundle 1.1.0 may look like:
+
+| Add-on     | Version              |
+|------------|----------------------|
+| Containerd | 1.6.20               |
+| Flannel    | 0.21.4               |
+| OpenEBS    | 3.4.0                |
+| Minio      | 2023-03-13T19-46-17Z |
+
+_Versions for Containerd and OpenEBS were updated._ 
+
+If the Bundle version was pinned then the Installer will need to be updated in order for this to be installed in the cluster:
+
+```yaml
+apiVersion: "cluster.kurl.sh/v1beta2"
+kind: "Installer"
+metadata:
+  name: "cluster"
+spec:
+  channel:  "1.24"
+  bundleVersion: "1.1.0"
+```
+
+To ensure that the add-on Bundle is always up-to-date users could use the example installer below, which is specifically tailored to meet this need:
+
+```yaml
+apiVersion: "cluster.kurl.sh/v1beta2"
+kind: "Installer"
+metadata:
+  name: "cluster"
+spec:
+  channel:  "1.24"
+```
+_i.e. if no `bundleVersion` is present then the most up-to-dated one is installed_
+
+### Additional Details
+
+We will develop an approach that does not require any modifications to the kURL installer otherwise pinning old installer versions will result in a failure. Instead, we will implement a new feature in the kURL-API that allows us to seamlessly translate the new Installer (_Channel_, _Installer Version_ and _Bundle Version_) into the old Installer on-the-fly, and then inject it into the installer script.
+
+For example, let's supposed that this is the most up-to-dated Bundle in the Channel 1.24:
+
+| Add-on     | Version              |
+|------------|----------------------|
+| Containerd | 1.6.20               |
+| Flannel    | 0.21.4               |
+| OpenEBS    | 3.4.0                |
+| Minio      | 2023-03-13T19-46-17Z |
+
+Upon attempting to translate the following _new Installer_ and taking into account the Bundle above:
+
+```yaml
+apiVersion: "cluster.kurl.sh/v1beta2"
+kind: "Installer"
+metadata:
+  name: "cluster"
+spec:
+  channel:  "1.24"
+  installerVersion: "v2023.04.24-0"
+```
+
+The API will render the following _old Installer_ inside the script:
+
+```yaml
+apiVersion: "cluster.kurl.sh/v1beta1"
+kind: "Installer"
+metadata: 
+  name: "installer"
+spec:
+  kubernetes: 
+    version: "1.24.12"
+  flannel: 
+    version: "0.21.4"
+  containerd: 
+    version: "1.6.20"
+  minio: 
+    version: "2023-03-13T19-46-17Z"
+  openebs: 
+    version: "3.4.0"
+  kurl:
+    installerVersion: "v2023.04.24-0"
+```
+
+## Status
+
+Proposed
+
+## Questions to be answered in future interactions on this ADR
+
+- How one disables or enables specific add-ons inside the Bundle ?
+- How one can provide custom config to specific add-ons inside the Bundle ?
+- How are we going to manage OpenEBS vs Rook when both are of the Bundle ?
+	- We might need to finish the "auto-migration" work before.
+
+## Consequences
+
+- This new type needs to be stored by the kURL-API and translated into the original Installer (_v1beta1_) on the fly.
+- We gonna need to keep track of what add-ons belong to what Bundle version in what Channel.
+- There might be some heavy lifting in the vendor side to support new promoted installers using this method.


docs/arch/adr-010-new-versioning-scheme.md
```diff
@@ -0,0 +1,193 @@
...
+In order to maintain greater control over the installation process, we have decided to modify our approach to add-on selection. Going forward, we will no longer allow users to freely select any add-on on any version of their choosing. Instead, we will provide users with specific bundles of add-ons (from here on this ADR will call them _Bundles_), each of which will be available in different versions within a designated Channel.
+
+This is an example of a Bundle on its v1.0.0 version:
+
+| Add-on     | Version              |
+|------------|----------------------|
+| Containerd | 1.6.20               |
+| Flannel    | 0.21.5               |
+| OpenEBS    | 3.5.0.               |
+| Minio      | 2023-03-13T19-46-17Z |
```
ricardomaraschini, line 68: I know, there are more add-ons that this but I don't think that they are relevant to the example

docs/arch/adr-010-new-versioning-scheme.md
```diff
@@ -0,0 +1,193 @@
...
+- How one disables or enables specific add-ons inside the Bundle ?
+- How one can provide custom config to specific add-ons inside the Bundle ?
+- How are we going to manage OpenEBS vs Rook when both are of the Bundle ?
+	- We might need to finish the "auto-migration" work before.
+
+## Consequences
+
+- This new type needs to be stored by the kURL-API and translated into the original Installer (_v1beta1_) on the fly.
+- We gonna need to keep track of what add-ons belong to what Bundle version in what Channel.
+- There might be some heavy lifting in the vendor side to support new promoted installers using this method.
```
ricardomaraschini, line 193: I meant "Vendor Portal" here.

docs/arch/adr-010-new-versioning-scheme.md
```diff
@@ -0,0 +1,193 @@
...
+## Status
+
+Proposed
+
+## Questions to be answered in future interactions on this ADR
+
+- How one disables or enables specific add-ons inside the Bundle ?
+- How one can provide custom config to specific add-ons inside the Bundle ?
+- How are we going to manage OpenEBS vs Rook when both are of the Bundle ?
+	- We might need to finish the "auto-migration" work before.
```
ricardomaraschini, line 187: We can't start implementing anything until we answer these questions.

docs/arch/adr-010-new-versioning-scheme.md
```diff
@@ -0,0 +1,193 @@
...
+To illustrate the concept of Bundle version within a Channel, we have provided an example of one that could be available within the Channel v1.24. Let's call this _Bundle v1.0.0_:
+
+| Add-on     | Version              |
+|------------|----------------------|
+| Containerd | 1.6.16               |
+| Flannel    | 0.21.4               |
+| OpenEBS    | 3.3.0.               |
+| Minio      | 2023-03-13T19-46-17Z |
+
+For users who wish to pin their Bundle version within a Channel, we will implemented a straightforward solution. By providing the `bundleVersion` parameter as follows, users can easily specify the exact version of the add-on Bundle that they wish to install:
```
laverya, line 82: Why do we provide both `bundleVersion` and `installerVersion`?

Why not just tie the addon versions to the installer?

emosbaugh: I agree with Andrew here. Let's tie add-on versions to installer versions. Introducing the additional bundle version only adds more unnecessary variability and complexity.

docs/arch/adr-010-new-versioning-scheme.md
```diff
@@ -0,0 +1,193 @@
...
+
+## Solution
+
+The solution involves three different entities named _Channels_, _Installers_ and _Bundles_, each one of them are described below.
+
+### Choosing the Channel
+
+Our revised versioning scheme entails organising our offering into Channels that align with specific Kubernetes minor versions. As a result, users will be required to select the Channel to which they want their installer assigned.
+
+Although we won't initially provide a Channel for every Kubernetes minor version, we have arbitrarily selected Kubernetes minor version 1.24 as our starting point. At the time of writing, here are the Channels that we will be offering to our users:
```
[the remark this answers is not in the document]
emosbaugh: I'd prefer that the channels available to the user are those that are actively supported by Kubernetes upstream. For example, the channels that are available for kURL version v2023.04.26 are 1.24, 1.25, 1.26 and 1.27. On 2023-07-28 when 1.24 becomes EOL and 1.28 is released any kURL version after that date will support channels 1.25, 1.26, 1.27 and 1.28 until the next cycle.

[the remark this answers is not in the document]
emosbaugh: I like this as a forcing function for users to upgrade.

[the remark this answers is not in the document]
emosbaugh: If we choose to adopt this approach can we address what happens to a spec where there is no installerVersion and the channel falls out of support?

docs/arch/adr-010-new-versioning-scheme.md
```diff
@@ -0,0 +1,193 @@
...
+| v1.27    | Most up-to-date Kubernetes version for v1.27.x. |
+
+By drafting an installation file similar to the model given below, users can link their setup with a particular kURL Channel (_therefore they will be adhering to Kubernetes versions 1.27_)
+```yaml
+apiVersion: "cluster.kurl.sh/v1beta2"
+kind: "Installer"
+metadata:
+  name: "cluster"
+spec:
+  channel:  "1.27"
```
camilamacedo86, line 35: What channel means for kURL? 
Channel seems to be a valid term for the vendor portal and not kURL.
Also, I understand that we are here trying to create an spec to indicates what is the k8s version which should be installed and managed by kurl therefore, why not make that clear by using "`kubernetes: 1.27`"

docs/arch/adr-010-new-versioning-scheme.md
```diff
@@ -0,0 +1,193 @@
...
+For users who wish to pin their Bundle version within a Channel, we will implemented a straightforward solution. By providing the `bundleVersion` parameter as follows, users can easily specify the exact version of the add-on Bundle that they wish to install:
+
+```yaml
+apiVersion: "cluster.kurl.sh/v1beta2"
+kind: "Installer"
+metadata:
+  name: "cluster"
+spec:
+  channel:  "1.24"
+  bundleVersion: "1.0.0"
```
camilamacedo86, line 91: Would we have installerVersion and bundleVersion?
I think we need a installerVersion only.
In the future of KURL I imagine the CLI (brainstorming-kurl) applying k0s and then our kURL Operator (brainstorming-ekco). Therefore, the installer release could be tag version of the CLI. Then , we know what ekco as any other dep component which would be installed within.

@emosbaugh @laverya WDYT

emosbaugh: yes this is what i was thinking

docs/arch/adr-010-new-versioning-scheme.md
chris-sanders: I kind of like the naming of channel. It helps people better understand that the "kurl" portion of the version is the "minor" version and the "channel" is the major version. As soon as we started talking about these as channels internally it cleared up a lot of confusion. Carrying that same phrasing through the rest of kURL is surly going to help others too.

docs/arch/adr-010-new-versioning-scheme.md
ricardomaraschini: These is a list of consequences of this change, I want to collect your input about them:

1) We are choosing to deliberately change the current `installerVersion` semantic.
2) We gonna have to choose if the `installerVersion` X under 1.24 installs the same add-ons as the same `installerVersion` X under 1.27 (of if the bundled add-ons change following the Kubernetes channel).
3) We will need to "block" certain `installerVersions` to be used in very old channels.

docs/arch/adr-010-new-versioning-scheme.md
ricardomaraschini: It seems like we are converging into removing the "bundle version". I am resolving this conversation as it is duplicate of https://github.com/replicatedhq/kURL/pull/4430#pullrequestreview-1402221415. Please add your input there.

docs/arch/adr-010-new-versioning-scheme.md
camilamacedo86: We have vendor Channels I think it started to bring misleading.

docs/arch/adr-010-new-versioning-scheme.md
ricardomaraschini: I have updated the ADR to remove any reference to "Bundle", please take a look.

docs/arch/adr-010-new-versioning-scheme.md
ricardomaraschini: I have updated the ADR to use "Kubernetes" instead of "Channel", please take a look.

emosbaugh: I dont think we have settled yet on Kubernetes vs Channel. No need to revert back but lets continue to discuss

docs/arch/adr-010-new-versioning-scheme.md
emosbaugh: > 1. We are choosing to deliberately change the current installerVersion semantic.

we are using the same versioning as we have today in kurl

> 2. We gonna have to choose if the installerVersion X under 1.24 installs the same add-ons as the same installerVersion X under 1.27 (or if the bundled add-ons change following the Kubernetes channel).

I think we would like to leave ourselves open to the add-on versions not always being the same as we do not want a release of one kuberentes version to block a release of the others say if the upstream add-on did not yet support a version.

> 3. We will need to "block" newer installerVersions to be used in very old channels.

I think we have decided on the following rules to prevent an infinite matrix of test scenarios:

1. You can never downgrade the installerVersion or Kubernetes.
2. You can never move both the installerVersion and the Kubernetes minor version at the same time (You can either keep the installerVersion the same, and move to the next Kubernetes minor version, or keep the Kubernetes minor version the same and move to any newer installerVersion).
3. Once a Kubernetes version is EOL by the upstream, we will no longer release new installerVersions with it.

laverya: > 3. We will need to "block" newer installerVersions to be used in very old channels.

we already do this today - if you try to install k8s 1.26 with an installerVersion from 2022, the API will return an error

ricardomaraschini: > > 1. We are choosing to deliberately change the current installerVersion semantic.
> 
> we are using the same versioning as we have today in kurl

Yes, the same versioning but with a different meaning. Installer versions, as they are today, do not include add-ons.

> > 2. We gonna have to choose if the installerVersion X under 1.24 installs the same add-ons as the same installerVersion X under 1.27 (or if the bundled add-ons change following the Kubernetes channel).
> 
> I think we would like to leave ourselves open to the add-on versions not always being the same as we do not want a release of one kuberentes version to block a release of the others say if the upstream add-on did not yet support a version.

Let me see if I understood you: version `v2023.04.24-0` when installed on Kubernetes 1.24 may bring different add-ons than the same version but installed on Kubernetes 1.25 ?

> > 3. We will need to "block" newer installerVersions to be used in very old channels.
> 
> I think we have decided on the following rules to prevent an infinite matrix of test scenarios:
> 
> 1. You can never downgrade the installerVersion or Kubernetes.
> 2. You can never move both the installerVersion and the Kubernetes minor version at the same time (You can either keep the installerVersion the same, and move to the next Kubernetes minor version, or keep the Kubernetes minor version the same and move to any newer installerVersion).
> 3. You can only move 1 Kubernetes minor version at a time.
> 4. Once a Kubernetes version is EOL by the upstream, we will no longer release new installerVersions with it.

Could you point me to where this has been decided ? Is this part of a document somewhere ? I would like to include this here.

emosbaugh: Sorry I would have been more clear about what I meant when i said "we have decided". We discussed this ADR on our weekly internal meeting and I am only summarizing the decision. You can find the notes in 4/27/2023. 

> Let me see if I understood you: version v2023.04.24-0 when installed on Kubernetes 1.24 may bring different add-ons than the same version but installed on Kubernetes 1.25 ?

the hope is that all version are the same but we may have to leave ourselves open to this possibility in order to prevent upstream projects from delaying us from shipping. this is only hypothetical and we could make this a rule until we are forced to break it.

docs/arch/adr-010-new-versioning-scheme.md
```diff
@@ -0,0 +1,203 @@
...
+ │                │       │                │      │              │ │      │                │
+ │                │       │                │      │ v2023.04.01-0│ │      │                │
+ │                │       │                │      │              │ │      │                │
+ │                │       │                │      │              │ │      │                │
+ │                │       │                │      │              ▼ │      │                │
+ │                │       │                │      │ v2023.05.01-0  ├──────► v2023.05.01-0  │
+ └────────────────┘       └────────────────┘      └────────────────┘      └────────────────┘
+```
+
+_Upgrade paths are supported on both horizontal and vertical axis, users can migrate from Kubernetes 1.24 to any other version but the upgrade will be executed sequentially._
```
ricardomaraschini, line 184: As discussed with @kyleprice here we may want to support diagonals as well: from the latest `installerVersion` on _Kubernetes Minor Version_ `1.24` to any other greater or equal `installerVersion` on _Kubernetes Minor Version_ `1.25`.

docs/arch/adr-010-new-versioning-scheme.md
```diff
@@ -0,0 +1,203 @@
...
+```yaml
+apiVersion: "cluster.kurl.sh/v1beta2"
+kind: "Installer"
+metadata:
+  name: "cluster"
+spec:
+  kubernetes:  "1.27"
+```
+
+_We have decided to transition to the newer **v1beta2** variant for this CRD. Existing users who prefer to continue working with the outdated format may still do so at their own discretion_
```
rrpolanco, line 43: Will `v1beta1` be deprecated?

docs/arch/adr-010-new-versioning-scheme.md
ricardomaraschini: This has already been addressed in the ADR.

docs/arch/adr-010-new-versioning-scheme.md
```diff
@@ -0,0 +1,203 @@
...
+
+#### Choosing the Installer Version
+
+Another feature we aim to offer is the ability for users to fix (pin) their installation to a particular kURL _Installer Version_. The organisation of kURL installer releases follows the pattern `YYYY.MM.DD-x`. A complete list of kURL versions can be seen [here](https://github.com/replicatedhq/kURL/releases).
```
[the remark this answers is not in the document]
ricardomaraschini: I have added a comment according to what we have discussed in our meeting.

docs/arch/adr-010-new-versioning-scheme.md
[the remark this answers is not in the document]
ricardomaraschini: This has been decided and it is already here in the document: we will install the most up-to-date `installerVersion` for the provided Kubernetes Minor Version. Let me know if there is something missing or wrong.

docs/arch/adr-010-new-versioning-scheme.md
ricardomaraschini: This has already been addressed on the ADR. Let me know if that is OK or if there is something else that needs to be added.

emosbaugh: LGTM as is

docs/arch/adr-010-new-versioning-scheme.md
```diff
@@ -0,0 +1,235 @@
...
+We will develop an approach that does not require any modifications to the kURL installer otherwise pinning old installer versions will result in a failure. Instead, we will implement a new feature in the kURL-API that allows us to seamlessly translate the new Installer (_Kubernetes Minor Version_ and _Installer Version_) into the old Installer object on-the-fly, and then inject the result into the installer script.
+
+For example, let's supposed that this is the most up-to-date add-on group for `installerVersion` version `v2023.04.24-0` in the _Kubernetes Minor Version_ 1.24:
+
+| Add-on     | Version              |
+|------------|----------------------|
+| Containerd | 1.6.20               |
+| Flannel    | 0.21.4               |
+| OpenEBS    | 3.4.0                |
+| Minio      | 2023-03-13T19-46-17Z |
```
camilamacedo86, line 98: Minio is Addon that has many releases, so I understand that we would still upgrading this one for example per release. Am I right? Ditto so, should we not have the bundle description on the release notes of each k8s version supported ? 

What do you think ?

docs/arch/adr-010-new-versioning-scheme.md
```diff
@@ -0,0 +1,235 @@
...
+ │ v2021.09.01-0  ├───┐   │ v2021.09.01-0  │      │                │      │                │
+ │ v2022.10.01-0  │   │   │ v2022.10.01-0  │      │                │      │                │
+ │                │   │   │ v2022.11.01-0  │      │                │      │                │
+ │                │   │   │ v2022.12.01-0  │      │ v2022.12.01-0  │      │                │
+ │                │   └───► v2023.01.01-0  ├──┐   │ v2023.01.01-0  │      │                │
+ │                │       │                │  │   │ v2023.02.01-0  │      │                │
+ │                │       │                │  │   │ v2023.03.01-0  │      │                │
+ │                │       │                │  │   │ v2023.04.01-0  │      │                │
+ │                │       │                │  └───► v2023.05.01-0  ├──────► v2023.05.01-0  │
+ └────────────────┘       └────────────────┘      └────────────────┘      └────────────────┘
```
camilamacedo86, line 157: I loved that ^. 🥇

docs/arch/adr-010-new-versioning-scheme.md
```diff
@@ -0,0 +1,235 @@
...
+|---------------------------|-------------------------------------------------|
+| v1.24                     | Most up-to-date Kubernetes version for v1.24.x. |
+| v1.25                     | Most up-to-date Kubernetes version for v1.25.x. |
+| v1.26                     | Most up-to-date Kubernetes version for v1.26.x. |
+| v1.27                     | Most up-to-date Kubernetes version for v1.27.x. |
+
+By drafting an installation file similar to the model given below, users can link their cluster installation with a particular _Kubernetes Minor Version_ (_by doing so they will be adhering to Kubernetes versions 1.27_)
+
+```yaml
+apiVersion: "cluster.kurl.sh/v1beta2"
```
laverya, line 35: this should be v2beta1

docs/arch/adr-010-new-versioning-scheme.md
```diff
@@ -0,0 +1,235 @@
...
+_We have decided to transition to the newer **v1beta2** variant for this CRD. Existing users who prefer to continue working with the outdated format may still do so at their own discretion_
+
+#### Choosing the Installer Version
+
+Another feature we aim to offer is the ability for users to fix (pin) their installation to a particular kURL _Installer Version_. The organisation of kURL installer releases follows the pattern `YYYY.MM.DD-x`. A complete list of kURL versions can be seen [here](https://github.com/replicatedhq/kURL/releases).
+
+For those seeking to pin their setup to a designated kURL _Installer Version_, the `installerVersion` field must be appropriately configured as shown in the example below.
+
+```yaml
+apiVersion: "cluster.kurl.sh/v1beta2"
```
emosbaugh, line 52: this should be v2beta1

emosbaugh (changes requested):

camilamacedo86 (dismissed):
It seems great for the purpose of this PR. 

/LGTM

docs/arch/adr-010-new-versioning-scheme.md
ricardomaraschini: After our last meeting we have decided to adopt "Channel".

docs/arch/adr-010-new-versioning-scheme.md
ricardomaraschini: Solved.

docs/arch/adr-010-new-versioning-scheme.md
ricardomaraschini: Solved.

camilamacedo86 (dismissed):

emosbaugh (dismissed):

PR #4436, opened 2023-04-27
remove NDM from OpenEBS, as it is unused

laverya:

This removes OpenEBS's NDM and NDM operator, as they consume significant resources and are not required for LocalPV usage.

Fixes #

```release-note
Improves OpenEBS by removing NDM, reducing resource usage
```


addons/openebs/template/base/install.sh
```diff
@@ -420,3 +423,9 @@ function openebs_prompt_migrate_from_longhorn() {
         bail "Not migrating"
     fi
 }
+
+function openebs_cleanup_ndm() {
+    kubectl delete configmap -n "$OPENEBS_NAMESPACE" openebs-ndm-config 2>/dev/null  || true
```
emosbaugh, line 428: can you add --ignore-not-found

laverya: Oh yep that would be better

emosbaugh (approved):

PR #4479, opened 2023-05-11
testgrid - add install test with all addons

camilamacedo86:
We have not been testing the install with a configuration where all addons not deprecated are getting installed. 
Also, in this case kots is faling and we need to ensure and fix it. 

Fixes #

```release-note

```


camilamacedo86:
We added a test that makes it no longer be required see: https://github.com/replicatedhq/kURL/blob/0475a279d8334a628d1492b8888970fef31d0744/testgrid/specs/full.yaml#L2152-L2186

Closing this one.

PR #4488, opened 2023-05-12
WIP: ADR for kURL Add-on operator

rrpolanco:

A proposal for managing kURL add-ons via a Kubernetes operaor.

Fixes #

```release-note

```

### DIFF
--- a/docs/arch/adr-011-kurl-addon-operator.md
+++ b/docs/arch/adr-011-kurl-addon-operator.md
@@ -1,0 +1,96 @@
+# ADR 011: kURL Add-on Operator
+
+## Context
+
+kURL today is a set of bash scripts that configure and install a Kubernetes cluster in addition to a set of add-on components. The components are installed in sequence by applying highly customized Kubernetes YAML resources. The current architecture makes kURL hard to develop, maintain, and support.
+
+As an amalgam of bash scripts, the code base suffers from low cohesion and high complexity which results in the code being difficult to read and reason, and difficult to test. The installation is applied sequentially and each component is highly dependent on each other with no scopes to limit access and enforce separation. The project is architecturally a monolith without any functional separation of components and concerns.
+
+The installation of all components is done procedurally in the script. When there is a failure, it is difficult to understand what components have failed and how that affects the installation process as a whole.
+
+The components are installed in a proprietary fashion by applying highly customized Kubernetes YAML resources. They are hard to customize as the spec is proprietary and hard to extend because each customization must be passed through configuration to the end-user. Changes to this spec must be documented and are less discoverable. This method is not supported by the upstream which makes it difficult to pin addon installation issues on the upstream project and difficult to request support from the upstream.
+
+
+## Decision
+
+We will develop a Kubernetes [Operator](https://kubernetes.io/docs/concepts/extend-kubernetes/operator/) in _Go_ that will install, configure and manage add-on applications in the cluster. The operator will extend the Kubernetes API through Custom Resource Definitions (CRD) and will allow users to declaritively manage add-ons with a Custom Resource (CR). 
+
+## Solution
+
+The operator will be implemented with the following objectives in mind:
+
+- It will be scaffolded using [kubebuilder](https://book.kubebuilder.io/introduction.html).
+- It will be developed using [SOLID](https://en.wikipedia.org/wiki/SOLID) principles and operator best practices.
+- It will distribution agnostic. I.e. it will not be tightly coupled to the kURL Kubernetes distribution.
+- It will manage application add-ons _only_ and _exclude_ managing CRI and CNI resources.
+- It will install and confiure the KOTS admin console.
+- It can be installed on an existing [kURL](https://kurl.sh/) cluster with only Kubernetes, containerd and flannel.
+- The installation mechanism for add-ons will be through the use of Helm Charts.
+- By default the add-on configuration will be opinionated by us but we will enable the end-user to override the configuration for maximum flexibility. Vendors and users will be able to override the configuration using a standard [Helm values](https://helm.sh/docs/chart_template_guide/values_files/) file.
+- It will rely as much as possible on Helm for managing the lifecyle of the add-ons.
+
+### The `Installer` API
+
+The `Installer` API defines a resource for installing specific add-ons.
+
+```yaml
+apiVersion: cluster.kurl.sh/v2beta1
+kind: Installer
+metadata:
+  name: installer-sample
+spec:
+  channel: "1.27"
+  minio:
+    enabled: true
+    configOverride:
+        persistence:
+            enabled: true
+        replicas: 1
+        mode: standalone
+
+```
+
+### The Controllers
+
+The operator will have at minimum one controller which ensures the `Installer` CR is reconciled with the cluster periodically. The `Installer` controller (main controller) will delegate responsiblity for managing add-ons to other controllers. The reconcilation for the main controller will consist of determining the add-ons to be managed from the `Installer` CR and creating a controller for each add-on (addon-on controller).
+
+The add-on controller will be responsible for
+- verifying the requirements for the add-on are met
+- managing the add-on itself
+- ensure add-on is in desired state
+- update the add-on CR status
+
+### Managing the Add-ons
+
+How the add-ons will be managed is still an open question. We have a couple of options:
+
+_Use the Helm Controller provided by Flux_
+
+The [Flux Helm Controller](https://fluxcd.io/flux/components/helm/helmreleases/) affords us the following benefits:
+- Stable [HelmRelease API](https://fluxcd.io/flux/components/helm/api/)
+- Helm release [dependencies](https://fluxcd.io/flux/components/helm/helmreleases/#helmrelease-dependencies)
+- Support for Helm install, upgrade, test, rollback and uninstall
+- Performs Helm uninstall on `HelmRelease` CR removal
+- Chart values override\
+- Pull charts from Helm repository directly
+
+_Develop our own Homebrew Helm Controller_
+
+This approach affords us with a controller that meets our specific needs and avoids any third party software dependency. 
+
+
+
+
+## Status
+
+Proposed
+
+## Open Questions
+
+- How to allow full configuration of add-ons for advanced vendors/users?
+- Do we use an existing feature rich Helm controller like the one provided by [Flux](https://fluxcd.io/flux/components/helm/) or build our own?
+- How to structure the add-on to adhere to SOLID principles?
+
+## Consequences
+
+- A kURL CLI will be needed to install the operator dependencies (if any)


docs/arch/adr-011-kurl-addon-operator.md
```diff
@@ -0,0 +1,96 @@
...
+## Solution
+
+The operator will be implemented with the following objectives in mind:
+
+- It will be scaffolded using [kubebuilder](https://book.kubebuilder.io/introduction.html).
+- It will be developed using [SOLID](https://en.wikipedia.org/wiki/SOLID) principles and operator best practices.
+- It will distribution agnostic. I.e. it will not be tightly coupled to the kURL Kubernetes distribution.
+- It will manage application add-ons _only_ and _exclude_ managing CRI and CNI resources.
+- It will install and confiure the KOTS admin console.
+- It can be installed on an existing [kURL](https://kurl.sh/) cluster with only Kubernetes, containerd and flannel.
```
camilamacedo86, line 27: I think this one we do not know yet right.

docs/arch/adr-011-kurl-addon-operator.md
```diff
@@ -0,0 +1,96 @@
...
+- The installation mechanism for add-ons will be through the use of Helm Charts.
+- By default the add-on configuration will be opinionated by us but we will enable the end-user to override the configuration for maximum flexibility. Vendors and users will be able to override the configuration using a standard [Helm values](https://helm.sh/docs/chart_template_guide/values_files/) file.
```
camilamacedo86, line 29: This is another point that for remains totally uncertain. It is not as simple as it may initially appear. We need to address several questions to ensure that users can pass ANY Helm chart value and that we can effectively handle them:

- Are we suggesting that users input the values through a value.yaml file? If so, we need to define the process for providing and processing this file. How that would work within the Operator?
- What happens when the specifications of the Helm charts change from one version to another? Dealing with these changes introduces the possibility of encountering breaking changes and invalid values. How we will deal within in the upgrades? 
- How do we handle breaking changes and invalid values? We need to develop robust mechanisms for validation and ensure a smooth transition in such cases.

Considering the challenges involved, this desire to accommodate any Helm chart value raises concerns. It is unclear whether and how this goal can be achieved in a maintainable and feasible manner. So, we need further investigation to determine if it is realistically feasible in the short and long term, as well as to understand the implications, in my humble opinion (IMHO).

docs/arch/adr-011-kurl-addon-operator.md
```diff
@@ -0,0 +1,96 @@
...
+        persistence:
+            enabled: true
+        replicas: 1
+        mode: standalone
+
+```
+
+### The Controllers
+
+The operator will have at minimum one controller which ensures the `Installer` CR is reconciled with the cluster periodically. The `Installer` controller (main controller) will delegate responsiblity for managing add-ons to other controllers. The reconcilation for the main controller will consist of determining the add-ons to be managed from the `Installer` CR and creating a controller for each add-on (addon-on controller).
```
camilamacedo86, line 55: We cannot make a controller create a another controller. I think what you want to say would be like
```suggestion
The solution involves implementing a minimum of one controller, the Installer controller, to periodically reconcile the Installer custom resource (CR) with the cluster. The Installer controller will delegate the management of add-ons and other required processes, such as Storage Migrations, to separate controllers. During reconciliation, the main controller will determine the add-ons and processes specified in the Installer CR and create individual controllers for each of them. This approach ensures that each component, be it an add-on or a required process, is managed by a dedicated controller, following the concept of single responsibility.
```

OR 

```suggestion
The solution involves implementing a minimum of one controller, the Installer controller, to periodically reconcile the Installer custom resource (CR) with the cluster. The Installer controller will not only handle the reconciliation process but also create custom resources (CRs) to represent the add-ons and processes, such as Storage Migrations, within the cluster.

During reconciliation, the Installer controller will determine the add-ons and processes specified in the Installer CR. It will then create separate CRs to represent each add-on or process. For example, if there is a need to perform a Storage Migration, the Installer controller will create a specific CR to represent that process.

In addition to creating the CRs, the Installer controller will delegate the management of each CR to dedicated controllers. These controllers will be responsible for monitoring and maintaining the desired state of their associated CRs, following the concept of single responsibility.

By adopting this approach, the operator can ensure that the Installer controller creates and manages the necessary CRs, representing add-ons and processes like Storage Migrations, while allowing separate controllers to handle the management and enforcement of the desired state for each specific component.
```

docs/arch/adr-011-kurl-addon-operator.md
```diff
@@ -0,0 +1,96 @@
...
+How the add-ons will be managed is still an open question. We have a couple of options:
+
+_Use the Helm Controller provided by Flux_
+
+The [Flux Helm Controller](https://fluxcd.io/flux/components/helm/helmreleases/) affords us the following benefits:
+- Stable [HelmRelease API](https://fluxcd.io/flux/components/helm/api/)
+- Helm release [dependencies](https://fluxcd.io/flux/components/helm/helmreleases/#helmrelease-dependencies)
+- Support for Helm install, upgrade, test, rollback and uninstall
+- Performs Helm uninstall on `HelmRelease` CR removal
+- Chart values override\
```
camilamacedo86, line 74: That seems nice

docs/arch/adr-011-kurl-addon-operator.md
```diff
@@ -0,0 +1,96 @@
...
+- Pull charts from Helm repository directly
```
camilamacedo86, line 75: That also seems a good feature

docs/arch/adr-011-kurl-addon-operator.md
```diff
@@ -0,0 +1,96 @@
...
+
+## Status
+
+Proposed
+
+## Open Questions
+
+- How to allow full configuration of add-ons for advanced vendors/users?
+- Do we use an existing feature rich Helm controller like the one provided by [Flux](https://fluxcd.io/flux/components/helm/) or build our own?
+- How to structure the add-on to adhere to SOLID principles?
```
camilamacedo86, line 92: What is your concerns?

docs/arch/adr-011-kurl-addon-operator.md
```diff
@@ -0,0 +1,96 @@
...
+
+```yaml
+apiVersion: cluster.kurl.sh/v2beta1
+kind: Installer
+metadata:
+  name: installer-sample
+spec:
+  channel: "1.27"
+  minio:
+    enabled: true
```
camilamacedo86, line 44: HelmChart values can have versions pinned. Am I right?
So, how are we trying to allow the users just gave all possible Options provided by it?

docs/arch/adr-011-kurl-addon-operator.md
```diff
@@ -0,0 +1,96 @@
...
+
+
+## Status
+
+Proposed
+
+## Open Questions
+
+- How to allow full configuration of add-ons for advanced vendors/users?
+- Do we use an existing feature rich Helm controller like the one provided by [Flux](https://fluxcd.io/flux/components/helm/) or build our own?
```
camilamacedo86, line 91:
```suggestion
- Do we use an existing feature rich Helm controller like the one provided by [Flux](https://fluxcd.io/flux/components/helm/) or should we use Helm as dependency to install, upgrade and uninstall the helm charts?
```

I think we never will build Flux because the only functionality that we need is manage Helm Charts itself right.

docs/arch/adr-011-kurl-addon-operator.md
```diff
@@ -0,0 +1,96 @@
...
+
+The [Flux Helm Controller](https://fluxcd.io/flux/components/helm/helmreleases/) affords us the following benefits:
+- Stable [HelmRelease API](https://fluxcd.io/flux/components/helm/api/)
+- Helm release [dependencies](https://fluxcd.io/flux/components/helm/helmreleases/#helmrelease-dependencies)
+- Support for Helm install, upgrade, test, rollback and uninstall
+- Performs Helm uninstall on `HelmRelease` CR removal
+- Chart values override\
+- Pull charts from Helm repository directly
+
+_Develop our own Homebrew Helm Controller_
```
camilamacedo86, line 77: I am not sure if I can go along with because we will need to create the controllers to manage the AddOns.
The Helm part that we need is only to manage helm charts itself (install/uninstall/upgrade) all the other needs still in place indeed if we use Flux.

docs/arch/adr-011-kurl-addon-operator.md
```diff
@@ -0,0 +1,96 @@
...
+This approach affords us with a controller that meets our specific needs and avoids any third party software dependency. 
+
+
+
+
+## Status
+
+Proposed
+
+## Open Questions
```
camilamacedo86, line 88: IMO we need to answer more less the following questions to be able to know if we should or NOT adopt flux:

**Complexity:**

- While FluxCD provides a powerful feature set, configuring and using it can be complex. How much effort will be required to integrate and become proficient in Helm + Flux?
- Considering the extensive features offered by Flux, do we truly need all of them, or would only a subset be sufficient? Will the integration justify the learning curve and potential limitations?
- How complex will it be to test the solution with this dependency in local development and CI environments?

**Compatibility:**
- Which Kubernetes versions are supported by Flux? Are there any limitations?
- What are the supported Helm versions and the extent of Helm features that Flux can handle? 
- Can Flux handle Helm hooks effectively?
- How does Flux manage its release lifecycle, and what are the support policies?

**Automation Limitations**

Despite its automation capabilities, are there any aspects of the deployment process that Flux may not be able to automate effectively, requiring manual intervention or additional tooling. Therefore, will adopting Flux truly eliminate the need to use Helm as a dependency for functions such as installation, upgrade, uninstallation, or checking the status of a release? Or will we end up with both implementations, adding unnecessary complexities?

**Flux as a Dependency:**

- Can Flux be effectively used as a dependency within the Operator? Is it possible to create a Flux CR from the Operator and interact with Flux's internal functionality? Or are not they allowing importing their APIs? Is their code implementation exported for usage or not?

**What happens when we need bug fixes and RFEs if we adopt flux?

- If we face issues within the project, will Flux provide support for such cases, and if so, what areas do they support?
- In the event that we need a feature enhancement or bug fix in Flux to address our specific needs, how will that process work?

**Flux Features**
- Which specific features provided by Flux would be advantageous to adopt in our project?
- What are our requirements that would be fulfilled by adopting Flux? In other words, what functionalities would we not need to implement separately due to Flux integration?
- Are there any scenarios where using Flux could create limitations or challenges in managing HelmCharts or maintaining control over them?

docs/arch/adr-011-kurl-addon-operator.md
```diff
@@ -0,0 +1,96 @@
...
+- Stable [HelmRelease API](https://fluxcd.io/flux/components/helm/api/)
+- Helm release [dependencies](https://fluxcd.io/flux/components/helm/helmreleases/#helmrelease-dependencies)
+- Support for Helm install, upgrade, test, rollback and uninstall
+- Performs Helm uninstall on `HelmRelease` CR removal
+- Chart values override\
+- Pull charts from Helm repository directly
+
+_Develop our own Homebrew Helm Controller_
+
+This approach affords us with a controller that meets our specific needs and avoids any third party software dependency.
```
camilamacedo86, line 79: IMO the advantages here are:

```suggestion
- **Simplified Architecture**: By not incorporating Flux, you can simplify the architecture of your Operator. It reduces the number of components and dependencies involved, streamlining the overall system design. 

> By adopting Flux, there is uncertainty about whether it completely eliminates the need to use Helm as a dependency. This uncertainty arises from the possibility of ending up with a design solution that requires both Flux and Helm to be used together. In such a scenario, the advantages of using Flux as a standalone solution may be limited if Helm is still required for certain functionalities. Indeed, if the design solution requires both Flux and Helm as dependencies, additional complexities, including race conditions and concurrency issues, may arise. Managing the interactions and synchronization between Flux and Helm components can introduce challenges related to concurrent modifications, conflicting actions, and maintaining the desired state of the deployment.

- **Reduced Learning Curve: Utilizing Helm directly as a dependency means we can leverage existing knowledge and experience with Helm and we do not need to either Flux experts. This reduces the learning curve of our team, as they can continue using familiar Helm concepts and workflows.
- **Fine-Grained Control:** Managing Helm directly allows you to have fine-grained control over the deployment process. We can customize and adjust Helm commands, flags, and options according to your specific requirements without the abstraction layer introduced by Flux.
- **Flexibility in Tooling:** Without Flux, we have the flexibility to choose and integrate other tools and solutions into your deployment pipeline as needed. We are not limited to the capabilities and features provided solely by Flux.
- **Direct Helm Features:** Helm offers a rich set of features, such as release management, versioning, and dependency resolution. Utilizing Helm directly allows us to leverage these features to their fullest extent, tailoring them to our specific use cases.
- **Compatibility:** Helm provides support to be consumed as lib and has a large user community that does. Using Helm directly ensures compatibility with the broader Helm ecosystem and the ability to easily integrate HelmCharts from various sources.
- **Community Support:** Helm has an active and supportive community, providing resources, documentation, and community-driven enhancements. We can rely on the Helm community for assistance, knowledge-sharing, and issue resolution.

```

docs/arch/adr-011-kurl-addon-operator.md
```diff
@@ -0,0 +1,96 @@
...
+
+How the add-ons will be managed is still an open question. We have a couple of options:
+
+_Use the Helm Controller provided by Flux_
+
+The [Flux Helm Controller](https://fluxcd.io/flux/components/helm/helmreleases/) affords us the following benefits:
+- Stable [HelmRelease API](https://fluxcd.io/flux/components/helm/api/)
+- Helm release [dependencies](https://fluxcd.io/flux/components/helm/helmreleases/#helmrelease-dependencies)
+- Support for Helm install, upgrade, test, rollback and uninstall
+- Performs Helm uninstall on `HelmRelease` CR removal
```
camilamacedo86, line 73: **From my perspective, the fact that Flux uninstalls the HelmChart when a flux HelmRelease is deleted is expected but not an advantage in our case and I think it should be considered a drawback.** This scenario exemplifies the limitations and complexities that may arise when adopting Flux, potentially restricting our control over our solutions according to our specific needs. It restricts our ability to control the deletion process and ensure that essential operations are completed before uninstalling the solution.

IMO: The fundamental purpose of our Operator is to establish a control-loop that triggers whenever a change occurs. Its primary objective is to ensure, through a consistent and idempotent set of operations, that the desired state matches the current state. This control-loop serves as a mechanism for continuously monitoring and maintaining the system in line with our defined configurations and expectations. 

For instance, if a user mistakenly attempts to delete an AddOn CR, we invoke a  [finalizer](https://kubernetes.io/docs/concepts/overview/working-with-objects/finalizers/) function that ensures the AddOn CR is only deleted once all necessary operations, such as uninstalling the HelmChart, have been successfully completed. You can refer to this example in our prototype [here](https://github.com/replicatedhq/brainstorming-ekco/pull/2/files#diff-4d05894b38940f59fcac351cf0cfb3f1ae42eddc423299655677f122dbe4bfe2R299-R343).
4d05894b38940f59fcac351cf0cfb3f1ae42eddc423299655677f122dbe4bfe2R299-R343) an example our prototype.  

However, when using Flux, if a user accidentally deletes the HelmRelease CR that Flux uses for managing Helm releases, it will uninstall the HelmChart solution through the FluxHelmController. In such a scenario, our control-loop can only recreate the Flux HelmRelease but cannot prevent this situation or guarantee that required backups or migrations, for example, have occurred, as we could if we relied solely on Helm. Unless would we be able to say to Flux that it can only uninstall the HelmRelease when a flag that it can be done by adding some info into its HelmRelease CR to prevent this scenario. Is that possible with Flux?

docs/arch/adr-011-kurl-addon-operator.md
```diff
@@ -0,0 +1,96 @@
...
+- update the add-on CR status
+
+### Managing the Add-ons
+
+How the add-ons will be managed is still an open question. We have a couple of options:
+
+_Use the Helm Controller provided by Flux_
+
+The [Flux Helm Controller](https://fluxcd.io/flux/components/helm/helmreleases/) affords us the following benefits:
+- Stable [HelmRelease API](https://fluxcd.io/flux/components/helm/api/)
```
camilamacedo86, line 70: What do you mean with stable API? 
It is an v2beta1 so it does not seems stable, right? Therefore, could have breaking changes.
That is a great point for we check, How often does Flux introduce breaking changes in their releases?

docs/arch/adr-011-kurl-addon-operator.md
```diff
@@ -0,0 +1,96 @@
...
+- Helm release [dependencies](https://fluxcd.io/flux/components/helm/helmreleases/#helmrelease-dependencies)
```
camilamacedo86, line 71: It is indeed a cool feature, but it appears to be more suitable for working within tooling environments rather than building an Operator and our use case. I would think that would be very nice if we have only a tooling.  

Consider the trade-off between to create a HelmRelease CR with dependencies versus directly invoking Helm to install the dependencies before deploying the HelmChart.

By using pure Helm we have greater control over the order and state of the dependencies before adding our HelmChart. However, with a HelmController managing the dependencies, we may face uncertainties about its reliability. In the event of a problem with a dependency, it raises concerns about how we can work around such issues. If we attempt to interact with Helm directly to address potential pitfalls, it can introduce complexities such as race conditions that are challenging to handle effectively in any project.

In summary, while the HelmController feature offers convenience within tooling environments, it may not provide the same level of reliability and control as using Helm directly when we are building an Operator. We need to carefully consider the implications, potential workarounds, and complexities that may arise when managing dependencies through a HelmController in our specific project context.

docs/arch/adr-011-kurl-addon-operator.md
```diff
@@ -0,0 +1,96 @@
...
+- Support for Helm install, upgrade, test, rollback and uninstall
```
camilamacedo86, line 72: In our prototype implementation [here](https://github.com/replicatedhq/brainstorming-ekco/pull/2/files#diff-8e7acbce7dcceea6f611e0a7e681d52cfc12fa9f4678fee8cebfa1e3bdbe5ec2), we have demonstrated the ability to interact with Helm for installing, uninstalling, upgrading, retrieving release information, and converting the release status for a HelmChart. This functionality is achieved with pure Helm, utilizing a single file with just a few lines of code. Since the logic is still handled by the Helm dependency, the implementation remains concise.

Considering this, I initially anticipated that Flux would provide similar capabilities for performing these operations. However, I don't necessarily view such features in Flux as a pro for us. I think we need to looking for what Flux provide and offer for us that would not be easily achieved by only calling Helm. 

However, if we would not go in the direction to build an Operator and we would only tooling the solution then I think Flux would be a game change because would provide all those things. But that also would mean we would not able to do any extra checks or operations before and after install / uninstall / update a Helm solution.

docs/arch/adr-011-kurl-addon-operator.md
```diff
@@ -0,0 +1,96 @@
...
+- Performs Helm uninstall on `HelmRelease` CR removal
+- Chart values override\
+- Pull charts from Helm repository directly
+
+_Develop our own Homebrew Helm Controller_
+
+This approach affords us with a controller that meets our specific needs and avoids any third party software dependency. 
+
```
camilamacedo86, line 80: By looking at this ADR it seems for me that Flux might fit a lot for a tooling purpose, so might be valid we either consider another option: using Flux in combination with the CLI prototype and have an Operator that specifically manages migrations. Following the main pros vs cons that I saw:

pros:
- harnessing the benefits of Flux and its capabilities
- as well as having a dedicated Operator to handle seamless migration management.

cons:
- we will probably NOT able to do operations prior, during or after installing the solutions. This could result in scenarios where certain tasks cannot be fully automated and would require manual intervention, potentially through support. In such cases, we may find ourselves resorting to using the CLI to address these limitations.
-

docs/arch/adr-011-kurl-addon-operator.md
camilamacedo86, line 74: See that with pure Helm we just need to pass the chart values to install/upgrade and we achieve the same, see [here](https://github.com/replicatedhq/brainstorming-ekco/pull/2/files#diff-4d05894b38940f59fcac351cf0cfb3f1ae42eddc423299655677f122dbe4bfe2R169) in the prototype.

So, we could either receive the input from the Installer CR and just pass as map for it, i.e.:

```go
	// Override chart values
	chart.Values = valueOverrides

	// Perform the install or upgrade operation
	if releaseExists(releaseName, actionConfig) {
		_, err = upgradeAction.Run(releaseName, chart)
	} else {
		_, err = installAction.Run(chart, nil)
	}
```

So, I either do not see it as a advantage. Am I overlooking any complexities that Flux handles in these scenarios that I haven't considered?

docs/arch/adr-011-kurl-addon-operator.md
camilamacedo86, line 75: It seems nice but shows either very easy to do with pure Helm, following the Golang code to do it:

```go
import (
	"fmt"

	"helm.sh/helm/v3/pkg/getter"
	"helm.sh/helm/v3/pkg/repo"
)

func pullChart(repository, chartName, chartVersion, destination string) error {
	// Create a new repository object
	r := &repo.ChartRepository{
		Name: repository,
		URL:  repository,
	}

	// Set up the repository
	settings := cli.New()
	repoFile := settings.RepositoryConfig
	chartPath := repoFile.Cache

	// Load the repository index
	index, err := r.DownloadIndexFile(chartPath)
	if err != nil {
		return fmt.Errorf("failed to download repository index: %w", err)
	}

	// Find the chart in the repository index
	chartVersion, err := index.Get(chartName, chartVersion)
	if err != nil {
		return fmt.Errorf("failed to find chart %s: %w", chartName, err)
	}

	// Initialize the downloader
	downloader := &downloader.ChartDownloader{
		Out:              os.Stdout,
		RepositoryConfig: repoFile,
		RepositoryCache:  chartPath,
		Getters:          getter.All(settings),
	}

	// Download the chart
	filename, err := downloader.DownloadTo(chartVersion, destination, false)
	if err != nil {
		return fmt.Errorf("failed to download chart: %w", err)
	}

	fmt.Printf("Chart downloaded to: %s\n", filename)

	return nil
}
```
Am I overlooking any complexities that Flux handles in these scenarios that I haven't considered?

docs/arch/adr-011-kurl-addon-operator.md
```diff
@@ -0,0 +1,96 @@
...
+  name: installer-sample
+spec:
+  channel: "1.27"
+  minio:
+    enabled: true
+    configOverride:
+        persistence:
+            enabled: true
+        replicas: 1
+        mode: standalone
```
camilamacedo86, line 49: I like this one. We can only:

```
var data map[string]interface{}

	err := yaml.Unmarshal([]byte(yamlData), &data)
	if err != nil {
		fmt.Printf("Failed to parse YAML: %v", err)
		return
	}
```

And then we will have the `map[string]interface{}` variable to create, upgrade the HelmChart.

docs/arch/adr-011-kurl-addon-operator.md
camilamacedo86, line 29: I think we need either call down the complexities that we will have within just for we are all aware of:
- a) Users might add invalid data we will need to validate those inputs
- b) Upper version of HelmCharts might need new values or no longer work with values informed previously which means that is a scenario that we need to handle into upgrades
- c) Some AddOns configuration might not work well with other AddOns configuration, that is something that is very complex to prevent or test out. We either will not able to test all possible values combinations.

docs/arch/adr-011-kurl-addon-operator.md
```diff
@@ -0,0 +1,96 @@
...
+
+## Open Questions
+
+- How to allow full configuration of add-ons for advanced vendors/users?
+- Do we use an existing feature rich Helm controller like the one provided by [Flux](https://fluxcd.io/flux/components/helm/) or build our own?
+- How to structure the add-on to adhere to SOLID principles?
+
+## Consequences
+
+- A kURL CLI will be needed to install the operator dependencies (if any)
```
camilamacedo86, line 96: To install the Operator we only need the k8s resources (which can be shipped in a single file and used via raw) and the Operator image. 

Therefore, besides be very cool and not so hard we have a CLI, we have a amazing prototype for that already is not be necessary to have a CLI to provide the Operator, see some options:

**We can do such as Prometheus Operator does**
- Provide all in a bundle/file: https://raw.githubusercontent.com/prometheus-operator/prometheus-operator/release-0.51/bundle.yaml
- Then, to instal it we just need kubectl apply -f https://raw.githubusercontent.com/prometheus-operator/prometheus-operator/release-0.51/bundle.yaml and then apply the kURL/Installer CR

Also, they allow download from the release page, see: https://github.com/prometheus-operator/prometheus-operator/releases/download/v0.65.1/bundle.yaml

- We can either have a small shell script to just ship the Operator
- We can either provide a tar with the instructions to untar and install

docs/arch/adr-011-kurl-addon-operator.md
```diff
@@ -0,0 +1,96 @@
...
+- Performs Helm uninstall on `HelmRelease` CR removal
+- Chart values override\
+- Pull charts from Helm repository directly
+
+_Develop our own Homebrew Helm Controller_
+
+This approach affords us with a controller that meets our specific needs and avoids any third party software dependency. 
+
+
+
```
camilamacedo86, line 82: Could we add the image here? WDYT?

Screenshot 2023-04-28 at 10 05 22

```suggestion
<img width="776" alt="Screenshot 2023-04-28 at 10 05 22" src="https://user-images.githubusercontent.com/7708031/235105356-2d499fcd-8089-44f8-9a90-77547d6cb5cc.png">
```

docs/arch/adr-011-kurl-addon-operator.md
```diff
@@ -0,0 +1,96 @@
...
+- ensure add-on is in desired state
+- update the add-on CR status
+
+### Managing the Add-ons
+
+How the add-ons will be managed is still an open question. We have a couple of options:
+
+_Use the Helm Controller provided by Flux_
+
+The [Flux Helm Controller](https://fluxcd.io/flux/components/helm/helmreleases/) affords us the following benefits:
```
camilamacedo86, line 69: One advantage I see with Flux is that it provides a code implementation for reconciling HelmCharts and includes watches and finalizers, as demonstrated in their [helmrelease_controller.go](https://github.com/fluxcd/helm-controller/blob/main/internal/controllers/helmrelease_controller.go) file on GitHub. 

However, it comes with the limitation that we cannot modify it according to our specific needs. The controller is generic and may not work effectively with all HelmCharts. By implementing our own solution per Addon, we can automate solutions based on the specific requirements of each Addon. This allows us to address specific needs and requirements more accurately, as a generic implementation will not perform in the same way. It's important to note that a generic automation for a HelmChart will never be able to handle specific issues, workarounds, or tailor solutions that are commonly done in KURL for what we ship.

docs/arch/adr-011-kurl-addon-operator.md
camilamacedo86, line 80: Could we add a sub item here with: https://github.com/replicatedhq/kURL/pull/4488/files#r1192957945 (i.e. Long Term considerations )

docs/arch/adr-011-kurl-addon-operator.md
```diff
@@ -0,0 +1,96 @@
...
+
+```
+
+### The Controllers
+
+The operator will have at minimum one controller which ensures the `Installer` CR is reconciled with the cluster periodically. The `Installer` controller (main controller) will delegate responsiblity for managing add-ons to other controllers. The reconcilation for the main controller will consist of determining the add-ons to be managed from the `Installer` CR and creating a controller for each add-on (addon-on controller).
+
+The add-on controller will be responsible for
+- verifying the requirements for the add-on are met
+- managing the add-on itself
```
camilamacedo86, line 59:
```suggestion
- managing the add-on itself and automate the specific needs for each AddOn solution
```

docs/arch/adr-011-kurl-addon-operator.md
camilamacedo86, line 77:
```suggestion
_Using pure Helm without add any third-party as dependency_
```

docs/arch/adr-011-kurl-addon-operator.md
rrpolanco: I will remove this. I got it mixed up with the [`GitRepository`](https://fluxcd.io/flux/components/source/gitrepositories/) API. But the point about relying on a beta API still stands.

docs/arch/adr-011-kurl-addon-operator.md
ricardomaraschini: I agree. Let me add one thing: we are doing a lot of effort to ensure we can guarantee seamless cluster upgrades and IMHO to expose `values.yaml` would render the same situation we have nowadays. We won't be able to test all possible combinations during upgrades.

docs/arch/adr-011-kurl-addon-operator.md
emosbaugh: the controller does have a dependsOn property

https://fluxcd.io/flux/components/helm/helmreleases/#helmrelease-dependencies

i can see us taking advantage of this in conjunction with helm tests for creating pre-requirement checks

docs/arch/adr-011-kurl-addon-operator.md
emosbaugh: anything valuable we find in the helmrelease_controller.go i assume we would have to build ourselves

https://github.com/fluxcd/helm-controller/blob/main/internal/controllers/helmrelease_controller.go

docs/arch/adr-011-kurl-addon-operator.md
emosbaugh: Flux achieves this though composeValues

https://github.com/fluxcd/helm-controller/blob/main/internal/controllers/helmrelease_controller.go#L555-L558

docs/arch/adr-011-kurl-addon-operator.md
emosbaugh: It looks to me like helm does not update crds

https://helm.sh/docs/chart_best_practices/custom_resource_definitions/#some-caveats-and-explanations

and thus flux has added logic and policies around crds

https://github.com/fluxcd/helm-controller/blob/main/internal/runner/runner.go#L229C31-L230

I cant say what other complexities we will find when trying to implement this ourselves and thus it may be best to offload this to flux

emosbaugh: Additionally flux has built in test support for validating releases. This could be beneficial

https://github.com/fluxcd/helm-controller/blob/main/internal/controllers/helmrelease_controller.go#L436

camilamacedo86: @emosbaugh, 

Flux is who manages and that is the whole issue. 
It means we are NOT allow to prevent the solution be uninstalled since we are NOT responsible for manage it. 
Therefore, see that HelmController is generic and flux never will allow us to do our specific checks prior allow or not solution get uninstalled. Example: is all PVC migrates? Have you other object store installed and etc.

docs/arch/adr-011-kurl-addon-operator.md
camilamacedo86: We'll need a controller in any scenario to manage statuses, finalizers, and so on. Flux, while a robust solution, is generic and therefore doesn't address our specific needs and requirements.

Our objective is highly targeted, and it's important to remember that we can't attempt to update or delete any resource created and managed by Flux. Doing so would trigger the control-loop which could lead to unintended consequences. In many scenarios, this could result in Flux re-executing all operations to ensure its processes are maintained, as that's what a control-loop is designed to do. Consequently, the issue we tried to resolve might resurface. Alternatively, we could encounter race or concurrency conditions, as both we and Flux would be trying to update/change the same resources simultaneously.

Flux isn't tailored to cater to specific needs, like what we're seeking here, nor is it designed for the kind of integration and usage we're proposing. Its main function is to provide a tool capable of automating the process of installing, updating, or uninstalling a HelmChart, without considering any additional requirements. For example, it wouldn't accommodate a scenario where Kots is configured with N, implying that X must be present. In situations where you try to uninstall X without Y, it won't work. Flux doesn't account for these kinds of complexities.

docs/arch/adr-011-kurl-addon-operator.md
ricardomaraschini: I don't think that the ADR mentions backward compatibility. The phrasing seems correct: as long as the kURL installation has enabled only Kubernetes, Containerd and Flannel it should work. On that regard a kURL with Kubernetes, Containerd and Flannel is not different from k0s, kind or minikube.

docs/arch/adr-011-kurl-addon-operator.md
emosbaugh: In these comments I am only listing the benefits of using flux. I understand you point about uninstalling.

docs/arch/adr-011-kurl-addon-operator.md
emosbaugh: If we use flux, I suggest we never modify the Kubernetes resources themselves and instead modify values through the HelmChart CR. If we ever had to we can use the suspend property of the [CRD](https://fluxcd.io/flux/components/helm/helmreleases/#specification).

Flux is a reconciliation loop. If one chart has a precondition it will fail until that precondition is met and then succeed.

I feel the beauty in the flux helm controller is it is built to do one thing only and that is manage the lifecycle of helm charts.

docs/arch/adr-011-kurl-addon-operator.md
laverya: we'll need to merge this config with whatever our default config is, too

docs/arch/adr-011-kurl-addon-operator.md
chris-sanders: > It's important to note that a generic automation for a HelmChart will never be able to handle specific issues, workarounds, or tailor solutions that are commonly done in KURL for what we ship

Can you provide an example of what you mean by this? Right now it's just a high level specific > generic and I'm not clear that's true in this case.

docs/arch/adr-011-kurl-addon-operator.md
chris-sanders: This is true, using another tool does mean loss of some control. It also means less effort to implement and maintain going forward. Our decision needs to be based on both and just stating that it's more precise without addressing the implementation and maintenance costs is not addressing the full story. Our dev time is limited, loosing precise control in favor of not having to implement and maintain is one of our goals moving to a new implementation.

My concern here is, why would we want to manage dependencies explicitly that doesn't sound like an advantage to me unless you have examples of bad decisions flux has made doing this that we need to implement in another way.

docs/arch/adr-011-kurl-addon-operator.md
chris-sanders: > anything valuable we find in the helmrelease_controller.go i assume we would have to build ourselves
>
>https://github.com/fluxcd/helm-controller/blob/main/internal/controllers/helmrelease_controller.go

I think this is a really insightful comment. Has anyone reviewed the controller? I'm curious how much of this controller do we think we need to implement for a robust solution?

docs/arch/adr-011-kurl-addon-operator.md
chris-sanders: Can you provide an example when it's beneficial to not delete the CR? I can't think of a situation and it would be helpful to understand why you see it beneficial to write custom deletion code ourself.

docs/arch/adr-011-kurl-addon-operator.md
rrpolanco: I got this objective from @emosbaugh add-on operator proposal [doc](https://docs.google.com/document/d/1ZnwgJxRZwxQKwEWCl9UQx7NRILk4hd28dwItPBuZjLU/edit#heading=h.isylz7bayrpk). I _think_ the point here is that it should work on a kURL Kubernetes minimal distro.

docs/arch/adr-011-kurl-addon-operator.md
chris-sanders: This is a good example where the prototype implementation is only a few lines but flux has a much more comprehensive implementation. What's the difference here, is the flux method better and addressing thing we haven't yet addressed or is it more complicated but functionally the same for our use case?

A second question, it looks like in the above prototype code, we replace the chart values with valuesOverrides. Is that true? The intention would be to mergeOverride, IE only override values that are set not replace all values with a new set of values.

docs/arch/adr-011-kurl-addon-operator.md
chris-sanders: > Therefore, see that HelmController is generic and flux never will allow us to do our specific checks prior allow or not solution get uninstalled. Example: is all PVC migrates? Have you other object store installed and etc.

I don't understand why we can't perform checks like PVC migrations and object store migrations before uninstalling via Flux. Flux won't remove the object store until we tell it too, why can't we do our migrations, verifications, and then tell flux to remove it when we're ready?

docs/arch/adr-011-kurl-addon-operator.md
rrpolanco: I can foresee the pitfalls with allowing users to pass/override the values for an add-on. However, there is a need to allow end-users to override _certain_ add-on configs for e.g. Rook (filters for OSD devices) and containerd (config for GPU support) to name a few. We can either i) restrict what configs can be overidden or ii) allow unmitigated access to override `values` but with the caveat that users "proceed at their own risk".

docs/arch/adr-011-kurl-addon-operator.md
rrpolanco: We need to reconcile the recently proposed [versioning](https://github.com/replicatedhq/kURL/pull/4430) scheme with this API.

docs/arch/adr-011-kurl-addon-operator.md
camilamacedo86: @chris-sanders,

> 
Can you provide an example of what you mean by this? Right now it's just a high level specific > generic and I'm not clear that's true in this case.

Certainly, I can provide an example. In the following KURL code, we delete resources as a workaround for certain issues: [Link to KURL code](https://github.com/replicatedhq/kURL/blob/b3e3c02ffb04334103c8ad0f80ba6938e2036694/addons/rook/1.9.12/install.sh#L75C4-L81)

However, with Flux, situations like these could trigger the Flux control loop to run again, potentially overwriting our changes since it's designed to ensure the HelmChart is properly installed. Another risk is that if both we and Flux try to update and monitor the same resources, we could encounter race conditions.

I hope that clarifies my concerns.

docs/arch/adr-011-kurl-addon-operator.md
camilamacedo86: @chris-sanders @emosbaugh @ricardomaraschini @rrpolanco @laverya:

The primary issue with this approach, as per SOLID principles, particularly the Single Responsibility Principle (SRP) and the Liskov Substitution Principle (LSP), is that each class (or in this case, operator) should have one, and only one, reason to change, and that subclasses (or dependent services) must be substitutable for their base classes without altering the correctness of the program.

If we were to start manipulating Persistent Volume Claims (PVCs), or any resources managed by Flux as part of its reconciliation loop, it essentially means that our custom operator would be taking on more than one responsibility. This would violate the SRP, as our operator should only be concerned with its unique tasks, while Flux is tasked with managing the Helm charts.

In addition, by performing operations on resources managed by Flux before Flux has completed its own tasks, we're violating the LSP. Flux and our operator can be seen as base and derived classes, respectively. By interfering with Flux's responsibilities, we risk introducing errors into the system, because our operator would no longer be a drop-in replacement for Flux.

In a real-world scenario, imagine a race where both operators want to update or modify the PVCs. If we migrate and validate the PVCs independently and then allow Flux to remove them, there's a chance Flux could overwrite our changes in its reconciliation loop. This can lead to inconsistencies in the state of resources, potentially causing conflicts, data loss, or system instability. In essence, any changes we make could be nullified by Flux, wasting resources and leading to an endless cycle of modifications. Therefore, it is critical to respect each operator's boundaries and allow them to manage their designated resources independently.

docs/arch/adr-011-kurl-addon-operator.md
camilamacedo86: @rrpolanco, both @ricardomaraschini and I seems agree with your perspective `We can either i) restrict what configs can be overidden or ii) allow unmitigated access to override values but with the caveat that users "proceed at their own risk".`. 

Considering the inherent risks and the specific business requirements that kURL aims to address - such as facilitating seamless upgrades, automating workarounds, and minimizing complexities for users - it's crucial to underscore the potential challenges that arise when users are permitted to add any chart value.

This flexibility comes with its own set of complexities: verifying the validity of these values, ensuring compatibility between various solutions, and handling possible conflicts. For instance, accommodating both OpenEBS and Rook in the same cluster with any given configuration could pose challenges and that we cannot test all possible combinations. Furthermore, we need to address what happens during upgrades when a particular value is no longer valid.

Balancing user freedom and system stability is a delicate act, and it's paramount that we consider these factors to avoid complicating the user experience or introducing instability to the system.

Then, I think we can propose this solution and call out all complexities and problems inherent to this approach as add an alternative option where we limit the possible options within the pros and cons.

docs/arch/adr-011-kurl-addon-operator.md
camilamacedo86: @chris-sanders 

The question isn't about choosing not to delete the Custom Resource (CR), but about ensuring that all necessary precautions for a safe uninstallation are undertaken before deleting a CR, which represents an Add-On and its corresponding solution in the cluster.

For example, consider a scenario where the OpenEBS Kurl/Installer CR is deleted manually or via any other solution, whether by mistake or intentionally. Before proceeding with the Helm uninstallation of OpenEBS, we could guarantee that an alternative storage solution is already in place and that the volume data has been successfully migrated to this new solution. Otherwise, the OpenEBS CR will not get deleted unless you explicitly force it manually. Furthermore, we could for example ensure that backup solution should be employed to store a copy of the data. This will allow data recovery via support, if necessary.

In essence, the focus is not on whether or not to delete the CR but rather on ensuring that all the required safety checks and procedures are completed prior to its deletion. 

Same is valid if someone try to delete the namespace where the AddON CR leaves. By using finalizers the namespace cannot get deleted prior all resources on it be first get deleted. To better understand you can check the how finalizers work in the k8s docs: https://kubernetes.io/docs/concepts/overview/working-with-objects/finalizers/ and see how the k8s api and us building an Operator would leverage in the cascading delation: https://kubernetes.io/docs/tasks/administer-cluster/use-cascading-deletion/

docs/arch/adr-011-kurl-addon-operator.md
camilamacedo86: @emosbaugh,

> If we use flux, I suggest we never modify the Kubernetes resources themselves and instead modify values through the HelmChart CR. If we ever had to we can use the suspend property of the [CRD](https://fluxcd.io/flux/components/helm/helmreleases/#specification)

If we choose to adopt Flux, it would imply that we could no longer directly install, delete, or update HelmCharts or Custom Resource Definitions (CRDs) in our controllers. Additionally, we would not be able to manage any resources created and managed by Flux during the installation of these HelmCharts.

However, it's important to note that our business requirements necessitate certain actions. These include ensuring seamless migrations, implementing workarounds for specific issues, and automating steps that are presently manual and only applicable within our particular operational context to guarantee smooth upgrades. Consequently, it's crucial that we retain control over these processes. While Flux is a powerful tool, it doesn't cater to the kind of specific use-cases that we often need to address. Therefore, we need a solution that accommodates our unique needs while providing comprehensive control over our resources. 

If we try to manipulated the same resources then you are proposing a solution that violates the SOLID principle and suggests the use of two overlapping Operators, it would introduce potential challenges related to race conditions and concurrency. It's important to note that having two instances of the same Operator is not feasible due to these reasons, and the idea of having two separate Operator solutions responsible for the same tasks clearly goes against the Single Responsibility Principle, as per SOLID principles

camilamacedo86: Hi @chris-sanders, 

I've checked a little the[ helmrelease_controller.go](https://github.com/fluxcd/helm-controller/blob/main/internal/controllers/helmrelease_controller.go) file from FluxCD's Helm Controller. Then, for me although it contains a wide range of features, not all shows that are relevant to our specific needs, and crucially, it doesn't address the unique requirements that we have. As a result, our implementation to meet our business requirements—striving for full automation to minimize human intervention, and utilizing a reduced, opinionated specification—**would IHMO not mirror Flux's approach.**

If the concern is about the potential effort involved in developing a similar solution, could we not consider Flux's implementation as a reference point? If our paths do align—**which I currently do not foresee see too much beyond mainly installing/uninstalling/deleting HelmCharts—Flux's** code could serve as a useful guide to expedite our progress. WDTY?

Following a couple of points. 

## Flux reconciliation logic is not the same approach proposed in the prototype

The approach of our prototype using Helm for reconciliation significantly differs from Flux's diff-based approach (see [here](https://github.com/fluxcd/helm-controller/blob/main/internal/controllers/helmrelease_controller.go#L162-L170)). Instead of relying on diffing the current cluster state against the desired state, our prototype takes a different approach as proposed [here](https://github.com/replicatedhq/brainstorming-ekco/pull/2/files#diff-98c935b2ef1adddcc8355cdf8f61bf1adb3c14ab90de326a27874d5bdbd7f6dbR130-R256).

In our prototype, we proposed directly handling reconciliation based on the desired state specified through user input into the Installer/Kurl CR, our defaults, and specific HelmChart versions. Unlike Flux's Diff() function, which shows to calculates the changes needed to reconcile the current state of Kubernetes resources with the state defined in the Helm release, our custom Operator avoids this diffing process. 

The diff approach shows that can bring for example:

- False Positives: Diffing relies on comparing the current cluster state with the desired state defined in configuration files. However, this comparison may result in false positives, indicating differences that are not actual drift. This can lead to unnecessary and potentially disruptive corrective actions.
- Continuous Drift: Depending solely on diff-based drift detection and correction may result in a continuous cycle of corrections, causing unnecessary system churn and instability. 
- Overwriting Manual Interventions and Customizations: The diff approach shows that cannot consider manual changes or configurations made to the cluster, leading to unintended consequences when the system attempts to automatically correct these changes. This can result in disruptions or loss of customized configurations that could not be addressed via the chart values for example.  

### Managing Dependencies and Pre-requirements/Conditions

In our custom solution using pure Helm, we prioritize validating that all pre-requisites for installing, uninstalling, or deleting an AddOn are satisfied for each specific scenario. We don't need to rely on a generic mode, which might not comprehensively cover our needs. For instance, while handling dependencies, Flux merely checks if a HelmRelease is installed and declares success (see [here](https://github.com/fluxcd/helm-controller/blob/main/internal/controllers/helmrelease_controller.go#L258-L272)).

When we were doing our prototype and demo, we checked that there are circumstances where shows that a HelmRelease might report that a solution is deployed when, in fact, it isn't fully functional or installed properly. In our controller, we wouldn't manage things this way. Instead, we can prioritize validating if ANY specific requirements in line with our business rules and customer demands - are not just deployed, but also HEALTHY and aligned with our particular opniated spec configurations, before performing any next step operation. This degree of detailed health-check is beyond Flux's capabilities for any pre-condition addon or dependency. As a result, our custom solution could provides superior assurance and control and grow according with the problems faced by the users.Therefore the custom approach has the potential to:

- mitigate scenarios where an attempted solution installation results in an unhealthy cluster state, often necessitating manual intervention—a common issue with the current KURL and a pain point for our users. 
- moreover, we would be able to effectively monitor and report status conditions, indicating when pre-requisite conditions are not met, as proposed [here](https://github.com/replicatedhq/brainstorming-ekco/issues/6). Thus, our approach using pure Helm enhances stability, user experience, and transparency of system status which is not achievable by adopting Flux. 
- prevent scenarios where an attempted solution installation could lead to an unhealthy cluster state --common user feedback and request. This minimizes the need for manual intervention, a significant issue with the current KURL and a recurrent user concern.

## Tracking the current state of the solutions

As previously emphasized, Flux is a generic solution and, consequently, lacks the capacity to comprehensively verify whether each specific Addon installation or upgrade has been successfully completed and is functioning healthily, particularly considering the unique intricacies of each specific solution.

In contrast, by retaining full control over our operations, we could perform granular checks on pre-requisites for each specific Addon, ensuring their individual compatibility and readiness within our specific system. This ties into the approach proposed [here](https://github.com/replicatedhq/brainstorming-ekco/issues/6). That can enables us to gain granular visibility into system health, pinpoint any issues, identify their causes, and rectify them effectively. 

Moreover, as demonstrated in our [prototype](https://github.com/replicatedhq/brainstorming-ekco/pull/2/files#diff-8e7acbce7dcceea6f611e0a7e681d52cfc12fa9f4678fee8cebfa1e3bdbe5ec2R132-R159), using pure Helm enables us to retrieve the status from Helm and convert it into our custom Status conditional structure with just a few lines of code.

### Remove the complexities to deal and parse Flux logs

In contrast, Flux would require us as discussed yesterday and shared for those that have been working with in many scenarios parse LOGS and not only monitor its CRS statuses, introducing a new layer of complexity. In POV, it would necessitate significantly more code implementation and maintenance. 

## Let's now speak about FLUX finalizers implementation

The Flux Controller handles the finalizer, a method that prevents deletion of a Custom Resource (CR) representing the solution on the cluster or the corresponding namespace where the resource is installed. It operates by simply verifying whether Helm has reported that the Chart was uninstalled, as evidenced [here](https://github.com/fluxcd/helm-controller/blob/74bc32e840e1b09e48036cb5553c62a71a49baba/internal/controllers/helmrelease_controller.go#L669-L725).

On the other hand, when using our custom Operator with pure Helm, we have the flexibility to perform additional operations prior to deletion. This allows us to prevent unintended consequences such as data loss, which is discussed in greater detail [here](https://github.com/replicatedhq/kURL/pull/4488#discussion_r1197553410). The use of Helm gives us more granular control over the lifecycle of our resources, a level of detail not provided by Flux's approach.

docs/arch/adr-011-kurl-addon-operator.md
rrpolanco: kURL cannot gatekeep add-on configuration or else we will fall into the same scenario with the current kURL offering which is that we have to release a new feature every time a vendor/end-user wants to configure something outside of the normal use case. E.g. configuring GPU support for containerd.
I think a reasonable approach is to document our opinionated (and supported) configuration (`values.yaml`) for each add-on and allow vendors/users to overwrite them with the warning that their applied config may not be supported (i.e. they can proceed but at their own risk).

docs/arch/adr-011-kurl-addon-operator.md
camilamacedo86: @rrpolanco I want to emphasize my opinion that adopting the Flux approach to address our requirements, such as smooth upgrades, migrations, and specific automations, would lead to a situation where we would need to manipulate the responsibilities of the Flux Operator. This would result in having two operators overlapping their responsibilities, which goes against the SOLID principles.

I have made efforts to highlight this point in various places throughout this ADR, as well as in our communication channels and other documentation.

c/c @ricardomaraschini

docs/arch/adr-011-kurl-addon-operator.md
camilamacedo86: I am totally OK with that.
I just think that in the ADR we should explicitly describe the limitations and potential problems. Just that. we need to keep it write down.

PR #4491, opened 2023-05-15
Adds cluster preflight check to verify if CNI is ready

camilamacedo86:

Ensure that the CNI is ready prior upgrade.
NONE

Note that because of bug https://github.com/replicatedhq/troubleshoot/issues/1089 we are facing duplicating results see: 

Screenshot 2023-05-16 at 18 25 37

However, it might not be a blocker. 

```release-note
Adds preflight check to verify that CNI is ready prior upgrades. If we found an issue then an failure will be raised.
```

### DIFF
--- a/pkg/preflight/assets/preflights.yaml
+++ b/pkg/preflight/assets/preflights.yaml
@@ -6,12 +6,12 @@ metadata:
 spec:
   collectors:
     - clusterResources: {}
+  hostCollectors:
     - run:
         collectorName: journalctl-kubelet
         exclude: '{{kurl or (not .IsPrimary) (not .IsUpgrade) }}'
-        image: busybox:1
         command: [ "journalctl" ]
-        args: [ "-u", "kubelet", "--no-pager", "-S", "7 days ago" ]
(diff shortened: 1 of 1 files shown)


pkg/preflight/assets/preflights.yaml
```diff
@@ -6,6 +6,12 @@ metadata:
 spec:
   collectors:
     - clusterResources: {}
+    - run:
```
xavpaice, line 9: the in-cluster 'run' collector is deprecated, replaced by runPod.

I think you're looking for the host collector though - https://troubleshoot.sh/docs/host-collect-analyze/run/

camilamacedo86: Done thank you.

emosbaugh (changes requested):
If the cni were not ready at any time and then it was fixed this will block the upgrade. this is the wrong way to go about this check. you could try to start a pod instead.

camilamacedo86:
>If the cni were not ready at any time and then it was fixed this will block the upgrade. this is the wrong way to go about this check. you could try to start a pod instead.

It is only a warning but I am looking in a more robust way to verify it
So, I will close this one.

PR #4497, opened 2023-05-16
chore(actions): deprecate set-output

emosbaugh:

A workflow using save-state or set-output like the following should be updated to write to the new GITHUB_STATE and GITHUB_OUTPUT environment files.

https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/


ricardomaraschini (approved):

emosbaugh:
[sc-62601]

PR #4501, opened 2023-05-16
adds preflight check to warning unhealthy pods in the upgrades

camilamacedo86:

This PR adds check to allow users know what are the Pods which are unhealthy prior upgrades

Fixes #

The following result was changed to show the name and namespace of the POD that is OK. 

Screenshot 2023-05-16 at 19 09 35

```release-note
Adds preflight check healthy state of all Pods in the namespaces (kube-flannel, default, kube-node-lease, kube-public, kube-system, kurl, projectcontour, rook-ceph, velero, longhorn-system, openebs) prior upgrades. If a Pod not be healthy then a warning will be raised. 
```

### DIFF
--- a/pkg/preflight/assets/preflights.yaml
+++ b/pkg/preflight/assets/preflights.yaml
@@ -22,10 +22,22 @@ spec:
               message: "All nodes are online."
     - clusterPodStatuses:
         checkName: Pods status check
+        namespace:
+          - kube-flannel
+          - default
+          - kube-node-lease
+          - kube-public
+          - kube-system
+          - kurl
+          - projectcontour
+          - rook-ceph
+          - velero
+          - longhorn-system
+          - openebs
         exclude: '{{kurl or (not .IsPrimary) (not .IsUpgrade) }}'
         outcomes:
+          - pass:
+              message: "The Pod {{ .Name }} into the namespace {{ .Namespace }} is OK."
           - warn:
               when: "!= Healthy" # Catch all unhealthy pods. A pod is considered healthy if it has a status of Completed, or Running and all of its containers are ready.
               message: "A Pod, {{ .Name }} into the namespace {{ .Namespace }}, is unhealthy with a status of: {{ .Status.Reason }}. Restarting the pod may fix the issue."
-          - pass:
-              message: "All Pods are OK."
--- a/pkg/preflight/assets/preflights.yaml
+++ b/pkg/preflight/assets/preflights.yaml
@@ -6,6 +6,7 @@ metadata:
 spec:
   collectors:
     - clusterResources: {}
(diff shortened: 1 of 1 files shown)


pkg/preflight/assets/preflights.yaml
```diff
@@ -19,3 +20,12 @@ spec:
...
               message: "All nodes are online."
+    - clusterPodStatuses:
+        checkName: Pods status check
+        exclude: '{{kurl or (not .IsPrimary) (not .IsUpgrade) }}'
+        outcomes:
+          - warn:
+              when: "!= Healthy" # Catch all unhealthy pods. A pod is considered healthy if it has a status of Completed, or Running and all of its containers are ready.
+              message: "A Pod, {{ .Name }} into the namespace {{ .Namespace }}, is unhealthy with a status of: {{ .Status.Reason }}. Restarting the pod may fix the issue."
+          - pass:
+              message: "All Pods are OK."
```
xavpaice, line 31: this analyzer runs once per pod with `{{ .Name }}` - so you might say rather than "All Pods" you could say "Pod {{ .Name }} is OK"

camilamacedo86: Oh cool so then we could move forward within this one. Great @xavpaice 
Thank you.

camilamacedo86, line 31:
```suggestion
              message: "Pod, {{ .Name }}, into the namespace {{ .Namespace }} is OK"
```

pkg/preflight/assets/preflights.yaml
```diff
@@ -19,3 +20,12 @@ spec:
               message: "Not all nodes are online."
           - pass:
               message: "All nodes are online."
+    - clusterPodStatuses:
+        checkName: Pods status check
+        exclude: '{{kurl or (not .IsPrimary) (not .IsUpgrade) }}'
```
camilamacedo86, line 25: @laverya @emosbaugh WDYT about this one?

emosbaugh: Does this apply to all pods in the cluster? If so I am opposed as non-infrastructure pods should no impact the install.

camilamacedo86: I added the namespaces now,
If the namespace does not exist that does not fail

emosbaugh (changes requested):
in my opinion this is overly aggressive. there may be pods that do not have to be healthy and the upgrade should still proceed. this will block an upgrade. i will not accept this pr or any pr like it.

camilamacedo86:
Closing this one as deffered based on the comment: https://github.com/replicatedhq/kURL/pull/4501#pullrequestreview-1455525589

PR #4504, opened 2023-05-17
Automated Rook version update 1.11.6

replicated-ci-kurl:
Automated changes by the [cron-rook-update](https://github.com/replicatedhq/kURL/blob/main/.github/workflows/update-rook.yaml) GitHub action

```release-note
Adds [Rook add-on](https://kurl.sh/docs/add-ons/rook) verison 1.11.6.
```


camilamacedo86:
By analysing the many runs of the tests we can see that this version is passing
The latest ones which fails one Oracle Linux  is because I new release was shipped and our preflight is not yet allowing it.

So, it shows good to fly,

PR #4513, opened 2023-05-18
(It is just to fix/improve the process to generate auto PRs for weave)fix to allow generate diff templates for 2.6.x

camilamacedo86:

Fix AddOn automation after we apply a change valid only for 2.8+


emosbaugh (changes requested):

addons/weave/template/generate.sh
```diff
@@ -2,6 +2,9 @@
 
 set -euo pipefail
 
+# Source the functions file to semver
+source ./../../../scripts/common/common.sh
```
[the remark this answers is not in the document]
camilamacedo86: What do you mean with that?
Why we would not able to use source?
Should better just copy and paste the implementation?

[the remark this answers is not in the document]
emosbaugh: could you just change it to this?

```
if [[ "$ADDON_VERSION" == 2.8.* ]]; then
```

addons/weave/template/generate.sh
```diff
@@ -63,7 +63,12 @@ function generate() {
 
     if [ -n "$WEAVE_EXEC_IMAGE_PATCH_VERSION" ]; then
         sed -i "s/weaveworks\/weaveexec:$VERSION/kurlsh\/weaveexec:${WEAVE_EXEC_IMAGE_PATCH_VERSION}/" "$dir/Manifest"
-        sed -i "s/weaveworks\/weaveexec:$VERSION/kurlsh\/weaveexec:${WEAVE_EXEC_IMAGE_PATCH_VERSION}/" "$dir/patch-daemonset.yaml"
+
+        if [[ "$ADDON_VERSION" == 2.8.* ]]; then
+            sed -i "s/weaveworks\/weaveexec:$VERSION/kurlsh\/weaveexec:${WEAVE_EXEC_IMAGE_PATCH_VERSION}/" "$dir/patch-daemonset.yaml"
+        else
+            sed -i "s/weaveworks\/weaveexec:$VERSION/kurlsh\/weaveexec:${WEAVE_EXEC_IMAGE_PATCH_VERSION}/" "$dir/patch-daemonset_legacy.yaml"
```
emosbaugh, line 70: i think you will have to rename this file to patch-daemonset.yaml so its in the kustomization.yaml

emosbaugh (changes requested):

addons/weave/template/generate.sh
```diff
@@ -64,6 +67,16 @@ function generate() {
...
+
+        # The following code will remove the initContainer added to sort out issue face with weave and symbolic links
+        semverParse "${ADDON_VERSION}"
+        local addon_major_minor_version="${major}.${minor}"
+        if [ "$addon_major_minor_version" != "2.8" ]; then
+            # Remove initContainers section
+            sed -i '/initContainers:/,/volumes:/d' "$dir/patch-daemonset.yaml"
+            # Remove volumes section
+            sed -i '/volumes:/,+1d' "$dir/patch-daemonset.yaml"
+        fi
```
camilamacedo86, line 79: @emosbaugh I think I found a better way

emosbaugh: could you instead have two different files entirely with a patch for each version? this seems error prone

emosbaugh (changes requested):

laverya (approved):

laverya:
After this gets merged, we should run it to see what changes it makes for sure

camilamacedo86:
Hi @laverya,

Thank you a lot. 
I did that now, see: https://github.com/replicatedhq/kURL/actions/runs/5318403531/jobs/9647260756. All fine.

PR #4522, opened 2023-05-20
Check if is possible to connect with proxy

camilamacedo86:

Fixes #

```release-note

```


scripts/common/proxy.sh
```diff
@@ -47,6 +47,44 @@ function proxy_bootstrap() {
...
+        ip=$(cat "$config_file" | grep -oP '(?<=plugins\."io.containerd.grpc.v1.cri"\.registry\.configs\.").*(?="\.tls\])')
+        ca_file=$(cat "$config_file" | grep -oP '(?<=ca_file = ").*(?=")')
+
+        # Execute curl command and capture the output
+        set +e
+        response=$(curl --cacert "$ca_file" "https://$ip/v2/" 2>&1)
+        set -e
+
+       # Check if the curl command failed with "Connection refused"
+       if [[ $response == *"Connection refused"* ]]; then
```
camilamacedo86, line 78:
```suggestion
       if [[ $response != *"UNAUTHORIZED"* ]]; then
```

scripts/common/proxy.sh
```diff
@@ -47,6 +47,44 @@ function proxy_bootstrap() {
...
+           logWarn "Connection with registry failed! Check the installer configuration."
+           logWarn "If you are using a Proxy, ensure that it is a valid Proxy and properly configured."
```
camilamacedo86, line 80:
```suggestion
           logWarn "See the response: $response"
           logWarn "If you are using a Proxy, ensure that it is a valid Proxy and properly configured."
```

scripts/join.sh
```diff
@@ -186,13 +186,13 @@ function main() {
     get_shared
     kubernetes_pre_init
     ${K8S_DISTRO}_addon_for_each addon_join
+    check_proxy_config
```
camilamacedo86, line 189: @emosbaugh @laverya 

Should we call only in the JOIN commands ?
WDYT?

scripts/common/proxy.sh
```diff
@@ -47,6 +47,55 @@ function proxy_bootstrap() {
...
+            printf "${NC}\n"
+            logWarn "If you are using a Proxy, ensure that it is a valid Proxy and properly configured."
+            logWarn "More info: https://kurl.sh/docs/install-with-kurl/proxy-installs"
+        else
+            logSuccess "Registry config with Containerd checked successfully"
+        fi
+    fi
+}
+
 function kubectl_no_proxy() {
```
camilamacedo86, line 99: @laverya @emosbaugh 

In the join command I am facing an issue. If I run the code outside of kURL script it works fine but in the script it takes more time to do the curl and it is not checking the result:

Example calling the func after package_cleanup
Screenshot 2023-05-21 at 09 32 12


Screenshot 2023-05-21 at 09 32 12

camilamacedo86: I think it might be because +e and -e to handle the issue scenario I will try the same in another way.

camilamacedo86: It seems sorted out the issue
Also, I change to use local vars.

scripts/common/proxy.sh
```diff
@@ -47,6 +47,56 @@ function proxy_bootstrap() {
...
+        if [ -f "$proxy_config_file" ]; then
+           echo "Containerd HTTP proxy config:"
+           grep -v -e '^\[Service\]' -e '^# Generated by kURL' "$proxy_config_file"
+           echo ""
+        fi
+
+        # Execute curl command and capture the output
+        echo "curl --cacert $ca_file https://$ip/v2/"
+        local response=""
+        if output=$(curl --cacert "$ca_file" "https://$ip/v2/" 2>&1); then
```
emosbaugh, line 81: it is very unclear to me the intention of this change and how this is testing containerd

emosbaugh (changes requested):

scripts/common/proxy.sh
camilamacedo86: @emosbaugh,

Our goal is validate if is possible to connect to registry, so we are here mainly:

Getting the registry configured on the containerd (/etc/containerd/config.toml)

```
[plugins."io.containerd.grpc.v1.cri".registry.configs."10.96.2.223".tls]
  ca_file = "/etc/kubernetes/pki/ca.crt"
```

Then, we are call the registry API: 

```
$ curl --cacert /etc/kubernetes/pki/ca.crt https://10.96.2.223/v2/
{"errors":[{"code":"UNAUTHORIZED","message":"authentication required","detail":null}]}
```

If the proxy be invalid for example or we have any issue to connect to the registry we would not face UNAUTHORIZED or `{}` (success when is possible authenticate)

Have you a better idea how can we verify the scenario?

emosbaugh: But what is the expectation here? Is it supposed to be using the proxy? Curl is not configured the same as containerd. This will behave differently depending on the environment.

camilamacedo86: Closing this one as sorted out since we agreed to change it out.

scripts/common/proxy.sh
```diff
@@ -47,6 +47,46 @@ function proxy_bootstrap() {
...
+            return
+        fi
+
+        # Get proxy configuration from http-proxy.conf file
+        local http_proxy=$(grep -E '^\s*Environment\s*=\s*"HTTP_PROXY=' "$proxy_config_file" | sed -E 's/^\s*Environment\s*=\s*"HTTP_PROXY=(.*)"/\1/')
+        local https_proxy=$(grep -E '^\s*Environment\s*=\s*"HTTPS_PROXY=' "$proxy_config_file" | sed -E 's/^\s*Environment\s*=\s*"HTTPS_PROXY=(.*)"/\1/')
+        local no_proxy=$(grep -E '^\s*Environment\s*=\s*"NO_PROXY=' "$proxy_config_file" | sed -E 's/^\s*Environment\s*=\s*"NO_PROXY=(.*)"/\1/')
+
+        # Make the Docker image pull call with proxy configuration
+        if response=$(sudo ctr image pull --http-proxy "$http_proxy" --https-proxy "$https_proxy" --no-proxy "$no_proxy" docker.io/library/nginx:latest 2>&1) && [[ $response == *"proxy"* ]]; then
```
ricardomaraschini, line 79: I think we could use an image we own for this, or at least a smaller one. What do you think ?

scripts/common/proxy.sh
```diff
@@ -47,6 +47,46 @@ function proxy_bootstrap() {
...
 
+# check_proxy_config tries to check if is possible connect with the registry
+# the following code will check if the proxy is invalid by doing a check as following
+# ~$ sudo HTTP_PROXY=http://77.89.35.50:8080 HTTPS_PROXY=http://77.89.35.50:8080 ctr image pull docker.io/library/nginx:latest
+# docker.io/library/nginx:latest: resolving      |--------------------------------------|
+# elapsed: 29.9s                  total:   0.0 B (0.0 B/s)
+# INFO[0030] trying next host                              error="failed to do request: Head \"https://registry-1.docker.io/v2/library/nginx/manifests/latest\": proxyconnect tcp: dial tcp 77.89.35.50:8080: i/o timeout" host=registry-1.docker.io
+# ctr: failed to resolve reference "docker.io/library/nginx:latest": failed to do request: Head "https://registry-1.docker.io/v2/library/nginx/manifests/latest": proxyconnect tcp: dial tcp 77.89.35.50:8080: i/o timeout
+function check_proxy_config() {
+    if [ -n "$CONTAINERD_VERSION" ]; then
```
ricardomaraschini, line 58: We could invert the logic here and then we would not need to indent the whole function body. What do you think ?

scripts/common/proxy.sh
```diff
@@ -47,6 +47,46 @@ function proxy_bootstrap() {
...
+        logStep "Checking proxy configuration with Containerd"
+
+        # Echo containerd Proxy config:
+        local proxy_config_file="/etc/systemd/system/containerd.service.d/http-proxy.conf"
+        if [ -f "$proxy_config_file" ]; then
```
ricardomaraschini, line 63: Here the logic could also be inverted avoiding an unneeded `else` statement.

scripts/common/proxy.sh
emosbaugh, line 79: is the message from this command helpful in debugging?

scripts/common/proxy.sh
```diff
@@ -47,6 +47,46 @@ function proxy_bootstrap() {
...
+            log "Proxy config:"
+            grep -v -e '^\[Service\]' -e '^# Generated by kURL' "$proxy_config_file"
+            echo ""
+        else
+            log "Skipping test. No HTTP proxy configuration found."
+            return
+        fi
+
+        # Get proxy configuration from http-proxy.conf file
+        local http_proxy=$(grep -E '^\s*Environment\s*=\s*"HTTP_PROXY=' "$proxy_config_file" | sed -E 's/^\s*Environment\s*=\s*"HTTP_PROXY=(.*)"/\1/')
```
ricardomaraschini, line 74: Do we need to read these from the containerd configuration ? I mean, if containerd is configured to use proxy then we would not need to pass this in the `ctr` command line. What am I not seeing here ?

camilamacedo86: Unfortunately, that's correct. If we simply use the ctr image pull command with an invalid configuration, it doesn't fail or raise any errors. This behavior suggests that when calling ctr directly, it doesn't utilize the settings that are configured for pulling images defined within a Deployment, for instance.

scripts/common/proxy.sh
camilamacedo86: We are outputting the config and if an issue be identified we are letting the user know that the problem is related with the proxy and we link the docs The docs are not great now. But is were we centralize the info right. so we can improve it in the long run.

Otherwise, what would be your suggestion here?

camilamacedo86: The image actually does not matter, so I update it to make clear. We check the error and we just inform that when the problem is proxy. If the error is image does not exist and etc then it will not raise the warning.

See:

```
⚙  Checking proxy configuration with Containerd

Proxy config:
Environment="HTTP_PROXY=http://invalidproxy:1234"
Environment="HTTPS_PROXY=http://invalidproxy:1234"

Proxy connection issues were identified
Please review the proxy configuration and ensure that it is valid.
More info: https://kurl.sh/docs/install-with-kurl/proxy-installs
```

The error for this call is:

```
~$ sudo HTTP_PROXY=http://invalidproxy:1234 HTTPS_PROXY=http://invalidproxy:1234 NO_PROXY= ctr image pull test/invalid/image:latest
INFO[0000] trying next host                              error="failed to do request: Head \"https://test/v2/invalid/image/manifests/latest\": proxyconnect tcp: dial tcp: lookup invalidproxy: Temporary failure in name resolution" host=test
ctr: failed to resolve reference "test/invalid/image:latest": failed to do request: Head "https://test/v2/invalid/image/manifests/latest": proxyconnect tcp: dial tcp: lookup invalidproxy: Temporary failure in name resolution
```

emosbaugh: But that message is helpful "Temporary failure in name resolution"

Other messages could be "Connection timeout", etc...

camilamacedo86: I see. Your suggestion is addressed now, see:

```sh
⚙  Checking proxy configuration with Containerd

Proxy config:
Environment="HTTP_PROXY=http://invalidproxy:1234"
Environment="HTTPS_PROXY=http://invalidproxy:1234"

Proxy connection issues were identified:
Proxy error: Temporary failure in name resolution

Please review the proxy configuration and ensure that it is valid.
More info: https://kurl.sh/docs/install-with-kurl/proxy-installs
```

scripts/common/proxy.sh
```diff
@@ -48,6 +48,53 @@ function proxy_bootstrap() {
...
+    grep -v -e '^\[Service\]' -e '^# Generated by kURL' "$proxy_config_file"
+    echo ""
+
+    # Get proxy configuration from http-proxy.conf file
+    local http_proxy=$(grep -E '^\s*Environment\s*=\s*"HTTP_PROXY=' "$proxy_config_file" | sed -E 's/^\s*Environment\s*=\s*"HTTP_PROXY=(.*)"/\1/')
+    local https_proxy=$(grep -E '^\s*Environment\s*=\s*"HTTPS_PROXY=' "$proxy_config_file" | sed -E 's/^\s*Environment\s*=\s*"HTTPS_PROXY=(.*)"/\1/')
+    local no_proxy=$(grep -E '^\s*Environment\s*=\s*"NO_PROXY=' "$proxy_config_file" | sed -E 's/^\s*Environment\s*=\s*"NO_PROXY=(.*)"/\1/')
+
+    # Make the Docker image pull call with proxy configuration
+    if ! response=$(sudo HTTP_PROXY=$http_proxy HTTPS_PROXY=$https_proxy NO_PROXY=$no_proxy ctr image pull test/invalid/image:latest 2>&1) || [[ $response =~ .*"proxy".* ]]; then
```
ricardomaraschini, line 83: We don't need this `sudo`, we are already root.

ricardomaraschini, line 83: It might be a good idea to use `crictl` instead of `ctr` directly. The first seems to respect whatever proxy is configured on containerd.

scripts/common/proxy.sh
```diff
@@ -48,6 +48,53 @@ function proxy_bootstrap() {
...
+
+    echo ""
+    log "Proxy config:"
+    grep -v -e '^\[Service\]' -e '^# Generated by kURL' "$proxy_config_file"
+    echo ""
+
+    # Get proxy configuration from http-proxy.conf file
+    local http_proxy=$(grep -E '^\s*Environment\s*=\s*"HTTP_PROXY=' "$proxy_config_file" | sed -E 's/^\s*Environment\s*=\s*"HTTP_PROXY=(.*)"/\1/')
+    local https_proxy=$(grep -E '^\s*Environment\s*=\s*"HTTPS_PROXY=' "$proxy_config_file" | sed -E 's/^\s*Environment\s*=\s*"HTTPS_PROXY=(.*)"/\1/')
+    local no_proxy=$(grep -E '^\s*Environment\s*=\s*"NO_PROXY=' "$proxy_config_file" | sed -E 's/^\s*Environment\s*=\s*"NO_PROXY=(.*)"/\1/')
```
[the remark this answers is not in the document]
ricardomaraschini: Have you noticed that "$http_proxy" value is set to the string `http://1.1.1.1" "HTTPS_PROXY=http://2.2.2.2" "NO_PROXY=localhost,company.com` while "$https_proxy" and "$no_proxy" values are empty?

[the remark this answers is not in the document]
camilamacedo86: Using crictl as you suggested we no longer need this one so I think we can mark it as sorted out 👍

scripts/common/proxy.sh
camilamacedo86: Done

scripts/common/proxy.sh
```diff
@@ -48,6 +48,45 @@ function proxy_bootstrap() {
...
+
+    if ! response=$(crictl pull test/invalid/image:latest 2>&1) || [[ $response =~ .*"proxy".* ]]; then
+        logWarn "Proxy connection issues were identified:"
+        error_message=$(echo "$response" | grep -oP '(?<=failed to do request: ).*' | sed -r 's/.*: //' | awk -F "\"" '{print $(NF-1)}' | sed -r 's/test\/invalid\/image:latest//')
+        logWarn "Proxy error: $error_message"
+        echo ""
+        logWarn "Please review the proxy configuration and ensure that it is valid."
+        logWarn "More info: https://kurl.sh/docs/install-with-kurl/proxy-installs"
+        return
+    fi
```
camilamacedo86, line 85: @ricardomaraschini @emosbaugh  example:

```sh
⚙  Checking proxy configuration with Containerd

Proxy config:
Environment="HTTP_PROXY=http://invalidproxy:1234"
Environment="HTTPS_PROXY=http://invalidproxy:1234"

Proxy connection issues were identified:
Proxy error: 
Temporary failure in name resolution

Please review the proxy configuration and ensure that it is valid.
More info: https://kurl.sh/docs/install-with-kurl/proxy-installs
```

scripts/common/proxy.sh
```diff
@@ -48,6 +48,45 @@ function proxy_bootstrap() {
...
+        log "Skipping test. No HTTP proxy configuration found."
+        return
+    fi
+
+    echo ""
+    log "Proxy config:"
+    grep -v -e '^\[Service\]' -e '^# Generated by kURL' "$proxy_config_file"
+    echo ""
+
+    if ! response=$(crictl pull test/invalid/image:latest 2>&1) || [[ $response =~ .*"proxy".* ]]; then
```
emosbaugh, line 77: Have you tested this with a valid proxy? Wont the pull fail since the image is invalid? Will it not enter this conditional in that case?

scripts/common/proxy.sh
camilamacedo86: You can either check the example of the command result called directly:

```
$ sudo crictl pull test/invalid/image:latest
E0525 09:01:01.952576 1399831 remote_image.go:167] "PullImage from image service failed" err="rpc error: code = Unknown desc = failed to pull and unpack image \"docker.io/test/invalid/image:latest\": failed to resolve reference \"docker.io/test/invalid/image:latest\": failed to do request: Head \"https://registry-1.docker.io/v2/test/invalid/image/manifests/latest\": proxyconnect tcp: dial tcp: lookup invalidproxy: Temporary failure in name resolution" image="test/invalid/image:latest"
FATA[0000] pulling image: rpc error: code = Unknown desc = failed to pull and unpack image "docker.io/test/invalid/image:latest": failed to resolve reference "docker.io/test/invalid/image:latest": failed to do request: Head "https://registry-1.docker.io/v2/test/invalid/image/manifests/latest": proxyconnect tcp: dial tcp: lookup invalidproxy: Temporary failure in name resolution 
```

scripts/common/proxy.sh
camilamacedo86, line 77:
```suggestion
    if ! response=$(crictl pull test/invalid/image:latest 2>&1) && [[ $response =~ .*"proxy".* ]]; then
```

@emosbaugh,

Good catcher. After all changes I did not re-tested the success case and the || would make that fail.
Fixed and re-tested:

```
$ sudo ./last.sh 
⚙  Checking proxy configuration with Containerd

Proxy config:
Environment="HTTP_PROXY=http://64.225.8.191:9992"
Environment="HTTPS_PROXY=http://64.225.8.191:9992"

✔ Unable to identify proxy problems
```

scripts/common/proxy.sh
```diff
@@ -48,6 +48,45 @@ function proxy_bootstrap() {
     fi
 }
 
+# check_proxy_config tries to check if is possible connect with the registry
+# the following code will check if the proxy is invalid by doing a check as following
+# ~$ sudo HTTP_PROXY=http://77.89.35.50:8080 HTTPS_PROXY=http://77.89.35.50:8080 ctr image pull docker.io/library/nginx:latest
```
ricardomaraschini, line 53: This comment became outdated.

scripts/common/proxy.sh
```diff
@@ -48,6 +48,44 @@ function proxy_bootstrap() {
...
+
+    echo ""
+    log "Proxy config:"
+    grep -v -e '^\[Service\]' -e '^# Generated by kURL' "$proxy_config_file"
+    echo ""
+
+    if ! response=$(crictl pull test/invalid/image:latest 2>&1) && [[ $response =~ .*"proxy".* ]]; then
+        logWarn "Proxy connection issues were identified:"
+        error_message=$(echo "$response" | grep -oP '(?<=failed to do request: ).*' | sed -r 's/.*: //' | awk -F "\"" '{print $(NF-1)}' | sed -r 's/test\/invalid\/image:latest//')
+        logWarn "Proxy error: $error_message"
```
camilamacedo86, line 79:
```suggestion
```

ricardomaraschini (approved):

emosbaugh (approved):

PR #4523, opened 2023-05-20
(do not merge - Requires discussions to address concerns and a window for automate tests after merge - backlog) replace patchesJson6902 by patches

camilamacedo86:

Fix patcgesJson6902 which is deprecated

> Warning: 'patchesJson6902' is deprecated. Please use 'patches' instead. Run 'kustomize edit fix' to update your Kustomization automatically.

Fixes # [sc-77464]

We will need to check it out in the daily as well. 
So, let's only get merged after the next release

```release-note
Fixes warning patchesJson6902 which is deprecated by replacing it for patches. AddOns prometheus versions equals or upper than `0.46.0` and Rook versions equals or upper than  `1.4.9` were changed to receive this fix. 
```

### DIFF
--- a/addons/prometheus/0.46.0-14.9.0/crds/kustomization.yaml
+++ b/addons/prometheus/0.46.0-14.9.0/crds/kustomization.yaml
@@ -1,7 +1,7 @@
 resources:
 - crds.yaml
 
-patchesJson6902:
+patches:
 - target:
     group: "apiextensions.k8s.io"
     version: v1 # apiVersion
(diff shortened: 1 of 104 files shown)


emosbaugh:
@rrpolanco looked into this. you should discuss the challenges with him

camilamacedo86:
Hi @emosbaugh,

>@rrpolanco looked into this. you should discuss the challenges with him

Thank you for the input. I think the challenges faced by @rrpolanco was because the change should work in old and new k8s apis. In this case, the patches is used by kustomize.

We distributed kustomize with the installer.
Therefore, the syntax of the files must work with the current version used by and we have not the same challanges.
Am I right?

camilamacedo86:
I am closing this one because I am afraid that this change may cause `invalid Kustomization: json: cannot unmarshal string into Go struct field Kustomization.patches of type types.Patch`

It end up not so simple as it shows at the first moment and we might need to better think and discuss how we will do that.

PR #4524, opened 2023-05-22
feat: Call multi-node storage migration from kURL installer scripts

ricardomaraschini:

This PR creates the function `rook_maybe_migrate_from_openebs`. This function may use Ekco to migrate data from OpenEBS (`scaling` storage class) to Rook (`distributed` storage class) when all the following conditions are met:

- Ekco version is >= 0.27.1.
- OpenEBS and Rook are selected on the installer.
- Rook's minimum node count is set to a value > 1.
- The number of Nodes on the cluster is >= than the Rook minimum node count.
- The `scaling` storage class exists.

`rook_maybe_migrate_from_openebs` is invoked from `install.sh` and `join.sh`. A new task, `migrate-to-multinode-storage` was added to initiate the multi-node storage migration from outside the installation workflow.

### DIFF
--- a/scripts/common/rook.sh
+++ b/scripts/common/rook.sh
@@ -490,6 +490,16 @@ function rook_maybe_migrate_from_openebs() {
 # rook_maybe_migrate_from_openebs_primary will call rook_maybe_migrate_from_openebs
 # after determining values for EKCO_AUTH_TOKEN and EKCO_ADDRESS from the cluster.
 function rook_maybe_migrate_from_openebs_primary() {
-    #TODO
+    local ekcoAddress=
+    local ekcoAuthToken=
+    ekcoAddress=$(get_ekco_addr)
+    ekcoAuthToken=$(get_ekco_storage_migration_auth_token)
+    if [ -z "$ekcoAddress" ] || [ -z "$ekcoAuthToken" ]; then
+        return 0
+    fi
+
+    export EKCO_ADDRESS="$ekcoAddress"
+    export EKCO_AUTH_TOKEN="$ekcoAuthToken"
+
     rook_maybe_migrate_from_openebs
 }
--- a/scripts/common/rook.sh
+++ b/scripts/common/rook.sh
@@ -432,3 +432,47 @@ function add_rook_store_object_migration_status() {
     kubectl patch configmap kurl-migration-from-rook -n kurl --type merge -p '{"data":{"DID_MIGRATE_ROOK_OBJECT_STORE":"1"}}'
     export DID_MIGRATE_ROOK_OBJECT_STORE=1
 }
+
+# rook_maybe_migrate_from_openebs may migrate data from OpenEBS to Rook when all the following
+# conditions are met:
+# - Ekco version is >= 0.27.1.
+# - OpenEBS and Rook are selected on the Installer.
+# - Rook's minimum node count is set to a value > 1.
+# - The number of nodes on the cluster is >= than the Rook minimum node count.
+# - The 'scaling' storage class exists.
+function rook_maybe_migrate_from_openebs() {
+    semverCompare "$EKCO_VERSION" "0.27.1"
+    if [ "$SEMVER_COMPARE_RESULT" -lt "0" ]; then
+        return 0
+    fi
+    if [ -z "$ROOK_VERSION" ] || [ -z "$OPENEBS_VERSION" ]; then
+        return 0
+    fi
+    if [ -z "$ROOK_MINIMUM_NODE_COUNT" ] || [ "$ROOK_MINIMUM_NODE_COUNT" -le "1" ]; then
+        return 0
+    fi
+    local num_nodes
+    num_nodes="$(kubectl get nodes --no-headers | wc -l)"
+    if [ "$num_nodes" -lt "$ROOK_MINIMUM_NODE_COUNT" ]; then
+        return 0
+    fi
+    if ! kubectl get sc scaling >/dev/null 2>&1 ; then
+        return 0
+    fi
+    logWarn "    The installer detected both OpenEBS and Rook installations in your cluster. Migration from OpenEBS to Rook"
+    logWarn "    is possible now, but it requires scaling down applications using OpenEBS volumes, causing downtime. You can"
+    logWarn "    choose to run the migration later if preferred."
+    log "Would you like to continue with the migration now? "
+    if ! confirmN; then
+        logWarn "Migration from OpenEBS to Rook skipped."
+        logWarn "If you would like to run the migration later, run the following command:"
+        logWarn "    $DIR/bin/kurl cluster migrate-multinode-storage"
+        return 0
+    fi
+    if ! $DIR/bin/kurl cluster migrate-multinode-storage; then
+        logFail "Failed to migrate from OpenEBS to Rook. The installation will move on."
+        logFail "If you would like to run the migration later, run the following command:"
+        logFail "    $DIR/bin/kurl cluster migrate-multinode-storage"
+        return 0
+    fi
+}
--- a/scripts/common/rook.sh
+++ b/scripts/common/rook.sh
@@ -464,10 +464,10 @@ function rook_maybe_migrate_from_openebs() {
 
     # check if OpenEBS to Rook multi-node migration is available - if it is, prompt the user to start it
     if "${DIR}"/bin/kurl cluster migrate-multinode-storage --ekco-address "$EKCO_ADDRESS" --ekco-auth-token "$EKCO_AUTH_TOKEN" --check-status; then
-        printf "    The installer detected both OpenEBS and Rook installations in your cluster. Migration from OpenEBS to Rook"
-        printf "    is possible now, but it requires scaling down applications using OpenEBS volumes, causing downtime. You can"
-        printf "    choose to run the migration later if preferred."
-        printf "Would you like to continue with the migration now? "
+        printf "    The installer detected both OpenEBS and Rook installations in your cluster. Migration from OpenEBS to Rook\n"
+        printf "    is possible now, but it requires scaling down applications using OpenEBS volumes, causing downtime. You can\n"
+        printf "    choose to run the migration later if preferred.\n"
+        printf "Would you like to continue with the migration now? \n"
         if ! confirmN ; then
             printf "Not migrating from OpenEBS to Rook\n"
             return 0
--- a/scripts/common/rook.sh
+++ b/scripts/common/rook.sh
@@ -471,3 +471,9 @@ function rook_maybe_migrate_from_openebs() {
     fi
 }
 
+# rook_maybe_migrate_from_openebs_primary will call rook_maybe_migrate_from_openebs
+# after determining values for EKCO_AUTH_TOKEN and EKCO_ADDRESS from the cluster.
+function rook_maybe_migrate_from_openebs_primary() {
+    #TODO
+    rook_maybe_migrate_from_openebs
+}
--- a/scripts/common/rook.sh
+++ b/scripts/common/rook.sh
@@ -457,7 +457,7 @@ function rook_maybe_migrate_from_openebs() {
 # rook_maybe_migrate_from_openebs_internal SHOULD NOT BE CALLED DIRECTLY.
 # it is called by rook_maybe_migrate_from_openebs and rook_maybe_migrate_from_openebs_tasks when all the conditions are met.
 # it will check that the required environment variables (EKCO_AUTH_TOKEN and EKCO_ADDRESS) are set and then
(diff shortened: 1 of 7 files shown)


scripts/common/rook.sh
```diff
@@ -432,3 +432,50 @@ function add_rook_store_object_migration_status() {
...
+        logWarn " Migration from OpenEBS to Rook skipped."
+        logWarn " If you would like to run the migration later, run the following command:"
+        logWarn "    $DIR/bin/kurl cluster migrate-multinode-storage"
+        return 0
+    fi
+    if ! $DIR/bin/kurl cluster migrate-multinode-storage; then
+        logFail "Failed to migrate from OpenEBS to Rook. The installation will move on."
+        logFail "If you would like to run the migration later, run the following command:"
+        logFail "    $DIR/bin/kurl cluster migrate-multinode-storage"
+        return 0
```
ricardomaraschini, line 479: I have chosen to let the installation to move on because the migration can be manually retried later on.

scripts/common/rook.sh
```diff
@@ -432,3 +432,47 @@ function add_rook_store_object_migration_status() {
...
+    fi
+    local num_nodes
+    num_nodes="$(kubectl get nodes --no-headers | wc -l)"
+    if [ "$num_nodes" -lt "$ROOK_MINIMUM_NODE_COUNT" ]; then
+        return 0
+    fi
+    if ! kubectl get sc scaling >/dev/null 2>&1 ; then
+        return 0
+    fi
+    logWarn "    The installer detected both OpenEBS and Rook installations in your cluster. Migration from OpenEBS to Rook"
```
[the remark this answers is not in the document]
ricardomaraschini: Keeping with concise with [this](https://github.com/replicatedhq/kURL/blob/main/addons/openebs/template/base/install.sh#L134-L143), [this](https://github.com/replicatedhq/kURL/blob/main/addons/openebs/template/base/install.sh#L407-L417), and [this](https://github.com/replicatedhq/kURL/blob/main/addons/rook/template/base/install.sh#L1027-L1051).

rrpolanco (dismissed):

scripts/common/rook.sh
```diff
@@ -432,3 +432,47 @@ function add_rook_store_object_migration_status() {
...
 }
+
+# rook_maybe_migrate_from_openebs may migrate data from OpenEBS to Rook when all the following
+# conditions are met:
+# - Ekco version is >= 0.27.1.
+# - OpenEBS and Rook are selected on the Installer.
+# - Rook's minimum node count is set to a value > 1.
+# - The number of nodes on the cluster is >= than the Rook minimum node count.
+# - The 'scaling' storage class exists.
+function rook_maybe_migrate_from_openebs() {
```
rrpolanco, line 443: @ricardomaraschini I'm curious why you decided to put this function here rather than in the Rook add-on code?

camilamacedo86: Because we have rook.sh whcih has the common functions for Rook. Therefore, I think that is the reason.

camilamacedo86 (approved):
It seems ok for me 👍 
Great work 🥇

scripts/tasks.sh
```diff
@@ -900,6 +908,45 @@ function flannel_images_present() {
...
+        bail "EKCO add-on version >= $requiredEkcoVersion required for migration"
+    fi
+
+    local ekcoAddress=
+    local ekcoAuthToken=
+    ekcoAddress=$(get_ekco_addr)
+    ekcoAuthToken=$(get_ekco_storage_migration_auth_token)
+
+    # Initiate OpenEBS to Rook multi-node migration
+    if ! "${DIR}"/bin/kurl cluster migrate-multinode-storage --ekco-address "$ekcoAddress" --ekco-auth-token "$ekcoAuthToken" < /dev/tty; then
```
emosbaugh, line 942: you may want to wrap this in the `prompts_can_prompt` function. also if the prompt should default to no we have the flag `yes` that you should support.

scripts/common/rook.sh
```diff
@@ -432,3 +432,42 @@ function add_rook_store_object_migration_status() {
...
+        return 0
+    fi
+
+    if [ -z "$EKCO_ADDRESS" ]; then
+        logFail "Internal Error: unable to determine network address of the kURL operator."
+        return 0
+    fi
+
+    # Initiate OpenEBS to Rook multi-node migration
+    if ! "${DIR}"/bin/kurl cluster migrate-multinode-storage --ekco-address "$EKCO_ADDRESS" --ekco-auth-token "$EKCO_AUTH_TOKEN" < /dev/tty; then
```
emosbaugh, line 466: you may want to wrap this in the prompts_can_prompt function. also if the prompt should default to no we have the flag yes that you should support.

scripts/install.sh
```diff
@@ -476,7 +476,7 @@ function outro() {
         printf "To add worker nodes to this installation, copy and unpack this bundle on your other nodes, and run the following:"
         printf "\n"
         printf "\n"
-        printf "${GREEN}    cat ./join.sh | sudo bash -s airgap kubernetes-master-address=${API_SERVICE_ADDRESS} kubeadm-token=${BOOTSTRAP_TOKEN} kubeadm-token-ca-hash=${KUBEADM_TOKEN_CA_HASH} kubernetes-version=${KUBERNETES_VERSION}${common_flags}\n"
+        printf "${GREEN}    cat ./join.sh | sudo bash -s airgap kubernetes-master-address=${API_SERVICE_ADDRESS} kubeadm-token=${BOOTSTRAP_TOKEN} kubeadm-token-ca-hash=${KUBEADM_TOKEN_CA_HASH} kubernetes-version=${KUBERNETES_VERSION} ekco-address=${EKCO_ADDRESS} ekco-auth-token=${EKCO_AUTH_TOKEN}${common_flags}\n"
```
rrpolanco, line 479: These new flags, `ekco-address=${EKCO_ADDRESS} ekco-auth-token=${EKCO_AUTH_TOKEN}` should be conditionally added to these commands when applicable. Ditto everywhere else where used.

scripts/common/rook.sh
```diff
@@ -432,3 +432,42 @@ function add_rook_store_object_migration_status() {
...
+        return 0
+    fi
+    if [ -z "$ROOK_VERSION" ] || [ -z "$OPENEBS_VERSION" ]; then
+        return 0
+    fi
+    if [ -z "$ROOK_MINIMUM_NODE_COUNT" ] || [ "$ROOK_MINIMUM_NODE_COUNT" -le "1" ]; then
+        return 0
+    fi
+
+    if [ -z "$EKCO_AUTH_TOKEN" ]; then
```
emosbaugh, line 455: will this and EKCO_ADDRESS be set in install.sh? or is there logic in the get_ekco_addr and get_ekco_storage_migration_auth_token functions to get those values from the cluster?

scripts/common/rook.sh
emosbaugh: my question still stands here. what happens if there is no terminal? say this is invoked from automation.

scripts/tasks.sh
emosbaugh: my question still stands here. what happens if there is no terminal? say this is invoked from automation.

emosbaugh (changes requested):

scripts/common/rook.sh
laverya: Because this isn't run as part of the rook addon, it's run as part of the join/upgrade code

scripts/common/rook.sh
laverya: it looks like it isn't currently set in install.sh, that'll be fixed

scripts/common/rook.sh
laverya: will do

scripts/tasks.sh
laverya: this should really share code with rook_maybe_migrate_from_openebs too

rrpolanco: If there is no TTY I would think we should default to _not_ initiate the migration and issue a message with the command which the user can then use to trigger the migration at a later time. For automation purposes, we could also add a `-y`/`--asume-yes` flag to the `cluster migrate-multinode-storage` command.

scripts/common/rook.sh
laverya: fixed

scripts/common/rook.sh
```diff
@@ -432,3 +432,74 @@ function add_rook_store_object_migration_status() {
...
+    local ekcoAddress=
+    local ekcoAuthToken=
+    ekcoAddress=$(get_ekco_addr)
+    ekcoAuthToken=$(get_ekco_storage_migration_auth_token)
+    if [ -z "$ekcoAddress" ] || [ -z "$ekcoAuthToken" ]; then
+        return 0
+    fi
+
+    export EKCO_ADDRESS="$ekcoAddress"
+    export EKCO_AUTH_TOKEN="$ekcoAuthToken"
```
rrpolanco, line 502: These vars are already set in the ekco-addon code. See my previous comment: https://github.com/replicatedhq/kURL/pull/4524/files#r1245649223

scripts/tasks.sh
```diff
@@ -900,6 +908,25 @@ function flannel_images_present() {
...
+    # Is Rook and OpenEBS installed
+    if ! kubectl get ns | grep -q rook-ceph && ! kubectl get ns | grep -q openebs; then
+        bail "Rook and OpenEBS must be installed in order to migrate to multi-node"
+    fi
+
+    # Get Rook.minimumNodeCount option from the configmap
+    local rookMinNodes=
+    rookMinNodes=$(kubectl get cm kurl-current-config -n kurl -ojsonpath='{.data.addons-rook}' | base64 -d | tr "," " "| awk '{print $1}' | cut -d ":" -f 2)
+    if [ -z "$rookMinNodes" ] || [ "$rookMinNodes" -le "1" ]; then
+        bail "Rook.MinimumNodeCount must be greater than 1"
```
rrpolanco, line 924:
```suggestion
    if [ -z "$rookMinNodes" ] || [ "$rookMinNodes" -lt "3" ]; then
        bail "Rook.MinimumNodeCount must be greater than or equal to 3"
```

scripts/common/rook.sh
laverya: Oh, I'll take another look!

laverya: It still needed for tasks.sh though

laverya: updated so only tasks.sh uses this

rrpolanco (approved):
Code lgtm. Just need to test.

emosbaugh (dismissed):

scripts/tasks.sh
```diff
@@ -911,11 +911,7 @@ function flannel_images_present() {
 # Determine if the cluster is eligible for a multi-node migration and then initiate the migration
 function migrate_to_multinode_storage() {
     export KUBECONFIG=/etc/kubernetes/admin.conf
-
-    # Is Rook and OpenEBS installed
-    if ! kubectl get ns | grep -q rook-ceph && ! kubectl get ns | grep -q openebs; then
-        bail "Rook and OpenEBS must be installed in order to migrate to multi-node"
-    fi
+    download_util_binaries
```
rrpolanco, line 914: why do we need to invoke this function?

laverya: Because the binary we're calling is installed by it

laverya: `"${DIR}"/bin/kurl` I mean

scripts/tasks.sh
```diff
@@ -900,6 +908,21 @@ function flannel_images_present() {
...
+# Determine if the cluster is eligible for a multi-node migration and then initiate the migration
+function migrate_to_multinode_storage() {
+    export KUBECONFIG=/etc/kubernetes/admin.conf
+    download_util_binaries
+
+    # Get Rook.minimumNodeCount option from the configmap
+    local rookMinNodes=
+    rookMinNodes=$(kubectl get cm kurl-current-config -n kurl -ojsonpath='{.data.addons-rook}' | base64 -d | tr "," " "| awk '{print $1}' | cut -d ":" -f 2)
+    if [ -z "$rookMinNodes" ] || [ "$rookMinNodes" -lt "3" ]; then
+        bail "Rook.MinimumNodeCount must be greater than or equal to 3"
```
rrpolanco, line 920:
```suggestion
        bail "Rook.minimumNodeCount must be greater than or equal to 3"
```

scripts/tasks.sh
rrpolanco: hmm, ok. I thought this binary already exist on the node since the installer would have already downloaded the binaries.

rrpolanco (approved):

PR #4534, opened 2023-05-22
Adds support for RHEL 9.2

camilamacedo86:
#### What this PR does / why we need it:

Adds support for RHEL 9.2

#### Which issue(s) this PR fixes:

Fixes # [sc-76521]

#### Does this PR introduce a user-facing change?

```release-note
Adds support for RHEL 9.2
```

PR #4535, opened 2023-05-22
update unsupported tests targeting rocky-91 with rocky-92

camilamacedo86:
Follow up of : https://github.com/replicatedhq/kURL/pull/4534
[sc-76521]

### DIFF
--- a/testgrid/specs/deploy.yaml
+++ b/testgrid/specs/deploy.yaml
@@ -37,7 +37,7 @@
       version: latest
   unsupportedOSIDs:
   - ubuntu-2204 # docker 19.x is not available on ubuntu 22.04
-  - rocky-91 # docker is not supported on rhel 9 variants
+  - rocky-92 # docker is not supported on rhel 9 variants
 - name: k8s121
   installerSpec:
     kubernetes:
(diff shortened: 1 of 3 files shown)


testgrid/specs/deploy.yaml
```diff
@@ -37,7 +37,7 @@
       version: latest
   unsupportedOSIDs:
   - ubuntu-2204 # docker 19.x is not available on ubuntu 22.04
-  - rocky-91 # docker is not supported on rhel 9 variants
```
emosbaugh, line 40: wouldnt it be best to append rocky-92 to the array in all cases rather than replace rocky-91? in case we decide to add back 9.1 to any of the suites

emosbaugh (changes requested):

testgrid/specs/deploy.yaml
camilamacedo86: If we decide to add back 9.1 then we can just find replace in all project by - rocky-91 and replace by : - rocky-91 \n - rocky-92. It is very low effort, I think the effort so we do not need to have a data/info that is unused.

emosbaugh (approved):

PR #4537, opened 2023-05-23
testgrid: verify proxy config installs/upgrades

camilamacedo86:

This PR adds the required logic to check if containerd is properly configured with the proxy info provided

 
Fixes #

See: https://testgrid.kurl.sh/run/CAMILA-TEST-AIRGAP-Check-containerd-config-Prority-0

```release-note

```

### DIFF
--- a/testgrid/specs/full.yaml
+++ b/testgrid/specs/full.yaml
@@ -2269,7 +2269,7 @@
         - .corporate.internal
       noProxy: false
     kubernetes:
-      version: 1.27.x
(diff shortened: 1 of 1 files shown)


camilamacedo86:
This PR was opened to check a scenario and as a draft of to check see if we could address this check.
However, we need be able to pass the ProxyAddress flag prior to the nodes.
I added a comment in the internal task (sc-77456) for we be able to check this draft and see if we could do a test like that after we accomplished this goal.

PR #4538, opened 2023-05-23
bug: configuring containerd and docker https proxy

ricardomaraschini:

Setting up `containerd` and `docker` to use `https_proxy` environment variable as its proxy.  Prior to this we were always setting `containerd` and `docker` to use the HTTP proxy instead.

Fixes: [sc-70304] [sc-77636]

```release-note
The installer has been enhanced to utilize the HTTP_PROXY and HTTPS_PROXY environment variables during the configuration of the container runtime. 
```


camilamacedo86 (approved):

camilamacedo86 (commented):
@ricardomaraschini,

/LGTM

Just a nit. Could you please add in the description of the PR

Fixes: [sc-70304] [sc-77636]

See that your PR solves both cases

Also, could you please add the message for the release notes so that we can properly track it down?

scripts/common/proxy.sh
```diff
@@ -1,53 +1,94 @@
...
+# HTTP_PROXY both must point to the same address. this function sets the
+# following variables ENV_HTTP_PROXY_ADDRESS, ENV_HTTPS_PROXY_ADDRESS,
+# and ENV_NO_PROXY.
+function read_proxy_config_from_env() {
+    # by default https proxy configuration inherits from http proxy config.
+    # we also make sure that we have both environment variables (with upper
+    # and lower case) set to the same value.
     if [ -n "$HTTP_PROXY" ]; then
-        ENV_PROXY_ADDRESS="$HTTP_PROXY"
-        export https_proxy="$HTTP_PROXY"
```
emosbaugh, line 5: this used to always export https_proxy and now that has changed. i imagine because modern linux tools reliy on that env var and not the others.

scripts/common/proxy.sh
```diff
@@ -1,53 +1,94 @@
...
+    # (with upper and lower case) set to the same value.
+    if [ -n "$HTTPS_PROXY" ]; then
+        ENV_HTTPS_PROXY_ADDRESS="$HTTPS_PROXY"
+        export https_proxy="$HTTPS_PROXY"
     elif [ -n "$https_proxy" ]; then
-        ENV_PROXY_ADDRESS="$https_proxy"
-        printf "The installer will use the proxy at '%s' (imported from env var 'https_proxy')\n" "$ENV_PROXY_ADDRESS"
+        ENV_HTTPS_PROXY_ADDRESS="$https_proxy"
+        export HTTPS_PROXY="$https_proxy"
     fi
```
emosbaugh, line 29: you could export https_proxy as it did before like so

```suggestion
    fi

    if [ -n "$ENV_HTTPS_PROXY_ADDRESS" ]; then
        export https_proxy="$ENV_HTTPS_PROXY_ADDRESS"
    fi
```

scripts/common/proxy.sh
```diff
@@ -71,15 +114,28 @@ function configure_proxy() {
...
+    fi
+
+    # if user hasn't provide any proxy configuration we use the ones
+    # present in the environment.
+    PROXY_ADDRESS="$ENV_HTTP_PROXY_ADDRESS"
+    PROXY_HTTPS_ADDRESS="$ENV_HTTP_PROXY_ADDRESS"
+    if [ -n "$ENV_HTTPS_PROXY_ADDRESS" ]; then
+        PROXY_HTTPS_ADDRESS="$ENV_HTTPS_PROXY_ADDRESS"
     fi
-    export https_proxy="$PROXY_ADDRESS"
```
emosbaugh, line 81: once again i dont think you can lose this but it may be unnecessary if it is set for all cases above

emosbaugh (changes requested):

scripts/common/proxy.sh
ricardomaraschini: Are you saying that we should always export HTTPS_PROXY to the HTTP_PROXY value when the first is not set ?

scripts/common/proxy.sh
ricardomaraschini: Not sure if I follow, the idea here is to sanitize the inputs and guarantee that both lower and upper case versions of the variable point to the same value. `ENV_HTTPS_PROXY_ADDRESS` is read from `https_proxy` or `HTTPS_PROXY` so it is set both also are.

scripts/common/proxy.sh
ricardomaraschini: Let's discuss this [here](https://github.com/replicatedhq/kURL/pull/4538#discussion_r1204163307)

scripts/common/proxy.sh
emosbaugh: i just dont want to change the existing behavior. im worried we will introduce a regression. how about exporting all 4 at the end?

```bash
    if [ -n "$ENV_HTTP_PROXY_ADDRESS" ]; then
        export http_proxy="$ENV_HTTP_PROXY_ADDRESS"
        export HTTP_PROXY="$ENV_HTTP_PROXY_ADDRESS"
    fi
    if [ -n "$ENV_HTTPS_PROXY_ADDRESS" ]; then
        export https_proxy="$ENV_HTTPS_PROXY_ADDRESS"
        export HTTPS_PROXY="$ENV_HTTPS_PROXY_ADDRESS"
    fi
```

ricardomaraschini: Sounds good to me. I have added this to the last commit. Haven't tested it yet but I will come back here as soon as I do so.

ricardomaraschini: I have tested an install the proxies are being set correctly after the change.

scripts/common/proxy.sh
```diff
@@ -1,51 +1,91 @@
...
 
-    if [ -n "$ENV_PROXY_ADDRESS" ]; then
-        export https_proxy="$ENV_PROXY_ADDRESS"
-        kubectl_no_proxy
-        log "Bootstrapped proxy address from ENV_PROXY_ADDRESS: $https_proxy"
+    local proxy=$(echo "$INSTALLER_YAML" | grep "proxyAddress:" | grep -o "http[^'\" ]*")
+    if [ -z "$proxy" ]; then
         return
     fi
+    ENV_HTTP_PROXY_ADDRESS="$proxy"
```
emosbaugh, line 82: im worried that you no longer call `kubectl_no_proxy` in this case

emosbaugh (changes requested):

scripts/common/proxy.sh
```diff
@@ -1,51 +1,91 @@
...
+    # read and sanitize proxy configuration from environment variables.
+    read_proxy_config_from_env
+
+    # users can still provide a different proxy by patching the installer
+    # yaml, we need to verify if this is the case and then use the proxy
+    # set in the yaml for both http and https.
     if [ -n "$INSTALLER_SPEC_FILE" ]; then
         local overrideProxy=$(grep "proxyAddress:" "$INSTALLER_SPEC_FILE" | grep -o "http[^'\" ]*")
         if [ -n "$overrideProxy" ]; then
+            ENV_HTTP_PROXY_ADDRESS="$overrideProxy"
```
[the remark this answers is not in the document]
ricardomaraschini: Yeah, that is correct! I have fixed it by calling `kubectl_no_proxy` before returning and making the function idempotent. Haven't tested it yet but I will do so tomorrow. Would you mind taking a look at it?

scripts/common/proxy.sh
ricardomaraschini: This comment is similar to this [other](https://github.com/replicatedhq/kURL/pull/4538#discussion_r1210310437) comment on the same review. I will resolve this one and we can discuss on the first one.

emosbaugh (approved):

PR #4549, opened 2023-05-25
Refining tests done against each PR

camilamacedo86:

Currently, we run a full set of tests when we get a PR merged
However, it has been overloading the testsgrid

- This PR moves ALL Deploy tests to Full tests (executed once by day)
- Ensure that we have tests that will check k8s upgrades from 1.22 to 1.27 and the migrations: longhorn to openebs, openebs to rook, weave to flannel, docker to containerd. We also have a test covering airgap. 

So that we can have a good coverage PR pr to ensure that we did not break anything and reduce the amount of tests done by each case. 
 
See: https://testgrid.kurl.sh/run/CAMILA-TEST-DEPLOY-PR-FINAL-MORE_CSI

### DIFF
--- a/testgrid/specs/full.yaml
+++ b/testgrid/specs/full.yaml
@@ -2294,3 +2294,381 @@
       localPVStorageClassName: default
     minio:
       version: latest
+- name: longhorn
+  installerSpec:
+    contour:
+      version: latest
+    kurl:
+      installerVersion: ""
+    containerd:
+      version: latest
+    ekco:
+      version: latest
+    kubernetes:
+      version: 1.24.x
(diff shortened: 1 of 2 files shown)


testgrid/specs/full.yaml
```diff
@@ -2294,3 +2294,381 @@
...
+    kurl:
+      installerVersion: ""
+    kubernetes:
+      version: 1.27.x
+    flannel:
+      version: latest
+    ekco:
+      version: latest
+      enableInternalLoadBalancer: true
+- name: k8s124x_cis_benchmarks_checks
```
emosbaugh, line 2487: i like this one in the deploy

PR #4568, opened 2023-05-30
Fixes stuck scenarios to delete the Pods to allow them be re-created using Weave

camilamacedo86:

Example"

Screenshot 2023-05-30 at 23 09 21

Fixes # [sc-77764]

## Steps to reproduce
Create an GCP VM with ubuntu-20.04
- curl https://kurl.sh/f070e69 | sudo bash -s ha
- join two additional master nodes
- update with curl https://staging.kurl.sh/f0a3e1c | sudo bash -s ha

```release-note
Fixes scenario where Pods deletion get stuck and in the migration from Weave to Flannel for weave versions equals or upper than `0.21.5`. 
```

### DIFF
--- a/addons/flannel/0.21.5/install.sh
+++ b/addons/flannel/0.21.5/install.sh
@@ -239,23 +239,72 @@ function weave_to_flannel() {
     flannel_ready_spinner
 
     logStep "Restarting CSI pods"
-    kubectl -n longhorn-system delete pods --all || true
-    kubectl -n rook-ceph delete pods --all || true
-    kubectl -n openebs delete pods --all || true
+    log "this may take several minutes"
+    # We will try delete the Pods without force it
+    # However, in some cases such as where it has many addons it seems
+    # getting stuck and we force the pods deletion
+    echo "this may take several minutes"
 
+    # List of namespaces
+    namespaces=("longhorn-system" "rook-ceph" "openebs")
+
+    for ns in "${namespaces[@]}"; do
+        # Delete pods with a grace period and timeout, and ignore errors
+        timeout 80 kubectl -n ${ns} delete pods --all --grace-period=60 || logWarn "Timeout reached while trying to delete pods in namespace: ${ns}"
+
+        # Use the spinner_until function to wait until all pods are deleted
+        if ! spinner_until 60 pods_remain ${ns}; then
+            # If pods remain after the timeout, force delete them
+            logWarn "Force deleting remaining pods in namespace: ${ns}"
+            kubectl -n ${ns} delete pods --all --grace-period=0 --force || true
+        fi
+    done
+
+    echo "waiting 60 seconds before continue"
     sleep 60
+
     logStep "Restarting all other pods"
     echo "this may take several minutes"
     local ns=
-    for ns in $(kubectl get ns -o name | grep -Ev '(kube-system|longhorn-system|rook-ceph|openebs|kube-flannel)' | cut -f2 -d'/'); do
-        kubectl delete pods -n "$ns" --all --grace-period=200
+
+    # Get the list of namespaces
+    namespaces=$(kubectl get ns -o name | grep -Ev '(kube-system|longhorn-system|rook-ceph|openebs|kube-flannel)' | cut -f2 -d'/')
+
+    for ns in ${namespaces}; do
+        # Delete pods with a grace period and timeout, and ignore errors
+        timeout 80 kubectl delete pods -n "${ns}" --all --grace-period=60 || logWarn "Timeout reached while trying to delete pods in namespace: ${ns}"
+
+        # Use the spinner_until function to wait until all pods are deleted
+        if ! spinner_until 60 pods_remain "${ns}"; then
+            # If pods remain after the timeout, force delete them
+            logWarn "Force deleting remaining pods in namespace: ${ns}"
+            kubectl -n "${ns}" delete pods --all --grace-period=0 --force || true
+        fi
     done
+
+    echo "waiting 60 seconds before continue"
     sleep 60
+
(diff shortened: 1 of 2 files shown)


emosbaugh:
i would like to get a better understanding of why the pods are stuck terminating. the fact that they are stuck in my opinion indicates a bigger issue. in my case when i was testing this migration the kubelet systemd process was stuck in a terminating state and would not come back up. this would not have been resolved by adding a grace period




[3:21 PM](https://replicated.slack.com/archives/D048WU1EV52/p1685485286496839)
this was nothing that adding a grace period would help resolve

camilamacedo86:
Hi @emosbaugh,

> i would like to get a better understanding of why the pods are stuck terminating. the fact that they are stuck in my opinion indicates a bigger issue. in my case when i was testing this migration the kubelet systemd process was stuck in a terminating state and would not come back up. this would not have been resolved by adding a grace period

The Pod is not allowed be terminate when the process running in the container is not terminate.
In this case we delete Weave and we must to delete the Pods so that can be re-created to use Flannel.

The reason for the POD get stuck is:

> network is not ready: container runtime network not ready: NetworkReady=false reason:NetworkPluginNotReady message:Network plugin returns error: cni plugin not initialized
  Warning  FailedKillPod    4m30s (x212 over 49m)  kubelet          error killing pod: failed to "KillPodSandbox" for "f34a2c44-4569-4879-9200-3a13da6e213c" with KillPodSandboxError: "rpc error: code = Unknown desc = failed to destroy network for sandbox \"8c0ca335eb3d5876bd82b65109ee36ddc2cbf5e6dcb4e95039f2c6de60e50dd0\": cni plugin not initialized"

Which is well expected since we deleted Weave.

addons/flannel/0.21.5/install.sh
```diff
@@ -239,25 +239,74 @@ function weave_to_flannel() {
...
 
     logStep "Restarting CSI pods"
-    kubectl -n longhorn-system delete pods --all || true
-    kubectl -n rook-ceph delete pods --all || true
-    kubectl -n openebs delete pods --all || true
+    log "this may take several minutes"
+    # We will try delete the Pods without force it
+    # However, in some cases such as where it has many addons it seems
+    # getting stuck and we force the pods deletion
+    echo "this may take several minutes"
```
emosbaugh, line 246: this is the same as above

```suggestion
```

addons/flannel/template/base/install.sh
```diff
@@ -239,25 +239,74 @@ function weave_to_flannel() {
...
 
     logStep "Restarting CSI pods"
-    kubectl -n longhorn-system delete pods --all || true
-    kubectl -n rook-ceph delete pods --all || true
-    kubectl -n openebs delete pods --all || true
+    log "this may take several minutes"
+    # We will try delete the Pods without force it
+    # However, in some cases such as where it has many addons it seems
+    # getting stuck and we force the pods deletion
+    echo "this may take several minutes"
```
emosbaugh, line 246:
```suggestion
```

addons/flannel/0.21.5/install.sh
```diff
@@ -239,25 +239,74 @@ function weave_to_flannel() {
...
+
     log "Awaiting up to 5 minutes to check Flannel Pod(s) are Running"
     if ! spinner_until 300 check_for_running_pods "kube-flannel"; then
         logWarn "Flannel has unhealthy Pod(s)"
     fi
     logSuccess "Migrated from Weave to Flannel"
 }
 
+# Function to check if pods are still present in a namespace
+function pods_remain() {
```
emosbaugh, line 296: it is our practice to prefix functions in addons and files with the name of the file to avoid collisions

```suggestion
function flannel_pods_remain() {
```

addons/flannel/0.21.5/install.sh
```diff
@@ -239,25 +239,74 @@ function weave_to_flannel() {
...
+
+    # Get the list of namespaces
+    namespaces=$(kubectl get ns -o name | grep -Ev '(kube-system|longhorn-system|rook-ceph|openebs|kube-flannel)' | cut -f2 -d'/')
+
+    for ns in ${namespaces}; do
+        # Delete pods with a grace period and timeout, and ignore errors
+        timeout 80 kubectl delete pods -n "${ns}" --all --grace-period=60 || logWarn "Timeout reached while trying to delete pods in namespace: ${ns}"
+
+        # Use the spinner_until function to wait until all pods are deleted
+        if ! spinner_until 60 pods_remain "${ns}"; then
```
emosbaugh, line 278:
```suggestion
        if ! spinner_until 60 flannel_pods_remain "${ns}"; then
```

addons/flannel/0.21.5/install.sh
```diff
@@ -239,25 +239,74 @@ function weave_to_flannel() {
...
+
+    # List of namespaces
+    namespaces=("longhorn-system" "rook-ceph" "openebs")
+
+    for ns in "${namespaces[@]}"; do
+        # Delete pods with a grace period and timeout, and ignore errors
+        timeout 80 kubectl -n ${ns} delete pods --all --grace-period=60 || logWarn "Timeout reached while trying to delete pods in namespace: ${ns}"
+
+        # Use the spinner_until function to wait until all pods are deleted
+        if ! spinner_until 60 pods_remain ${ns}; then
```
emosbaugh, line 256:
```suggestion
        if ! spinner_until 60 flannel_pods_remain ${ns}; then
```

emosbaugh (changes requested):

addons/flannel/0.21.5/install.sh
camilamacedo86: True good catcher.

emosbaugh (approved):

PR #4582, opened 2023-06-01
revert changes from PR 4568

camilamacedo86:
## Why this PR is required

In the PR https://github.com/replicatedhq/kURL/pull/4568/files we added with the logic to move us forward by using --force and --gracePeriod options. However, it is dangerous usage and in some scenarios is resulting in kubelet unresponsive.

Motivated by # [sc-77764]

### DIFF
--- a/addons/flannel/0.21.5/install.sh
+++ b/addons/flannel/0.21.5/install.sh
@@ -233,95 +233,25 @@ function weave_to_flannel() {
     restart_systemd_and_wait kubelet
 
     logStep "Restarting pods in kube-system"
-    log "this may take several minutes"
-    local ns=
-    namespaces=("kube-system" "kube-flannel")
-
-    for ns in "${namespaces[@]}"; do
-        log "Waiting up to 5 minutes to delete all pods from ${ns}"
-        if ! spinner_until 300 kubectl -n ${ns} delete pods --all --ignore-not-found; then
-            logWarn "Timeout reached while trying to delete pods in namespace: ${ns}"
-        fi
-    done
-
-    for ns in "${namespaces[@]}"; do
-        log "Waiting up to 2 minutes to verify if pods will be terminated"
-        if ! spinner_until 120 flannel_pods_remain ${ns}; then
-            logWarn "The namespace ${ns} has pods which were not terminated"
-        fi
-    done
-
+    kubectl -n kube-system delete pods --all --grace-period=200
+    kubectl -n kube-flannel delete pods --all --grace-period=200
     flannel_ready_spinner
 
     logStep "Restarting CSI pods"
-    log "this may take several minutes"
-    local ns=
-    namespaces=("longhorn-system" "rook-ceph" "openebs")
+    kubectl -n longhorn-system delete pods --all || true
+    kubectl -n rook-ceph delete pods --all || true
+    kubectl -n openebs delete pods --all || true
 
-    for ns in "${namespaces[@]}"; do
-        log "Waiting up to 5 minutes to delete all pods from ${ns}"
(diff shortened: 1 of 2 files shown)


addons/flannel/0.21.5/install.sh
```diff
@@ -233,29 +233,42 @@ function weave_to_flannel() {
...
     logStep "Restarting pods in kube-system"
-    kubectl -n kube-system delete pods --all --grace-period=200
-    kubectl -n kube-flannel delete pods --all --grace-period=200
+    log "this may take several minutes"
+    local ns=
+    namespaces=("kube-system" "kube-flannel")
+
+    for ns in "${namespaces[@]}"; do
+        log "Waiting up to 5 minutes to delete all pods from ${ns}"
+        if ! spinner_until 300 kubectl -n ${ns} delete pods --all --ignore-not-found; then
```
emosbaugh, line 242: it is required that these pods be deleted to get new pod sandboxes. i do not think a timeout with a warning and proceeding is a good idea.

emosbaugh (changes requested):

addons/flannel/0.21.5/install.sh
```diff
@@ -233,29 +233,42 @@ function weave_to_flannel() {
...
-
-    # List of namespaces
+    local ns=
     namespaces=("longhorn-system" "rook-ceph" "openebs")
 
     for ns in "${namespaces[@]}"; do
-        # Delete pods with a grace period and timeout, and ignore errors
-        timeout 80 kubectl -n ${ns} delete pods --all --grace-period=60 || logWarn "Timeout reached while trying to delete pods in namespace: ${ns}"
+        log "Waiting up to 5 minutes to delete all pods from ${ns}"
+        if ! spinner_until 300 kubectl -n ${ns} delete pods --all --ignore-not-found; then
```
laverya, line 263: spinner_until is REALLY not the function you should be running here. This will call `kubectl -n ${ns} delete pods --all --ignore-not-found` once per second for 300 seconds.

addons/flannel/0.21.5/install.sh
```diff
@@ -269,15 +282,18 @@ function weave_to_flannel() {
...
 
-        # Use the spinner_until function to wait until all pods are deleted
-        if ! spinner_until 60 flannel_pods_remain "${ns}"; then
+    for ns in "${namespaces[@]}"; do
+        log "Waiting up to 2 minutes to verify if pods will be terminated"
+        if ! spinner_until 120 flannel_pods_remain ${ns}; then
             # If pods remain after the timeout, force delete them
-            logWarn "Force deleting remaining pods in namespace: ${ns}"
-            kubectl -n "${ns}" delete pods --all --grace-period=0 --force || true
+            logWarn "The namespace ${ns} has pods which were not terminated"
```
laverya, line 296: if you're going to change the logic, you can't just leave the comment describing the logic the same.

I also disagree with removing the force deletion here.

laverya (commented):
The overall direction this is moving in assumes that failing to delete pods is not a fatal error when migrating CNIs. *Are you positive?*. Because my intuition would suggest otherwise.

addons/flannel/0.21.5/install.sh
camilamacedo86: @laverya the force was added in the PR https://github.com/replicatedhq/kURL/pull/4568/files
However, besides it sort out some scenarios I found others ones where the cluster will end up broken
We cannot use force .

laverya (approved):

camilamacedo86:
As we discussed in the slack I am concerned with the force option usage here and it might let the cluster inresponsive.
I want to revert and then, we can discuss the best approach. So, I changed this PR to be only a revert of the code.

Now that we have the Rook tests added we will be able to easily ensure the changes.

PR #4588, opened 2023-06-05
Automated Prometheus version update 0.65.1-46.6.0

replicated-ci-kurl:
Automated changes by the [cron-prometheus-update](https://github.com/replicatedhq/kURL/blob/main/.github/workflows/update-prometheus.yaml) GitHub action

```release-note
Adds [Prometheus add-on](https://kurl.sh/docs/add-ons/prometheus) version 0.65.1-46.6.0.
```


camilamacedo86 (approved):
We can see that it passed in manly all test.
The failure test has been failing for a long period and has not been a reason for we move forward since it is a upgrade from the oldest version to the latest and just fail for an specific target OS. 
Screenshot 2023-06-06 at 09 12 40

So, moving forward with this one.

PR #4597, opened 2023-06-08
Automated Prometheus version update 0.65.2-46.8.0

replicated-ci-kurl:
Automated changes by the [cron-prometheus-update](https://github.com/replicatedhq/kURL/blob/main/.github/workflows/update-prometheus.yaml) GitHub action

```release-note
Adds [Prometheus add-on](https://kurl.sh/docs/add-ons/prometheus) version 0.65.2-46.8.0.
```


camilamacedo86 (approved):
The following failure test for Amazon linux is a test that has been failing in all versions and it is not a blocker. The other one is a flake.

Screenshot 2023-06-10 at 08 27 16

So it is good to fly.

PR #4598, opened 2023-06-08
feat(ekco): Generate storage migration API auth token

rrpolanco:

When installing EKCO, generate a storage migration API auth token.

Fixes #

```release-note

```


camilamacedo86 (approved):
It seems fine for me.
I understand that we store the random value an a configMap
so that we can use it to trigger the migration after.

PR #4599, opened 2023-06-08
(Support Case) fix rook check to allow manual fix when is upgrading from only

camilamacedo86:

In certain edge cases, while migrating away from Rook, we may encounter issues.  Specifically, after we execute a pvmigrate operation to migrate the PVCs and to migrate the Object store, the system may transition to an unhealthy state. This problem appears to be connected to specific modules

```
[root@rook-ceph-operator-747c86774c-7v95s /]# ceph health detail 
HEALTH_ERR 2 mgr modules have failed
MGR_MODULE_ERROR 2 mgr modules have failed
    Module 'dashboard' has failed: error('No socket could be created',)
    Module 'prometheus' has failed: error('No socket could be created',) 
``` 

The proposed workaround ensures a smooth transition during the migration and upgrade processes, ultimately allowing for the successful deletion of Rook. To this end, this PR automates the resolution process by rectifying the Rook Ceph state and allowing the migration to proceed, given that Rook will be removed in the end. **It's important to note that this automated fix is only applied during the checks performed when we are in the process of migrating away from Rook and when Rook's removal is the intended outcome.**

Because of this we are duplicating the check and ensuring that just this process will use it. 

Fixes # [sc-79289]

This automated fix has been implemented specifically to mitigate unnecessary support calls in scenarios where they aren't required. It's worth noting that we perform a thorough status recheck after applying the workaround, prior to continuing with the process.

```release-note
Fixes Rook Ceph status prometheus and modules when migrating from Rook and we check the status is unhealthy with the errors `Module 'dashboard' has failed` and ` Module 'prometheus' has failed` to allow the migration continue.
```

### DIFF
--- a/scripts/common/rook.sh
+++ b/scripts/common/rook.sh
@@ -385,13 +385,13 @@ function rook_operator_ready() {
 # Note this method is a duplication of rook_is_healthy_to_upgrade which now is called in the migration process
 # ONLY when we are moving from Rook. We should not try to fix it in other circumstances
 function rook_is_healthy_to_migrate_from() {
-    log "Awaiting 2 minutes to check Rook Ceph Pod(s) are Running"
+    log "Awaiting up to 5 minutes to check Rook Ceph Pod(s) are Running"
     if ! spinner_until 300 check_for_running_pods "rook-ceph"; then
         logFail "Rook Ceph has unhealthy Pod(s)"
         return 1
     fi
 
-    log "Awaiting Rook Ceph health ..."
+    log "Awaiting up to 10 minutes to check that Rook Ceph is health"
     if ! $DIR/bin/kurl rook wait-for-health 600 ; then
         logWarn "Rook Ceph is unhealthy"
 
@@ -446,13 +446,13 @@ function rook_is_healthy_to_migrate_from() {
 }
 
 function rook_is_healthy_to_upgrade() {
-    log "Awaiting 2 minutes to check Rook Ceph Pod(s) are Running"
-    if ! spinner_until 120 check_for_running_pods "rook-ceph"; then
+    log "Awaiting up to 5 minutes to check Rook Ceph Pod(s) are Running"
+    if ! spinner_until 300 check_for_running_pods "rook-ceph"; then
         logFail "Rook Ceph has unhealthy Pod(s)"
         return 1
     fi
 
-    log "Awaiting Rook Ceph health ..."
+    log "Awaiting up to 10 minutes to check that Rook Ceph is health"
     if ! $DIR/bin/kurl rook wait-for-health 600 ; then
         kubectl -n rook-ceph exec deploy/rook-ceph-tools -- ceph status
         logFail "Rook Ceph is unhealthy"
--- a/scripts/common/rook.sh
+++ b/scripts/common/rook.sh
@@ -368,6 +368,83 @@ function rook_operator_ready() {
     return 0
 }
 
+# In certain edge cases, while migrating away from Rook, we may encounter issues.
+# Specifically, after we execute a pvmigrate operation to migrate the PVCs and to migrate the Object store, the system
+# may transition to an unhealthy state. This problem appears to be connected to specific modules
+# [root@rook-ceph-operator-747c86774c-7v95s /]# ceph health detail
+# HEALTH_ERR 2 mgr modules have failed
+# MGR_MODULE_ERROR 2 mgr modules have failed
+#     Module 'dashboard' has failed: error('No socket could be created',)
+#     Module 'prometheus' has failed: error('No socket could be created',)
+# The proposed workaround ensures a smooth transition during the migration and upgrade processes, ultimately allowing
+# for the successful deletion of Rook. To this end, this PR automates the resolution process by rectifying the Rook Ceph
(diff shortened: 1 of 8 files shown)


scripts/common/rook.sh
```diff
@@ -368,6 +368,67 @@ function rook_operator_ready() {
     return 0
 }
 
+function rook_is_healthy_to_migrate_from() {
```
camilamacedo86, line 371: See that this check is a duplication of rook_is_healthy_to_upgrade which now is called in the migration process ONLY when we are moving from Rook

scripts/common/rook.sh
```diff
@@ -368,6 +368,83 @@ function rook_operator_ready() {
...
+
+        echo ""
+        echo output
+        echo ""
+
+        if [[ $output == *"Module 'dashboard' has failed"* ]] || [[ $output == *"Module 'prometheus'"* ]]; then
+            echo "Disable modules to try fix status"
+            kubectl -n rook-ceph exec deployment/rook-ceph-tools -- ceph mgr module disable prometheus
+            kubectl -n rook-ceph exec deployment/rook-ceph-tools -- ceph mgr module disable dashboard
+        fi
```
camilamacedo86, line 408: It is the DELTA ONLY

scripts/common/rook.sh
```diff
@@ -368,6 +368,83 @@ function rook_operator_ready() {
...
+        logWarn "Rook Ceph is unhealthy"
+
+        output=$(kubectl -n rook-ceph exec deploy/rook-ceph-tools -- ceph status)
+
+        echo ""
+        echo output
+        echo ""
+
+        if [[ $output == *"Module 'dashboard' has failed"* ]] || [[ $output == *"Module 'prometheus'"* ]]; then
+            echo "Disable modules to try fix status"
```
rrpolanco, line 405:
```suggestion
            echo "Disabling Ceph manager modules in order to get Ceph healthy again."
```

rrpolanco (approved):

scripts/common/rook.sh
```diff
@@ -368,6 +368,83 @@ function rook_operator_ready() {
...
+    if ! $DIR/bin/kurl rook wait-for-health 600 ; then
+        logWarn "Rook Ceph is unhealthy"
+
+        output=$(kubectl -n rook-ceph exec deploy/rook-ceph-tools -- ceph status)
+
+        echo ""
+        echo output
+        echo ""
+
+        if [[ $output == *"Module 'dashboard' has failed"* ]] || [[ $output == *"Module 'prometheus'"* ]]; then
```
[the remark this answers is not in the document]
camilamacedo86: @rrpolanco I found one nit.

camilamacedo86 (commented):
@rrpolanco I found a nit: https://github.com/replicatedhq/kURL/pull/4599/commits/c236023fe6cd4ce15a4b7b5c50ca93b4f3dd12c9

Could you please re-check again?

rrpolanco (approved):

scripts/common/rook.sh
```diff
@@ -368,6 +368,83 @@ function rook_operator_ready() {
...
+# for the successful deletion of Rook. To this end, this PR automates the resolution process by rectifying the Rook Ceph
+# state and allowing the migration to proceed, given that Rook will be removed in the end. It's important to note that
+# this automated fix is only applied during the checks performed when we are in the process of migrating away from Rook
+# and when Rook's removal is the intended outcome.
+#
+# Note this method is a duplication of rook_is_healthy_to_upgrade which now is called in the migration process
+# ONLY when we are moving from Rook. We should not try to fix it in other circumstances
+function rook_is_healthy_to_migrate_from() {
+    log "Awaiting 2 minutes to check Rook Ceph Pod(s) are Running"
+    if ! spinner_until 300 check_for_running_pods "rook-ceph"; then
```
banjoh, line 389: Should this be `300` or `120`? Either update the log message or the timeout

scripts/common/rook.sh
```diff
@@ -368,6 +368,83 @@ function rook_operator_ready() {
...
+        # More info: https://github.com/rook/rook/pull/6610
+        if [ "$(echo "${ceph_versions_found}" | wc -l)" == "2" ] && [ "$(echo "${ceph_versions_found}" | grep "0.0.0-0")" ]; then
+            log "Found two ceph versions but one of them is 0.0.0-0 which will be ignored"
+            echo "${ceph_versions_found}"
+        else
+            logFail "Multiple Ceph versions detected"
+            logFail "${ceph_versions_found}"
+            return 1
+        fi
+    fi
```
banjoh, line 444: These entire code looks similar to code in `rook_is_healthy_to_upgrade()`. Wouldn't it be better to DRY it by extracting in into some other function

camilamacedo86: I do not think so because that could be risk we introduce an issue.
So, think in the risk vs maintainability I think would be better for now we just keep duplicated and explained why.

scripts/common/rook.sh
camilamacedo86: 300 we have an issue in the text. Good catcher

camilamacedo86: Fixed

ricardomaraschini (approved):

ricardomaraschini (approved):

PR #4602, opened 2023-06-09
bug: make weave_to_flannel idempotent

ricardomaraschini:

This PR contains some improvements on the `weave_to_flannel` migration function. We want to guarantee that if something goes wrong with the migration we will try it once more when the script runs again and that it will eventually finish its work correctly.

#### Notes

Depending where the Weave to Flannel migration fails we may end up with a cluster in a unhealthy state (no CNI running thus some pods may be failing). As the cluster may not be in a healthy state after the failure the user may need to provide the `-s host-preflight-ignore` flag to the script during a second attempt.

The second attempt to migrate from weave to flannel is dispatched directly from the flannel pre_init phase, this is done because we can't guarantee that all addons preflights will succeed in case of a cluster running without CNI. 

#### About testing

In order to test this code I had to randomly `ctrl+c` the installer during different stages of the weave to flannel migration and then re-run the install script passin gin `-s host-preflight-ignore` flag. I have tested this in both single and multi nodes scenarios.

### DIFF
--- a/addons/flannel/0.21.5/install.sh
+++ b/addons/flannel/0.21.5/install.sh
@@ -34,15 +34,35 @@ function flannel_pre_init() {
 
     flannel_init_pod_subnet
 
-    if flannel_weave_conflict; then
-        logWarn "The migration from Weave to Flannel will require whole-cluster downtime."
-        logWarn "Would you like to continue?"
+    if ! flannel_weave_conflict; then
+        return 0
+    fi
+
+    logWarn "The migration from Weave to Flannel will require whole-cluster downtime."
+    logWarn "Would you like to continue?"
+    if ! confirmY ; then
+        bail "Not migrating from Weave to Flannel"
+    fi
+
+    # if the migration from weave has failed before we may have a cluster with a
+    # non working network, in this scenario we can't simply move forward because
+    # other addons may fail during their pre_init phase. we need to jump from here
+    # directly into the weave to flannel migration code. subsequent call to install
+    # flannel again will be, essentially, a noop.
+    if flannel_weave_migration_has_failed; then
+        logWarn "Previous attempt to migrate from Weave to Flannel has failed."
+        logWarn "Attempting to migrate again, do you wish to continue?"
         if ! confirmY ; then
             bail "Not migrating from Weave to Flannel"
         fi
-        flannel_check_nodes_connectivity
-        flannel_check_rook_ceph_status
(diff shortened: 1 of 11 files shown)


addons/flannel/0.21.5/install.sh
```diff
@@ -295,6 +317,10 @@ function flannel_scale_up_rook() {
 function weave_to_flannel() {
     local dst="$DIR/kustomize/flannel"
 
+    # start by creating a config map indicating that this migration has been started. if we fail
+    # midair this ensure this function will be ran once more.
+    kubectl create configmap -n kube-system migrating-from-weave --from-literal=started="true" --dry-run=client -o yaml | kubectl apply -f -
```
camilamacedo86, line 322: Currently, we have already `kurl-migration-from-rook` see https://github.com/replicatedhq/kURL/blob/a99174144c2aa8a26717b48709ea975cddf44456/scripts/common/rook.sh#L419

Could we follow the same standard and keep the configmap in the kurl namespace since it is a kurl thing?

scripts/common/preflights.sh
```diff
@@ -30,10 +30,15 @@ function init_preflights() {
 # if kurl pods like ekco not be running then we should bail
 function bail_if_kurl_pods_are_unhealthy() {
     if commandExists kubectl; then
-      log "Awaiting 2 minutes to check kURL Pod(s) are Running"
-      if ! spinner_until 120 check_for_running_pods kurl; then
-          bail "Kurl has unhealthy Pod(s). Check the namespace kurl. Restarting the pod may fix the issue."
-      fi
+        # if host preflight are being ignored we just move forward.
```
camilamacedo86, line 33: Should not it be in another PR so that we can either generate the release notes accordingly to know what is fixed or not in each version which makes easier when we need to gathering this info in the support

> Fixes pods check to allow ignore it failures when HOST_PREFLIGHT_IGNORE flag is used

ricardomaraschini: For sake of simplicity I am keeping this here.

addons/flannel/0.21.5/install.sh
ricardomaraschini: Done.

rrpolanco (approved):
LGTM. We should also add a story to fix the test **weave to flannel, docker to containerd, single node** on Rocky Linux.

camilamacedo86 (approved):

PR #4604, opened 2023-06-09
add new test to the daily - weave to flannel

camilamacedo86:
I noticed that we have not in the daily a test covering weave to flannel
So this new test covers single node and with both addons OpenEBS and Rook

See: https://testgrid.kurl.sh/run/CAMILA-TEST-WEAVE-FLANNEL-ALL-2


camilamacedo86:
@emosbaugh I would like to add this one to the daily so that we can start to cover this scenario.
Today we are not covering it in the daily. WDYT?

emosbaugh (approved):

PR #4609, opened 2023-06-12
Change ekco svc to node port type

rrpolanco:

Change the `ekco` service to be of type `NodePort` so we can access it outside the cluster, namely during the join process.
Also, expose global vars `EKCO_AUTH_TOKEN` and `EKCO_ADDRESS` to be used as argument to `joins.sh`:
```bash
cat join.sh | sudo bash -s kubernetes-master-address=10.128.0.23:6443 kubeadm-token=ozlhkb.gzcs2bptis6itlxv kubeadm-token-ca-hash=sha256:e4af661f45c39c5f7a70e57718e23b745d466d947f0e1e1d749c4dd93a97076a kubernetes-version=1.27.2 ekco-address=10.128.0.23:31880 ekco-auth-token=8kt0lkjIQwoPsTdQozrxwrgy9VDIeqaqHSgUeTMCZoYMk6T34mhu2df9oLNrjfXy docker-registry-ip=10.96.0.118 additional-no-proxy-addresses=10.96.0.0/22,10.32.0.0/20 primary-host=10.128.0.23
```

Fixes #

```release-note

```

### DIFF
--- a/addons/ekco/0.27.1/service.tmpl.yaml
+++ b/addons/ekco/0.27.1/service.tmpl.yaml
@@ -3,8 +3,10 @@ kind: Service
 metadata:
   name: ekc-operator
 spec:
+  type: NodePort
   selector:
     app: ekc-operator
   ports:
     - protocol: TCP
       port: 8080
+      nodePort: $EKCO_NODE_PORT
--- a/addons/ekco/0.27.1/configmap.tmpl.yaml
+++ b/addons/ekco/0.27.1/configmap.tmpl.yaml
@@ -49,4 +49,4 @@ data:
     enable_ha_kotsadm: $EKCO_MAINTAIN_KOTSADM
     rook_minimum_node_count: ${ROOK_MINIMUM_NODE_COUNT:-0}
     rook_storage_class: "${STORAGE_CLASS:-distributed}"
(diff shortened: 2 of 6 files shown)


addons/ekco/0.27.1/service.tmpl.yaml
```diff
@@ -3,8 +3,10 @@ kind: Service
...
 spec:
+  type: NodePort
   selector:
     app: ekc-operator
   ports:
     - protocol: TCP
-      port: 8080
\ No newline at end of file
+      port: 8080
+      nodePort: $EKCO_NODE_PORT
```
laverya, line 12: is there a reason to specify the nodeport, instead of getting it after the fact?

laverya (approved):

ricardomaraschini (approved):

addons/ekco/0.27.1/service.tmpl.yaml
rrpolanco: if `nodePort` not explicitly specified, the control planes will assign one for you in this range: `30000-32767`. I'd rather we assign a specific port to the EKCO server for consistency and ease of discoveribility.

PR #4629, opened 2023-06-20
fix daily testgrid check to verify weave to flannel

camilamacedo86:

The tests are failing because we are using `__testver__` instead of latest.
The `__testver__` is only valid/useful to the tests defined for each addon.


ricardomaraschini:
I am not sure I follow. Aren't these tests going to run only when there is a new version of flannel? Are they also run daily?

ricardomaraschini (approved):

PR #4631, opened 2023-06-20
(testgrid) Add OS rock linux 8.8

camilamacedo86:

Add tests for 8.8 since we have not

Fixes # [sc-78497]

```release-note

```

### DIFF
--- a/testgrid/specs/os-latest.yaml
+++ b/testgrid/specs/os-latest.yaml
@@ -29,6 +29,10 @@
     curl -L -o centos2ol.sh https://raw.githubusercontent.com/oracle/centos2ol/main/centos2ol.sh
     chmod +x centos2ol.sh
     bash centos2ol.sh -r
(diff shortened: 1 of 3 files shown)


testgrid/specs/os-latest.yaml
```diff
@@ -29,6 +29,10 @@
     curl -L -o centos2ol.sh https://raw.githubusercontent.com/oracle/centos2ol/main/centos2ol.sh
     chmod +x centos2ol.sh
     bash centos2ol.sh -r
+- id: rocky-88
```
emosbaugh, line 32: should you remove centos-84 from os-latest?

laverya: Agreed I would love this if we removed centos 8.4

camilamacedo86:
We decided to not move with this one in the standup

PR #4632, opened 2023-06-20
use patches _correctly_

laverya:

Fixes #

```release-note

```

### DIFF
--- a/addons/flannel/template/base/yaml/kustomization.yaml
+++ b/addons/flannel/template/base/yaml/kustomization.yaml
@@ -2,5 +2,5 @@ resources:
   - kube-flannel.yml
   - troubleshoot.yaml
 
-patchesStrategicMerge:
-  - kube-flannel-cfg.patch.yaml
+patches:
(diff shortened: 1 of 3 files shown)


addons/flannel/template/base/yaml/kustomization.yaml
```diff
@@ -2,5 +2,5 @@ resources:
   - kube-flannel.yml
   - troubleshoot.yaml
 
-patches:
+patchesStrategicMerge:
```
rrpolanco, line 5: This won't work with kubernetes >= 1.27.x due to Kustomize v5 removing support for patchesStrategicMerge.
Instead you can do:
```yaml
patches:
- path: kube-flannel-cfg.patch.yaml
```

laverya: hah ok

ricardomaraschini (approved):

rrpolanco (approved):

PR #4635, opened 2023-06-21
Add preflight check to ensure storage or object store when Registry add-on is set

camilamacedo86:

Because we cannot allow have a Registry without one of those options

Fixes # [sc-78397]

```release-note
Adds preflight check when Registry add-on is set to ensure that an object store or storage provision is also be in place. Otherwise, fails the install to prevent issues scenarios.
```


scripts/common/preflights.sh
```diff
@@ -863,6 +864,17 @@ function bail_when_no_object_store_and_s3_enabled() {
     fi
 }
 
+# bail_when_no_object_store_or_storage_and_has_registry will bail if no object store or any storage addon be select
+# for an install with registry
+function bail_if_no_object_store_or_storage_and_has_registry() {
+    if [ -n "$REGISTRY_VERSION" ]; then
+        if [ -z "$MINIO_VERSION" ] && [ -z "$ROOK_VERSION" ] && [ -z "$OPENEBS_VERSION" ] && [ -z "$LONGHORN_VERSION" ]; then
+             logFail "Registry add-on requires an object store or a storage provision"
```
ricardomaraschini, line 872:
```suggestion
             logFail "Registry add-on requires an object store or a storage provisioner"
```

scripts/common/preflights.sh
```diff
@@ -863,6 +864,17 @@ function bail_when_no_object_store_and_s3_enabled() {
...
+             bail "Please ensure that your installer also provides an object store or a storage storage provision with either the MinIO or Rook or OpenEBS add-on."
```
ricardomaraschini, line 873:
```suggestion
             bail "Please ensure that your installer also provides an object store or a storage provisioner with either the MinIO or Rook or OpenEBS add-on."
```

PR #4656, opened 2023-06-27
feat(ekco): Disable Rook storage node reconciliation in EKCO if Rook min. node count is set

rrpolanco:
When Rook and OpenEBS are installed and `Rook.minimumNodeCount` is set, disable ekco from [reconciling the Ceph cluster pool replication levels](https://github.com/replicatedhq/ekco/blob/main/pkg/ekcoops/operator.go#L155-L179). When `Rook.minimumNodeCount` cluster nodes is met, the EKCO operator will create the CephCluster via a Helm chart apply. This change will prevent EKCO from changing the default CephCluster created.

Fixes #

```release-note

```


rrpolanco:
Will add this fencing logic to EKCO.

PR #4665, opened 2023-06-30
feat(rook): Don't deploy Ceph cluster if minimumNodeCount > 2 (instead of > 1)

rrpolanco:
When eligible for automatic storage scaling (local to distributed), the Rook add-on should not create a Ceph cluster if `minimumNodeCount` > 2. In such case, EKCO will create the Ceph Cluster.

Fixes #

```release-note

```


rrpolanco:
Not needed.

PR #4677, opened 2023-07-06
improve no-subnet-found messaging

laverya:

remove panics from subnet discovery binary

report subnet ranges that were searched

fix 'how to set range manually' messaging, and link to docs

Fixes #

```release-note

```


kurl_util/cmd/subnet/main.go
rrpolanco, line 1: looks like the `errors` package is no longer needed.

laverya: not quite, but updated now

rrpolanco (approved):

PR #4684, opened 2023-07-10
do not require registry IP when joining non-docker nodes

laverya:

This will allow upgrading k8s and adding docker registry at the same time on clusters using containerd, as the IP was only required for docker clusters.

Fixes #

```release-note

```


rrpolanco (approved):

laverya:
I'm going to wait for testgrid before merging this

PR #4689, opened 2023-07-14
feat: Add kurl in-cluster support bundle spec

banjoh:

This PR adds a change to install an in-cluster support bundle during installation and upgrade of kURL clusters. The support bundle spec is installed in `kurl-supportbundle-spec` secret in the `kurl` namespace. It will be discovered by the `support-bundle` binary whenever the `--load-cluster-specs` option is used

The purpose of having this support bundle spec here is to ensure that we always (assuming `--load-cluster-specs` is passed) get kURL specific collectors and analysers run even if a vendor does not have them specified in their specs, or even if they do not exist in specs provided by Replicated.

Fixes [79290](https://app.shortcut.com/replicated/story/79290/collect-kurl-installer-logs-from-embedded-clusters-in-the-support-bundle)

```release-note

```

### DIFF
--- a/scripts/common/kurl.sh
+++ b/scripts/common/kurl.sh
@@ -22,6 +22,31 @@ function kurl_set_current_version() {
     kubectl patch configmaps -n kurl kurl-current-config --type merge -p "{\"data\":{\"kurl-version\":\"${KURL_VERSION}\"}}"
 }
 
+function kurl_install_support_bundle_configmap() {
+    cat <<EOF | kubectl apply -n kurl kurl-current-config -f -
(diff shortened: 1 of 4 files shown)


scripts/upgrade.sh
```diff
@@ -131,6 +131,7 @@ function main() {
     journald_persistent
     configure_proxy
     configure_no_proxy
+    kurl_install_support_bundle_configmap
```
laverya, line 134: this should be removed - upgrade won't always have kubectl access, and is only called upon prompting from install.sh

PR #4692, opened 2023-07-17
do not check for rook block devices if minimumNodeCount >= 3

laverya:

Fixes #

```release-note

```

### DIFF
--- a/addons/rook/template/base/host-preflight.yaml
+++ b/addons/rook/template/base/host-preflight.yaml
@@ -15,7 +15,7 @@ spec:
   - blockDevices:
       includeUnmountedPartitions: true
       minimumAcceptableSize: 10737418240 # 1024 ^ 3 * 10, 10GiB
-      exclude: '{{kurl .IsUpgrade }}'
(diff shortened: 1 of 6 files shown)


addons/rook/template/base/host-preflight.yaml
rrpolanco, line 1: Why are we doing this? If the vendor has `.Installer.Spec.Rook.MinimumNodeCount` set to >= 3 then it would imply they intend to migrate to Rook/Ceph as the storage provider. If that is true then having the block device check will be applicable.

laverya: Because they might NOT intend to migrate yet! They might be installing single node, and then upgrading to rook later either by attaching block devices after the fact, or adding 3 new nodes with block devices.

ricardomaraschini (approved):

addons/rook/template/base/host-preflight.yaml
rrpolanco: Understand. However, it would be better if there was a way to exclude this particular block device preflight w/o manipulating `Rook.MinimumNodeCount`.

PR #4704, opened 2023-07-25
create the ceph mgr blockpool before making the ceph cluster

laverya:

When the ekco operator creates the cephcluster for automated storage migrations, it the `.mgr` pool is created with one replica. This should force it to use 3 replicas.

Fixes #

```release-note

```

### DIFF
--- a/addons/rook/1.11.8/mgr-blockpool.yaml
+++ b/addons/rook/1.11.8/mgr-blockpool.yaml
@@ -1,0 +1,22 @@
+apiVersion: ceph.rook.io/v1
+kind: CephBlockPool
+metadata:
+  # If the built-in Ceph pool used by the Ceph mgr needs to be configured with alternate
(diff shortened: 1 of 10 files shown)


addons/rook/1.11.8/mgr-blockpool.yaml
rrpolanco, line 1: can we add this to values.yaml in ekco?

Something like this?
```yaml
  cephBlockPools:
    - name: builtin-mgr
      spec:
        name: .mgr
        failureDomain: host
        replicated:
          size: 3
          requireSafeReplicaSize: true
        parameters:
          compression_mode: none
        mirroring:
          enabled: false
```

laverya: Maybe, but I know that this will work, and it was what was suggested here: https://github.com/rook/rook/issues/11368

PR #4729, opened 2023-08-04
migrate from longhorn to openebs, then upgrade k8s past 1.24

laverya:

This would install openebs, migrate from longhorn, delete longhorn, and only then upgrade k8s past what longhorn supported

Fixes #

```release-note
Fix an issue where storage could not be moved from Longhorn to OpenEBS at the same time as Kubernetes was upgraded to 1.25.x+
```


rrpolanco (dismissed):
lgtm

rrpolanco (approved):

PR #4744, opened 2023-08-10
Revert "Automated Containerd version update 1.6.22"

laverya:
Reverts replicatedhq/kURL#4720


laverya:
This version is not supported on Ubuntu 18.04. Until we can come up with a better story around how to handle that, we shouldn't release this

rrpolanco (approved):

PR #4914, opened 2023-10-26
chore: inform users that 15% of the root disk is reserved

ricardomaraschini:

When evaluating the amount of free disk space to migrate to OpenEBS we reserve 15% of the disk for the operating system (avoid Kubernetes triggering of its clean up unused images routine). We haven't been informing users of that causing confusion on why the available + used amounts did not add up to the full root disk size.

### DIFF
--- a/pkg/cluster/space/openebs_disk_space_validator.go
+++ b/pkg/cluster/space/openebs_disk_space_validator.go
@@ -56,7 +56,7 @@ func (o *OpenEBSDiskSpaceValidator) NodesWithoutSpace(ctx context.Context) ([]st
 
 		var reservedMsg string
 		if vol.RootVolume {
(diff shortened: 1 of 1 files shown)


pkg/cluster/space/openebs_disk_space_validator.go
```diff
@@ -54,13 +54,19 @@ func (o *OpenEBSDiskSpaceValidator) NodesWithoutSpace(ctx context.Context) ([]st
 			continue
 		}
 
+		var reservedMsg string
+		if vol.RootVolume {
+			reservedMsg = "(15% of the root disk is reserved for the OS)"
```
laverya, line 59: I would say instead that 15% is reserved to prevent diskpressure evictions

ricardomaraschini: Good. Done.

PR #4954, opened 2023-11-20
bug: validate if received data is an uuid

ricardomaraschini:

When validating nodes can talk to each other we send uuid from one point to another but we are not validating if the output matches the expected uuid. in some scenarios the pod may log some error and if we don't and on these cases the log is not an uuid and we should retur an error instead.

```release-note
Improve error reporting capabilities during weave to flannel migration.
```

### DIFF
--- a/pkg/cli/nodes_connectivity.go
+++ b/pkg/cli/nodes_connectivity.go
@@ -78,11 +78,11 @@ func newNetutilNodesConnectivity(_ CLI) *cobra.Command {
 		Example: usageExamples,
 		PreRunE: func(cmd *cobra.Command, args []string) error {
 			if opts.port == 0 {
-				return fmt.Errorf("--port flag is required.")
+				return fmt.Errorf("--port flag is required")
 			}
 			opts.proto = strings.ToUpper(opts.proto)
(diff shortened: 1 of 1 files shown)


pkg/cli/nodes_connectivity.go
```diff
@@ -256,7 +256,13 @@ func attachToListenersPods(ctx context.Context, opts nodeConnectivityOptions, po
 			for scanner.Scan() {
 				txt := scanner.Text()
 				opts.debugf("Pod %s log: %s\n", pod.Name, txt)
-				out <- logLine{message: scanner.Text()}
+				if _, err := uuid.Parse(txt); err != nil {
+					opts.debugf("Pod line is not an UUID\n")
+					out <- logLine{err: fmt.Errorf("invalid output found: %s", txt)}
+					continue
+				}
+				opts.debugf("Pod line is an UUID\n")
```
laverya, line 264:
```suggestion
					opts.debugf("Pod line is not a UUID\n")
					out <- logLine{err: fmt.Errorf("invalid output found: %s", txt)}
					continue
				}
				opts.debugf("Pod line is a UUID\n")
```

PR #4980, opened 2023-12-12
feat: increase goldpinger timeouts and add goldpinger troubleshoot spec

banjoh:

- Increases goldpinger addon timeouts. In slowish networks, pings tend to fail unnecessarily. The default ping timeout is `300ms`.
- Add a support bundle spec whenever goldpinger addon is added to a kurl spec.

Fixes [#sc-91987](https://app.shortcut.com/replicated/story/91987/analyzer-goldpinger-pingmap)

## Steps to reproduce

- Create a kURL cluster
- Run `support-bundle --load-cluster-specs` and notice goldpinger analysis in the results. This bit is dependant on https://github.com/replicatedhq/troubleshoot/pull/1398 changes being released

```release-note
NONE
```

NONE


addons/goldpinger/template/yaml/troubleshoot.yaml
```diff
@@ -0,0 +1,19 @@
+apiVersion: v1
```
banjoh, line 1: The testgrid tests are passing, but for some reason I cannot figure out how this `troubleshoot.yaml` spec is applied to the cluster when installing the add-on!

laverya: I can guarantee that it's not being applied if it's only here in the template directory

banjoh: The file is in `template/yaml`.

I have specific tests checking that the secret gets created and they are passing. It seems the secret gets applied, I just can't tell where that happens

laverya: No, I mean that the template directory is only used to generate the actual versioned addons - it's not included in the things being run on testgrid.

And `template/yaml` isn't a directory used by anything to the best of my knowledge - if you want this to be included in future versions it would need to be added to https://github.com/replicatedhq/kURL/tree/main/addons/goldpinger/template/base instead

laverya (approved):

laverya:
The latest version appears to work, but the previous ones are failing

laverya (changes requested):
A lot of versions appear to be failing: https://testgrid.kurl.sh/run/pr-4980-6e467f4-goldpinger-3.7.0-5.5.0-tests-2023-12-14T16:30:41Z

laverya (approved):

PR #4991, opened 2023-12-14
enable building new kubernetes

laverya:

This does two things:
1. it changes our k8s package build process to use the new k8s repos
2. it changes k8s packages < 1.24 to always copy from the previous kurl version instead of being built fresh, as they do not exist in the new k8s repos.

The new k8s repos are also minor version dependent, so there is code to pick the right repo based on the k8s version.

Fixes #

```release-note

```

### DIFF
--- a/bin/upload-dist-staging.sh
+++ b/bin/upload-dist-staging.sh
@@ -81,6 +81,7 @@ function is_old_kubernetes() {
     if echo "${package}" | grep -q "kubernetes-1.23" ; then
         return 0
     fi
+    return 1
 }
 
 function build_and_upload() {
--- a/bin/upload-dist-staging.sh
+++ b/bin/upload-dist-staging.sh
@@ -29,6 +29,12 @@ function package_has_changes() {
         return 0
     fi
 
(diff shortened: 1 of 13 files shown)


bin/upload-dist-staging.sh
```diff
@@ -45,6 +51,38 @@ function package_has_changes() {
...
+    fi
+    if echo "${package}" | grep -q "kubernetes-1.21" ; then
+        return 0
+    fi
+    if echo "${package}" | grep -q "kubernetes-1.22" ; then
+        return 0
+    fi
+    if echo "${package}" | grep -q "kubernetes-1.23" ; then
+        return 0
+    fi
```
emosbaugh, line 83: wanna explicitly return 1 here?

```suggestion
    fi
    return 1
```

emosbaugh (approved):

bin/upload-dist-staging.sh
laverya: yeah I didn't think that one through

emosbaugh (approved):

PR #5071, opened 2024-02-09
update flannel v0.24.2

laverya:

Adds a systemd drop-in to disable TCP checksum offloading for VMware VMXNET3 NIC drivers - this is defaulted to ON in modern Linux kernels, but in certain VMWare deployments, the checksum offloading is handled incorrectly leading to dropped packets on the interface.  Detecting it is difficult because it only manifests in multi-node configurations once additional nodes have been added, but host networking works just fine so kURL does not detect a problem at installation time.  Disabling this for all VMXNET3 driver users seems like a safe assumption and disabling TCP checksum offloading incurs very little performance penalty.

The `ethtool` command can be used to fix the problem in-situ but changes made with that tool are not persistent, so this systemd unit hooks into the flannel device unit file and disables the offloading option whenever the flannel interface is configured.

Researching the issue 

See also 
- https://github.com/projectcalico/calico/issues/4727
- https://t.du9l.com/2020/03/kubernetes-flannel-udp-packets-dropped-for-wrong-checksum-workaround/

Fixes https://app.shortcut.com/replicated/story/91846/hostpreflight-detect-if-host-nic-uses-vmxnet3-driver-from-vmware-and-if-so-warn-about-vxlan-checksum-offloading

## Steps to reproduce

1. Install a VMWare ESXi cluster with NSX networking enabled for software-defined-networking at the VMWare cluster level 
2. configure >2 VMs with the VMXNet3 NIC driver 
3. install kurl on primary
4. add secondary
5. observe that pod-pod traffic between nodes fails silently

```release-note
NONE
```

NONE

### DIFF
--- a/addons/flannel/0.24.2/flannel-ethtool.service
+++ b/addons/flannel/0.24.2/flannel-ethtool.service
@@ -1,0 +1,12 @@
+[Unit]
+Description=Disable vxlan checksum offloading for flannel.1
+After=sys-devices-virtual-net-flannel.1.device
+Requires=sys-devices-virtual-net-flannel.1.device
+
+[Service]
+Type=oneshot
+ExecStart=/sbin/ethtool -K flannel.1 tx-checksum-ip-generic off
+RemainAfterExit=yes
+
+[Install]
+WantedBy=sys-devices-virtual-net-flannel.1.device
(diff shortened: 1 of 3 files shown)


ricardomaraschini:
I don't want this to become bigger and bigger but shouldn't we be also shipping the ethtool host package if we are using it ?

adamancini:
I created https://github.com/replicatedhq/kURL/pull/5076 because I had to add some fixes to my code, and I don't have push access to kURL repo

adamancini:
@laverya ready for review again, thanks!

laverya (commented):
Looks good - I can't approve because I'm the PR author

adamancini (approved):
@laverya approves

PR #5076, opened 2024-02-12
update flannel v0.24.2

adamancini:

updates https://github.com/replicatedhq/kURL/pull/5071

```release-note

```


adamancini:
@laverya @ricardomaraschini so far I've been looking at base installations for the distros on our supported list at kurl.sh and it looks like `ethtool` is included by default in all the Ubuntus and centos-7, going to check the redhats but I have a feeling this will reliably exist

laverya:
@adamancini you need to make sure you're making the same edits to both the template and the actual version

adamancini:
closed in favor of #5071

PR #5077, opened 2024-02-13
Automated Contour version update 1.28.2

replicated-ci-kurl:
Automated changes by the [cron-contour-update](https://github.com/replicatedhq/kURL/blob/main/.github/workflows/update-contour.yaml) GitHub action

```release-note
Adds [Contour add-on](https://kurl.sh/docs/add-ons/contour) version 1.28.2.
```


laverya:
looks like this is failing because we're adding a "x-kubernetes-validations" extension, which is _quite_ new
I'm going to test with k8s 1.29, and then look at removing that extension so that we don't have that as our minimum k8s version

ricardomaraschini (approved):

PR #5129, opened 2024-04-09
bug: stop installing fio on amazon linux

ricardomaraschini:
we can't install fio on amazon linux because it brings the centos-release package as one of its dependencies. the latter overwrites /etc/os-release in the target system making it look like a centos installation

Fixes #100040


ricardomaraschini:
Screenshot 2024-04-09 at 12 17 46

laverya (approved):

PR #5169, opened 2024-04-29
bug: does not attempt online install if unnecessary

ricardomaraschini:
if all dependency packages are already installed on the system we don't need to try to go and fetch them online. this pr disables all repos other than the local kurl one if all dependencies were already installed.

Fixes #103780

```release-note
Stop using default yum repos if all dependencies are already installed on RHEL 9.
```

### DIFF
--- a/scripts/common/host-packages.sh
+++ b/scripts/common/host-packages.sh
@@ -279,10 +279,18 @@ gpgcheck=0
 repo_gpgcheck=0
 EOF
 
-    if [[ "${packages[*]}" == *"containerd.io"* ]] ; then
-        yum install --allowerasing -y "${packages[@]}"
+    if [ -n "$KURL_USE_RHEL9_REMOTE_YUM_REPOS" ] ; then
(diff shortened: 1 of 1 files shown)


scripts/common/host-packages.sh
```diff
@@ -279,10 +279,18 @@ gpgcheck=0
...
+            yum install -y "${packages[@]}"
+        fi
     else
-        yum install -y "${packages[@]}"
+        if [[ "${packages[*]}" == *"containerd.io"* ]] ; then
+            yum install --disablerepo=* --enablerepo="$reponame" --allowerasing -y "${packages[@]}"
+        else
+            yum install --disablerepo=* --enablerepo="$reponame" -y "${packages[@]}"
+        fi
     fi
```
laverya, line 294: Instead of adding this if block, can we pass an env var that either includes the `--disablerepo=*` flag or not?

ricardomaraschini: Let me know if this is what you meant.

laverya (approved):

PR #5173, opened 2024-05-01
chore(testgrid): pre install rhel 9 dependencies

ricardomaraschini:

we aren't installing the dependencies anymore. now we need to pre install them every time we run test on top of rhel 9.


ricardomaraschini:
We are going to make these packages install when building the base image. Closing this one.

PR #5187, opened 2024-05-15
Automated EKCO version update 0.28.7

replicated-ci-kurl:
Automated changes by the [cron-ekco-update](https://github.com/replicatedhq/kURL/blob/main/.github/workflows/update-ekco.yaml) GitHub action

```release-note
Adds [EKCO add-on](https://kurl.sh/docs/add-ons/ekco) version 0.28.7.
```


laverya (approved):

laverya (approved):

laverya:
This is failing a whole row of tests in Testgrid - however we have validated that this is also broken with the current version of the addon 😱 , and so this does not break things further but does resolve CVEs
As such it's worth merging

PR #5248, opened 2024-06-27
Use cluster-cidr from kube-controller-manager.yaml

hedge-sparrow:
Instead of using the in-cluster kubeadm config which sometimes? loses the configured podSubnet, instead pull the podSubnet from the kube-controller-manager static manifest. this should prevent an issue where kurl sometimes resets the podSubnet on upgrade.

Fixes #

```release-note

```


laverya (approved):
I would love to see a 'upgrade with custom cidr' integration test that actually validated the IPs after the upgrade though

hedge-sparrow:
testgrid tests look alright, some seemingly unrelated failures

PR #5323, opened 2024-08-16
feat: support for amazon linux 2023

ricardomaraschini:

Adds support for Amazon Linux 2023.

Fixes #106636

We have decided to stop installing Docker `containerd.io` packages on Amazon Linux 2023. We are installing the Containerd version provided by the underlying OS. On this case it does not matter the version the user has selected in the Kurl Installer, it is always going to install whatever the OS says is the last version.

As this impacts all versions of Containerd I had to backport the code to all versions. As the code gets changing while we keep going back in time I have opted to cut at `1.5.10`. If one attempts to install while selecting a version older than that the installer will report back an error as the code is not prepared to do so. This is far from ideal, if you can think in a better solution let me know.

```release-note
Added support for Amazon Linux 2023.
```

### DIFF
--- a/addons/containerd/template/script.sh
+++ b/addons/containerd/template/script.sh
@@ -152,6 +152,7 @@ function copy_generated_files() {
 
 UNSUPPORTED_CONTAINERD_MINORS="01234"
 
+
 VERSIONS=()
 function find_common_versions() {
     docker build --no-cache --pull -t centos7 -f Dockerfile.centos7 .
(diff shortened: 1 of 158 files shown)


laverya:
> As this impacts all versions of Containerd I had to backport the code to all versions. As the code gets changing while we keep going back in time I have opted to cut at 1.5.10. If one attempts to install while selecting a version older than that the installer will report back an error as the code is not prepared to do so. This is far from ideal, if you can think in a better solution let me know.

Just to make sure, containerd pre-1.5.10 still works on other OSes right?

addons/containerd/template/script.sh
```diff
@@ -269,6 +274,14 @@ function find_common_versions() {
             add_supported_os_to_manifest_file "$version" "ubuntu-22.04" "Dockerfile.ubuntu22"
         fi
 
+        if ! contains "$version" ${AMAZON2023_VERSIONS[*]}; then
+            echo "Amazon 2023 lacks version $version"
+            add_unsupported_os_to_preflight_file "$version" "amzn" "2023"
+        else
+            add_supported_os_to_preflight_file "$version" "amzn" "2023"
+        fi
```
laverya, line 282: if we're not including packages we probably don't ever need to add amzn 2023 to the unsupported list, we could just always say it's supported

ricardomaraschini: This PR has no reason to exist. I will rewrite the whole thing from scratch.

ricardomaraschini:
I will be reopening this. This is convoluted to the point that I don't even can reason about it anymore.

PR #5363, opened 2024-08-26
bug: bail out if no policycoreutils-python-utils

ricardomaraschini:

bails the installation if the installer detects that we have selinux rules to apply but the underlying system does not provide us with the semanage command (policycoreutils-python-utils).

Fixes #

```release-note

```

### DIFF
--- a/scripts/common/utilbinaries.sh
+++ b/scripts/common/utilbinaries.sh
@@ -301,6 +301,12 @@ function apply_selinux_config() {
         return
     fi
 
+    if is_amazon_2023 ; then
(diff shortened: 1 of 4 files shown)


scripts/common/utilbinaries.sh
```diff
@@ -301,6 +301,12 @@ function apply_selinux_config() {
         return
     fi
 
+    if is_amazon_2023 ; then
```
laverya, line 304: This package should be added to the deps list, not checked for here

laverya (approved):

PR #5374, opened 2024-08-28
chore: add missing collectd package

ricardomaraschini:

adds `collectd-disk` on amazon linux 2023.

Fixes #

```release-note

```

### DIFF
--- a/addons/collectd/v5/Manifest
+++ b/addons/collectd/v5/Manifest
@@ -5,3 +5,6 @@ yum8 collectd-disk
 yum9 collectd
 yum9 collectd-rrdtool
 yum9 collectd-disk
+yum2023 collectd
+yum2023 collectd-rrdtool
+yum2023 collectd-disk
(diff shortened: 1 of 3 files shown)


addons/collectd/v5/Manifest
```diff
@@ -5,3 +5,6 @@ yum8 collectd-disk
 yum9 collectd
 yum9 collectd-rrdtool
 yum9 collectd-disk
+yum2023 collectd
```
ricardomaraschini, line 8: These lines got removed (very early in the process) when I locally ran:

```
pkg=dist/collectd-v5.tar.gz
make $pkg && tar xzvf $pkg
```

Because they got removed they were not commited when I raised the Amazon 2023 PR. I guess that it was happening due to some dangling files I had under `build/`, once I deleted the `build/` directory it started to work as expected.

laverya (approved):

PR #5430, opened 2024-10-14
Remove npm from dependabot as there is no longer js code in this repo

laverya:
#### Which issue(s) this PR fixes:

Fixes #

#### Does this PR introduce a user-facing change?

```release-note

```

PR #5431, opened 2024-10-14
Revert "Remove npm from dependabot as there is no longer js code in this repo"

laverya:
Reverts replicatedhq/kURL#5430

... we actually still use JS in the build process. I'm an idiot.


St0rmz1 (approved):
approved

PR #5750, opened 2025-08-21
chore(velero): update to version 1.16.2

emosbaugh:

- velero-restore-helper image was rolled into velero image
- local-volume-provider image tag no longer has a v prefix
- velero no longer works with k8s < 1.25
- longhorn will not work with k8s > 1.24 so tests have been migrated to use openebs

Fixes #

```release-note

```


diamonwiggins (approved):
lgtm

PR #5800, opened 2025-09-22
Automated Velero version update 1.17.2

replicated-ci-kurl:
Automated changes by the [cron-velero-update](https://github.com/replicatedhq/kURL/blob/main/.github/workflows/update-velero.yaml) GitHub action

[sc-138665]

```release-note
Adds [Velero add-on](https://kurl.sh/docs/add-ons/velero) version 1.17.2.
```


xavpaice:
TODO: we need some logic to prevent folks from upgrading from 1.16 if the LVP is in use (and therefore backups would be broken).

nvanthao (approved):

PR #6008, opened 2026-05-11
feat(ci): add CMX cluster validation workflow

kriscoleman:
## Summary
This PR introduces a new GitHub Actions workflow that tests kURL deployments on Compatibility Matrix (CMX) clusters. It creates a real kURL cluster via the replicated CLI, validates provisioning, and ensures the cluster is functional before cleanup.

## Changes
- Added `.github/workflows/test-cmx-clusters.yaml`
- Workflow runs on `workflow_dispatch` and daily at 06:00 UTC via cron schedule
- Installs replicated CLI and creates a kURL CMX cluster (`r1.large` instance type)
- Validates cluster health by:
  - Retrieving kubeconfig and listing nodes, namespaces, pods, and services
  - Verifying the `kurl` namespace exists
  - Waiting for all nodes to be Ready
  - Testing DNS resolution within the cluster
- Implements automatic cleanup with `always()` condition to remove CMX clusters even if tests fail
- Restricted to run only on the `replicatedhq` organization

## UAT
1. Go to the **Actions** tab in the repository
2. Select **test-cmx-clusters** workflow
3. Click **Run workflow** and trigger it manually
4. Verify that:
   - The CMX cluster is created successfully
   - Validation steps pass (nodes ready, DNS works, kurl namespace exists)
   - Cluster is removed automatically after completion (check the cleanup step)

Fixes: [sc-137574]

### DIFF
--- a/.github/workflows/test-cmx-clusters.yaml
+++ b/.github/workflows/test-cmx-clusters.yaml
@@ -1,0 +1,93 @@
+# This workflow tests kURL on CMX (Compatibility Matrix) clusters.
+# It creates a real cluster via the replicated CLI and validates that kURL provisions it correctly.
+# Required secret: REPLICATED_API_TOKEN
+name: test-cmx-clusters
+
+on:
+  workflow_dispatch: {}
+  schedule:
+    - cron: "0 6 * * *"  # daily at 06:00 UTC
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: false
+
+jobs:
+  test-kurl-cmx-cluster:
+    if: ${{ github.repository_owner == 'replicatedhq' }}
+    runs-on: ubuntu-latest
+    timeout-minutes: 30
+    steps:
+      - uses: actions/checkout@v6
+
+      - name: Install dependencies
+        run: |
+          sudo apt-get update && sudo apt-get install -y jq
+
+      - name: Install replicated CLI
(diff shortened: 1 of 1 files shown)


.github/workflows/test-cmx-clusters.yaml
```diff
@@ -0,0 +1,93 @@
+# This workflow tests kURL on CMX (Compatibility Matrix) clusters.
+# It creates a real cluster via the replicated CLI and validates that kURL provisions it correctly.
+# Required secret: REPLICATED_API_TOKEN
+name: test-cmx-clusters
+
+on:
+  workflow_dispatch: {}
+  schedule:
+    - cron: "0 6 * * *"  # daily at 06:00 UTC
```
xavpaice: can we make this on release, rather than cron?  I figure noone will look at the results of a daily check, but when we do a release do we do want to have a look at the release jobs to make sure it's clean.

xavpaice:
Update: I've removed the schedule from the workflow, there's a followup PR to make that triggers this workflow on release.  What I'd like to do is have this reviewed and merged, then test the workflow (can't test unless the file is in Main), prior to automating the trigger.

.github/workflows/test-cmx-clusters.yaml
kriscoleman: ooops, I missed this comment

nvanthao (approved):

PR #5342, opened 2024-08-21
kurl ubuntu 24 04 lts support

laverya:
#### What this PR does / why we need it:

This begins support for Ubuntu 24.04.

#### Which issue(s) this PR fixes:

Fixes #

#### Does this PR introduce a user-facing change?

```release-note

```

PR #6083, opened 2026-08-04
refactor(os-support): render the OS-support matrix from a single os-matrix.yaml source

### LINKED ISSUE #6081: chore(os-support): make adding a supported OS a single-source, DRY change
Adding one Ubuntu version to kURL currently requires a sprawling, error-prone, copy-paste PR. [replicatedhq/kURL#6072](https://github.com/replicatedhq/kURL/pull/6072/changes) ("Add Ubuntu 26.04 support") touches **32 files / +238 −19** and is almost entirely mechanical duplication — the same OS id, cloud-image URL, package lists, preflight rules, and `unsupportedOSIDs` entries pasted next to their `ubuntu-2404` twins. This is repeated in full every time a distro is added (24.04 was the same shape — see #5342), and every hand-copied line is a place a regression or a missed test-matrix entry can hide.

We want a **DRY design where the OS support matrix lives in one source of truth and every downstream artifact is rendered from it** (statically, checked-in), so a future "add Ubuntu 28.04" is a one-entry change plus a regenerate — with **no regressions to installer behavior and no regressions to the Testgrid test plan**.

## The redundancy, mapped (from #6072)

Six distinct duplication buckets, all keyed off the OS:

1. **Testgrid OS definitions** — the identical `ubuntu-2604` block (id / name / version / `vmimageuri` / `preinit`) is pasted into 4 existing spec files (`os-customer-common`, `os-firstlast`, `os-full`, `os-latest`) plus two new files (`os-ubuntu-2604.yaml`, `ubuntu-2604.yaml`). The cloud-image URL and preinit package list are repeated verbatim 6×.
2. **`unsupportedOSIDs` enumeration** — the bulk of the diff. Every spec that already excludes `ubuntu-2404` must be hand-edited to also exclude `ubuntu-2604`, with a copy-pasted `# Kubernetes versions < 1.24 ...` comment. ~90 such lines across `full.yaml`, `deploy.yaml`, and every addon `template/testgrid/*.yaml`. The *rule* ("k8s < 1.24 unsupported on this OS") is expressed once per OS but enumerated once per spec.
3. **Per-addon host-package install calls** — `ubuntu_2604_install_host_packages <pkgs>` pasted next to every existing `ubuntu_2404_install_host_packages <pkgs>` call across collectd, containerd, longhorn, openebs, prometheus, rook, velero, and the shared specs.
4. **Preflight OS rules** — the 26.04 fail/pass rules duplicated in `pkg/preflight/assets/host-preflights.yaml`, in per-containerd-version `host-preflight.yaml` (2.2.5, 2.2.6, template), and generated again in `addons/containerd/template/script.sh`.
5. **Shell OS predicates & lists** — `is_ubuntu_2604()` hand-added to `host-packages.sh` and `containerd-test.sh`; the OS appended to `bailIfUnsupportedOS` case list, `host_packages_shipped`, apparmor quirks, etc.
6. **Build plumbing** — a whole new near-identical `bundles/k8s-ubuntu2604/Dockerfile`, a near-identical Makefile target, and an `apt26` case in `save-manifest-assets.sh` cloned from `apt24`, plus `apt26`/`apt24` manifest tokens per containerd version.

## Existing anchor — the repo already templates

kURL already renders static artifacts from templates: addons use `addons/<name>/template/` → checked-in versioned dirs, and `addons/containerd/template/script.sh` already **generates** OS preflight and manifest entries via `add_supported_os_to_preflight_file` / `add_os_package_to_manifest_file`. The fix extends this established "template → render static, commit output" pattern to the OS matrix rather than inventing a new mechanism.

## Proposed DRY design

- **One OS registry** (single data file, e.g. `os-matrix.yaml`) — each entry: `id`, `distro`, `version`, package family (`apt24`/`apt26`/`yum2023`…), `vmimageuri`, `preinit`, and capability constraints (e.g. `minKubernetes: 1.24`, `dockerSupported: false`, `apparmorWorkaround: true`).
- **Render every downstream artifact from it**:
  - Testgrid OS spec fragments (or teach the tgrun harness to read the registry directly instead of per-file OS blocks).
  - `unsupportedOSIDs` **computed** from capability rules ("exclude OSes whose `minKubernetes` > the spec's k8s version") instead of hand-enumerated.
  - `install_host_packages` dispatch keyed off detected OS (or generated per-OS helper fns).
  - Preflight fail/pass rules generated from the registry constraints.
  - Shell `is_ubuntu_XXXX` predicates + `bailIfUnsupportedOS` case list generated into the assembled scripts.
  - **One parametrized** Ubuntu Dockerfile (`ARG UBUNTU_VERSION`) + a Makefile pattern rule over the OS list; a data-driven `apt<NN>` case.
- **Commit generated outputs** (as kURL already does for addons) so runtime behavior is byte-for-byte unchanged.

## No-regression strategy (hard requirement)

- **Golden diff**: generate against current `main` and prove the rendered artifacts are identical (or provably semantically equal) to today's checked-in files for all *existing* OSes. Zero behavioral delta is the gate.
- **Test-plan preservation**: the set of (spec × OS) combinations Testgrid runs must be unchanged for existing OSes — add a CI check that the computed matrix equals the pre-refactor matrix.
- **CI drift guard**: a `make verify-generated` (or equivalent) that fails if committed generated files are stale, so the single source can't silently diverge.
- Land as a pure refactor first (no new OS); prove Ubuntu 26.04 can then be added by a **single registry entry + regenerate** as the acceptance demo.

## Acceptance Criteria

- [ ] A single OS-matrix source of truth exists; no OS id, cloud-image URL, or package list is defined in more than one place.
- [ ] Testgrid OS specs, `unsupportedOSIDs`, host-package install calls, preflight OS rules, shell OS predicates, and the Ubuntu bundle Dockerfile/Makefile target are all derived from that source (full scope — all six buckets, including build plumbing).
- [ ] `unsupportedOSIDs` are computed from declarative per-OS capability rules, not hand-enumerated per spec.
- [ ] Generated/committed artifacts are byte-identical (or provably equivalent) to current `main` for all existing OSes — **no installer behavior change**.
- [ ] The Testgrid (spec × OS) coverage matrix is unchanged for existing OSes — **no test-plan regression**; a CI check enforces this.
- [ ] A CI guard fails when committed generated files are out of date.
- [ ] Demonstrated: adding Ubuntu 26.04 (redoing #6072) reduces to a single registry entry + regenerate, yielding an equivalent result to #6072 with a fraction of the hand-edited surface.
- [ ] `make test` (Go tests, lint, vet, manifest check) passes.

## Notes

- Reference prior full-shape PRs: #6072 (26.04), #5342 (24.04).
- Scope guard: this is a build/generation + test-harness refactor. It must not change what gets installed on any currently-supported OS.
- Follow-on: the same registry can later absorb RHEL/CentOS/Rocky/Amazon variants, which show the same enumeration pattern in `unsupportedOSIDs`.

kriscoleman:
Adding one Ubuntu version to kURL is a slog right now. #6072 (Ubuntu 26.04) touched 32 files for +238/-19, and almost all of it was copy-paste: the same OS id, cloud-image URL, package lists, preflight rules, and `unsupportedOSIDs` entries pasted next to their `ubuntu-2404` twins. #5342 (24.04) was the same shape. Every hand-copied line is a place a regression or a missed test-matrix entry can hide.

So I pulled the OS support matrix into one place and made everything downstream render from it.

## The design

One source of truth: `os-matrix.yaml` at the repo root. Each OS is declared once (id, name, version, cloud image, preinit, package family, and capability flags like `minKubernetes` / `dockerSupported`). A small Go tool, `os-matrix-gen`, renders every OS-keyed artifact from it, and a `make verify-generated` target keeps the committed output honest.

The six duplication buckets from #6072, and how each is handled:

- **Testgrid OS specs (bucket 1):** the `testgrid/specs/os-*.yaml` pool files are rendered whole and come out byte-identical to `main`. The ubuntu-2604 block that used to live in four files now lives once.
- **`unsupportedOSIDs` (bucket 2):** the OS-capability exclusions (k8s < 1.24, docker unsupported) are computed from the registry's capability rules and spliced into `BEGIN/END` marker regions. Spec-specific exclusions stay hand-authored, outside the markers. The transform only touches regions that already list the full capability set, and it's append-only, so it can't quietly change what runs.
- **Host-package calls (bucket 3), preflight rules (bucket 4), shell predicates (bucket 5):** same marker-region approach. `is_ubuntu_2604`, the `bailIfUnsupportedOS` case, the host-preflight fail rules, the containerd apparmor guard, all generated.
- **Build plumbing (bucket 6):** one `ARG`-parametrized Ubuntu bundle Dockerfile, a generated Makefile target list, and the `apt<NN>` case in `save-manifest-assets.sh`.

Everything outside bucket 1 is generated in place with markers, so the diff against `main` for those files is only the marker comment lines. No behavior change.

## No regressions, and I mean provably

This was the whole point, so it's guarded hard:

- `TestGoldenPools` proves the rendered `os-*.yaml` are byte-identical to the committed files.
- `TestGoldenBundleDockerfiles` does the same for the Ubuntu bundle Dockerfiles.
- `TestGoldenMatrixSnapshot` freezes the (spec × OS) coverage: the exact set of excluded OS ids per test. The refactor can re-author the lists all it wants, but it can't change which combinations Testgrid runs.
- `make verify-generated` is a prereq of `make test`, so CI fails the moment a committed generated file drifts from the registry.

## Registry-input hardening

The registry is a source file people edit, so its values get format-validated before anything renders from them. `Matrix.validate()` checks every scalar that gets interpolated into a shell script, Dockerfile, YAML doc, or write path: slugs for ids, digits for version majors, an injection-safe charset for versions and URLs, no newlines in names. A bad edit fails generation loudly instead of injecting into generated output. `safeJoin()` also keeps every generated path inside the repo root.

## The payoff

Adding or removing an OS is now one registry entry plus a regenerate, and it's symmetric: drop the `ubuntu-2604` entry, run `make generate-os-matrix`, and it disappears from every generated bucket; add it back and it returns. That's the #6072-sized PR collapsed to a couple of lines.

## Intentionally left alone

A couple of things stay hand-authored on purpose, and it's noted in the registry header:

- The pinned, frozen older containerd versions (their `apt<NN>` Manifest lines and install.sh guards). Those are released and immutable.
- The legacy Ubuntu Makefile targets for 18.04/20.04/22.04, which use an older build flow than the modern apt-bundle one.

Fixes #6081.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


kriscoleman:
## Review record (con-voyage escort)

Reviewed over 3 rounds by independent code (Go/Shell) and security reviewers before this draft opened.

**Code review — PASS.** Independently recomputed the coverage-matrix snapshot from `main` and confirmed it is byte-identical to the committed baseline — so the equivalence proof (`TestGoldenPools`, `TestGoldenMatrixSnapshot`, `TestGoldenBundleDockerfiles`, `TestGoldenSpliceRegionsInSync`) freezes `main`'s real behavior, not self-referential output. Confirmed `verify-generated` is a `make test` prerequisite (CI drift guard), the marker-region transform is append-only/subset-safe, and `os-matrix-gen check` reports no drift.

**Security review — PASS ("ship it").** Full OS/Pool struct audit: every string scalar reaching generated shell/Dockerfile/YAML/paths is now format-validated by `Matrix.validate()` (registry boundary), closing injection classes F1–F4; `safeJoin` gives path containment. Verified the registry-input hardening is behavior-neutral (no golden drift).

Adding a new OS is now a single `os-matrix.yaml` entry + `make generate` (symmetric add/remove demoed). Intentional residuals noted in the PR body (pinned older containerd versions; legacy 18/20/22.04 Makefile targets). Kept as a **draft** pending human review.

kriscoleman:
Addressed both Greptile P1 findings (commit `23d26e707`):

- **Capability exclusions never shrink** (`unsupported.go`): `renderUnsupportedCapability` is now authoritative — still-justified lines kept byte-identical, newly-constrained appended, and registry-unjustified lines **dropped**, so removing an OS or relaxing `minKubernetes`/`dockerSupported` correctly shrinks the marked region. Both shrink **and** grow are now per-category, so mixed-category marker regions (e.g. amazon-2023 `#docker` beside ubuntu `#k8s`) are handled per line without double-listing. New golden tests cover the shrink and grow directions (OS-removed, constraint-relaxed, non-primary-category grow).
- **Post-command flags ignored** (`os-matrix-gen/main.go`): now parses a per-subcommand `flag.FlagSet` on the args after the subcommand, so `generate -matrix <path>` / `check -root <dir>` are honored; a bogus subcommand errors clearly. Test added.

`make test` green, `verify-generated` reports zero drift (committed specs byte-identical). Reviewed clean by independent code + security reviewers.

kriscoleman:
Traced the new Greptile P1 (absent capability category stays omitted) and I'm confident it's a false positive here, so I've locked the reasoning in rather than change behavior.

The claim is that a docker-only region skips a newly-added k8s-constrained OS, and TestGrid then tries the unsupported install. That omission is actually correct. `validate()` forces every k8s-constrained OS to share one floor (divergent floors fail at parse). So a region with no k8s lines belongs to a spec whose k8s version is already at or above that floor, otherwise it'd already be excluding the existing constrained OSes. A newly-added k8s OS shares that same floor, so the spec still supports it and it has to stay omitted. Growing an untracked category would wrongly skip a combination the matrix says should run, and break the golden snapshot.

The concrete scenario doesn't hold either: `deploy` runs k8s 1.25, and a floor of 1.26 would get rejected by `validate()` as divergent before it could ever land.

Pushed a doc comment spelling out the invariant plus `TestRenderUnsupportedCapabilityDoesNotGrowUntrackedCategory`, which also fails if someone "fixes" this by blind-growing every category. No output change: golden snapshot byte-identical, `verify-generated` green, lint and vet clean.

xavpaice (approved):

PR #6115, opened 2026-08-28
fix(velero): include s3cmd image in Velero 1.18.2 manifest and sync Registry 3.0.0/3.1.0

xavpaice:
Addresses the review comment on the automated Velero update PR (#6114).

## Issue

The `cron-velero-update` PR bumps the `kurlsh/s3cmd` image used by the Velero S3-to-PVC migration patch, but the `kurlsh/s3cmd:20260825-a9ea1c6` image is missing from the Velero 1.18.2 `Manifest`. Because the manifest does not list it, the image is not packaged for air-gap installs, so the `migrate-s3` init container cannot start during the migration path.

## Changes

1. **Velero template & 1.18.2 add-on** – restore the `image s3cmd kurlsh/s3cmd:__S3CMD_TAG__` line to `addons/velero/template/base/Manifest.tmpl` and regenerate the 1.18.2 add-on. This also picks up the latest `local-volume-provider` tag (0.6.15) in the generated `install.sh`.
2. **Registry 3.0.0 & 3.1.0** – sync the `kurlsh/s3cmd` image tag to the latest `20260825-a9ea1c6` in both the manifests and the S3 migration/Velero backup patches, so the image is present in both selected add-on packages for the air-gap migration path.

## Verification

- Confirmed the generated `addons/velero/1.18.2/Manifest` now includes `image s3cmd kurlsh/s3cmd:20260825-a9ea1c6`.
- Confirmed Registry 3.0.0 and 3.1.0 manifests and patches reference the same tag.
- Confirmed the registry template base already includes the s3cmd placeholder; only the pinned 3.0.0 and 3.1.0 versions needed the sync.

### DIFF
--- a/addons/registry/3.0.0/Manifest
+++ b/addons/registry/3.0.0/Manifest
@@ -1,2 +1,2 @@
 image registry registry:3.0.0
-image s3cmd kurlsh/s3cmd:20260224-0d00dd0
+image s3cmd kurlsh/s3cmd:20260825-a9ea1c6
--- a/addons/registry/3.0.0/patch-deployment-migrate-s3.yaml
+++ b/addons/registry/3.0.0/patch-deployment-migrate-s3.yaml
@@ -10,7 +10,7 @@ spec:
     spec:
       initContainers:
       - name: migrate-s3
-        image: kurlsh/s3cmd:20260224-0d00dd0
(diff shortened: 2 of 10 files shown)


kriscoleman (approved):
:shipit:

